PR #1, opened 2018-02-25
fix stuff doctr wrote over

CJ-Wright:


CJ-Wright:
Now all the additional patterns are auto generated, so if we find a new one we can just add it.

PR #3, opened 2018-02-25
use requests rather than github3.py for 01 so we don't run out of

CJ-Wright:
api calls

### DIFF
--- a/01-make_graph.py
+++ b/01-make_graph.py
@@ -8,6 +8,7 @@ from base64 import b64decode
 
 import github3
 import networkx as nx
+import requests
 import yaml
 from jinja2 import UndefinedError, Template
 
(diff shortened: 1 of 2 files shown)


01-make_graph.py
```diff
@@ -74,8 +75,12 @@ def source_location(meta_yaml):
...
-        feedstock = gh.repository('conda-forge', name + '-feedstock')
-        meta_yaml = feedstock.contents('recipe/meta.yaml')
+        r = requests.get('https://api.github.com/repos/conda-forge/'
+                         '{}-feedstock/contents/recipe/meta.yaml'.format(name),
+                         auth=(os.environ['USERNAME'], os.environ['PASSWORD']))
+        if r.status_code != 200:
+            raise github3.GitHubError(r)
+        meta_yaml = r.json()['content']
         if meta_yaml:
             text = codecs.decode(b64decode(meta_yaml.content))
```
justcalamari: `meta.yaml` is just a string now, so should just be `text = codecs.decode(b64decode(meta_yaml))`

CJ-Wright: thank you!

PR #5, opened 2018-02-25
write non 403 errors into bad

CJ-Wright:


01-make_graph.py
```diff
@@ -83,6 +83,8 @@ def source_location(meta_yaml):
         elif r.status_code != 200:
             print('Something odd happened to this recipe '
                                '{}'.format(r.status_code))
+            with open('bad.txt', 'a') as f:
+                f.write('{}\n'.format(name))
```
justcalamari: Might want to also put in what the error code is

CJ-Wright: maybe, but we directly parse bad without a second look I think

PR #6, opened 2018-02-25
set to get the hashes

CJ-Wright:


01-make_graph.py
```diff
@@ -115,11 +115,10 @@ def source_location(meta_yaml):
...
                     'url': yaml_dict['source']['url'],
                     'req': req,
-                    'time': time.time(),}
-                for k in yaml_dict['source']:
-                    if k in hashlib.algorithms_available:
-                        sub_graph['hash_type'] = k
-                        break
+                    'time': time.time(),
+                    'hash_type': (set(yaml_dict['source'].keys())
+                                  & hashlib.algorithms_available).pop()
```
justcalamari: This will give an error if there is no key for the hash type in source. This shouldn't happen, but if it does we don't want the whole thing to break.

CJ-Wright: fair, not certain what the best way to get an item out of a set is

PR #14, opened 2018-03-02
use warehouse pypi.org instead of legacy pypi.python.org

minrk:
fixes latest version when a prerelease is available

legacy PyPI uses the last release to populate 'info', whereas warehouse correctly uses the latest *stable* release

compare https://pypi.org/pypi/tornado/json (info.version = 4.5.3) with https://pypi.python.org/pypi/tornado/json (info.version = 5.0b1)

retrieving prerelease version produced an incorrect message about tornado being out of date in https://github.com/conda-forge/nbdime-feedstock/pull/11

The alternative is to parse, sort, and filter all of the versions in the legacy report yourself, I think.

PS autotick-bot is super cool!


CJ-Wright:
This is great! I think we will also need some filtering tools for those packages which pull from GH, but there is no reason not to have both!

PR #47, opened 2018-03-03
Replace `this repo` with link to cf-graph in README

sodre:
- After the package was split, the graph is kept on a separate repository.


CJ-Wright:
Thank you!

PR #95, opened 2018-03-20
FIX: if no url in meta.yaml add to upstream bad

CJ-Wright:
Caused by https://github.com/conda-forge/esmf-feedstock/blob/master/recipe/meta.yaml#L9


CJ-Wright:
@justcalamari @isuruf ready for review

PR #110, opened 2018-03-29
blacklist a and b in version

CJ-Wright:
This blacklists version numbers containing ``a`` or ``b``.
Attn:
@justcalamari I guess I got to it first :)
@ocefpaf
@jakirkham
@ericdill

I'm putting up this PR so we can merge it if we want to and to continue the discussion. I guess I'd rather default to the safer approach (not bumping to dev versions) until we have the infrastructure to handle dev releases more formally (pushing to dev branches,/labels not updating everything to the dev versions when installing things, etc.) However, I would like to get input/consensus if possible, and am more than happy to follow the majority's opinion.

### DIFF
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -33,7 +33,7 @@ def next_version(ver):
 
 class VersionFromFeed:
     ver_prefix_remove = ['release-', 'releases%2F', 'v']
-    dev_vers = ['rc', 'beta', 'alpha', 'dev']
+    dev_vers = ['rc', 'beta', 'alpha', 'dev', 'a', 'b']
 
     def get_version(self, url):
         data = feedparser.parse(url)


jakirkham:
So I've heard from some people that they like the ability of getting automated version bumps for dev releases. Though am sure it depends on whether the feedstocks maintainers are project developers or not. Perhaps instead of continuing to blacklist these across the board we should devise an option for users to add an option to the `meta.yaml` to determine whether dev releases should happen and which ones. If the bot is able to inspect multiple branches, it would then know which branch to PR for free.

ref: https://github.com/ioam/holoviews/issues/2217#issuecomment-372400994

CJ-Wright:
I'm not against having dev releases (resource stress aside) and would prefer to not stand in the way of how people version/release their code. However, my main concern about dev releases is that most of the maintainers lack training/knowledge about what dev releases mean (some of which we are still discussing).
1. Dev releases shouldn't go into master.
1. Dev release at the very least need their own label.

This can cause some annoying breakage downstream (eg. everything with `github3.py` deps for the last couple months).

Currently I'm thinking of this PR as more of a stop gap to put us on solid ground before pursuing dev releases (if that is what we decide to do) and to prevent breakage. There is nothing stopping us from removing things from the blacklist as we develop more advanced methods and policies to deal with development releases.

I think that having an option in the `meta.yaml` would be a good thing in the long run. I think that the above issues could also be mitigated by enhancements to the bot; having the bot PR into a different branch and setup the configurations for dev releases properly. A note in the PR body stating that this release is a dev release and should be inspected with care might also be helpful.

scopatz:
Since we are already blacklisting dev releases, I think the additional spellings don't change the current policy, though I am in favor of opening up this in the future.

CJ-Wright:
Thank you!

Issue #54, opened 2018-03-05
Moving to a web service


jakirkham:
It’s certainly reasonable to start out with a corn job for these sorts of things. Also as we resolve some technical debt, the cron job is very helpful. That said, we have generally found in conda-forge that cron jobs inevitably struggle to scale.

To solve this problem, have ultimately moved all of them to web services that use webhooks. This allows them to deal with notifications as they come in and respond by doing some task. This approach seems well suited for updates. However it will require some thought into how we can get notifications from package indexes, GitHub, etc. Expect this will iron out any issues related to load.

CJ-Wright:
I think that part of this is wrapped up in https://github.com/regro/cf-scripts/issues/53, since it is difficult to know exactly what kind of stress to expect on the system without some ballpark numbers. The whole bootstrap of the system also exaggerates some things (since we max out the CI time on every run of `03`).

Currently we are clearing about 20 feedstocks per run with `03` and all the feedstocks with the others (although we'll fall behind when we hit 5000 feedstocks on `01`).

I'm not opposed to moving this to a webservice, but the notification wrangling could be hard.

We also could do this in steps/build a hybrid system:
`00` can be removed with a hook on staged
`01` can be removed with a hook on all the feedstocks when PRs are merged
`02` might be the most difficult to remove since listening for new releases from PYPI, CRAN, and GitHub may be difficult.
`03` I'm less certain about, since I don't know what triggers it (I guess whatever triggers `02`). If we can find a trigger for `02` then `02` and `03` could be merged.

jakirkham:
Honestly our experience doing this at conda-forge has taught us that the system ends up being less stressed when converted from batch to a webservice. If you think about it a bit, this actually makes sense. The reason being updates in a web service don't all come at once in a big batch (this could be re-renderings, updating Circle SSH keys, or package updates). Instead they are sprinkled throughout the days at various times. The result ends up being things stay pretty light and the system handles events right away, which makes the whole thing more maintainable.

[...]

Issue #44, opened 2018-03-03
Triggering rebuilds of dependency chains


jakirkham:
Admittedly this may not always make sense. However in case that it does, this would be very useful. Namely would be good to trigger rebuilds of downstream dependencies when an upstream dependency is rebuilt. As a simple example, `oniguruma` and `jq`.

CJ-Wright:
What is the best way to trigger a rebuild?
This sounds like it would need its own script. We could have a list of nodes who's downstreams need rebuilds.

jakirkham:
Presumably this would send a PR with a build number bump. It might also update the pinning, but I don't think this will matter for much longer (i.e. cb3 would make this step irrelevant).

Right was figuring we could already use the graph you have to solve this problem.

Yeah it would probably be a new script. Admittedly we might only care about doing this for things that require dependencies from our central pinning file at build time. In fact, it might be best to only look at changes in the central pinning file to determine when a rebuild needs to occur.

CJ-Wright:
Hmm ok, I think I get it.

We could have a flag in some nodes which notes that when the node is version bumped then all downstream nodes need a build bump (we could pull this from central pinning).
The flag defaults to false and is flipped to true when it is version bumped (and the version bump is accepted).
Another script follows behind and find all the nodes with a central pin where the flag is true and then issues build number bumps to all downstream packages, then flips the flag back to false.

We might something to note that we don't need a build number bump if a downstream package is getting a version bump itself (since that should trigger a rebuild).

jakirkham:
Right. Something like that sounds good.

Yeah, starts to overlap with issue ( https://github.com/regro/cf-scripts/issues/16 ).

Should add that re-rendering will be needed as part of this process as that is how we update the dependencies.

cc @isuruf

CJ-Wright:
How often do we want to do something like this? Every pinning update? If we change the `gcc` pinning how much of the graph are we prepared to rebuild?

jakirkham:
My suspicion is that this is going to be painful in the beginning. Partly because some things weren't using the latest recommended pinnings (i.e. in the old pinning script) and partly because the latest recommended pinnings are a bit outdated.

[...]

PR #115, opened 2018-04-02
write conda-smithy version for re-render into graph so we don't

CJ-Wright:
re-render if the versions are the same.
This should cause the PRs for the bot to go faster
(after a bootstrap phase)

### DIFF
--- a/03-auto_tick.xsh
+++ b/03-auto_tick.xsh
@@ -256,6 +256,7 @@ $REVER_DIR = './feedstocks/'
 gh = github3.login($USERNAME, $PASSWORD)
 
 t0 = time.time()
+smithy_version = ![conda smithy --version].output.strip()
 for node, attrs in gx2.node.items():
     # Don't let travis timeout, break ahead of the timeout so we make certain
     # to write to the repo
@@ -273,9 +274,14 @@ for node, attrs in gx2.node.items():
             if gh.rate_limit()['resources']['core']['remaining'] == 0:
                 break
             else:
-                run(pred=pred, gh=gh, rerender=True, protocol='https',
-                    hash_type=attrs.get('hash_type', 'sha256'))
-                gx.nodes[node]['PRed'] = attrs['new_version']
+                if gx.nodes[node].get('smithy_version') != smithy_version:
+                    run(pred=pred, gh=gh, rerender=True, protocol='https',
+                        hash_type=attrs.get('hash_type', 'sha256'))
+                else:
+                    run(pred=pred, gh=gh, rerender=False, protocol='https',
+                        hash_type=attrs.get('hash_type', 'sha256'))
+                gx.nodes[node].update({'PRed': attrs['new_version'],
+                                       'smithy_version': smithy_version})
         except github3.GitHubError as e:
             print('GITHUB ERROR ON FEEDSTOCK: {}'.format($PROJECT))
             print(e)


CJ-Wright:
@justcalamari @scopatz

dougalsutherland:
Worth keeping in mind that there are situations where a re-render will change something even with the same conda-smithy version: if e.g. a dependency on a certain platform gets added, or in the CB3 stuff when there's a change to conda-forge-pinning, right?

CJ-Wright:
That's a fair point, although I think it is made moot by two issues:
1. The bot doesn't add dependencies. If a maintainer adds a dep they would have to call for a rerender.
1. CB3 isn't out yet in production. This PR is mostly a stop gap, when CB/CS3 come out the bot will need some not so subtle changes to get everything humming along again. I expect this issue to go away/morph a bit with the CB3 things, eg [rerendering entire sections of the graph](https://github.com/regro/cf-scripts/issues/44)

dougalsutherland:
Yeah, I guess lots of changes will be needed for case 2.

In case 1, I was thinking of the following case:
- Package `a` depends on `foo`.
- When `a` was last rerendered, there was no Windows version of `foo`. So AppVeyor wasn't set up.
- In the meantime, a Windows version of `foo` was added.
- Now rerendering would include a Windows version.

This is a fairly unlikely case though, since there would usually be a `skip` statement. It's probably not worth worrying about; can always be fixed by asking the webservices bot for a rerender.

CJ-Wright:
I think there would have to be a skip statement. The maintainers would need to remove the skip statement manually at which point rerendering is in their hands.

dougalsutherland:
It's *possible* for there not to be – maybe with a version upgrade of `foo` – but it'd be unusual and probably not worth worrying about for the bot. Most likely case is probably with a race condition on upgrade releases.

CJ-Wright:
Ah, ok. The race condition thing is why we include the upstream/current versions in the PR body, so users can check. However, a newer version of the bot could make PRs in dependency order, so that the dep is build earlier.

jakirkham:
Yeah, if there is not a `skip`, it is probably `noarch`, which will just now be installable. Re-rendering wouldn't matter here either.

Even when dependencies do get handled by the bot, expect re-rendering with the same version of `conda-smithy` won't matter much actually. Thinking mainly of Python dependency management here. As the dependencies will be other Python packages, which we generally don't pin on a global level. Though story doesn't really differ for R, Perl, Go, etc.. The one case where this does matter is when we change the versions of Python, R, Perl, Go, etc., in which case we need to re-render. C/C++ dependencies will pretty much always be handled manually unless some radical change happens to both languages.

The main value to re-rendering (particularly with cb3) will be handling pinning changes. These will happen for existing dependencies. This includes changes to versions supported by Python, R, Perl, Go, etc. So maybe the best strategy will be what you have proposed here combined with some knowledge about what things live in that pinning file (and whether they showed up in some particular recipe that is being updated). Only re-rendering recipes where these pinned dependencies have changed somehow.

Stepping back a moment though. Would like to better understand the motivation behind this change. Expect there is some underlying issue we need to address.

CJ-Wright:
Just for context, currently we re-render for every bot PR. This can be a rather big time sink. Additionally our time on the CI is a bit limited, so it would be great to not re-render if it isn't needed. I thought that a good way to do this would be to cache the latest version we re-rendered with in the graph somewhere so if the version of smithy is the same as when we most recently re-rendered then we won't re-render.

We can have things be more intelligent about the pinning file and inspect the `meta.yaml` for changes in the pinning which requires a re-render. One could imagine the PR section of the bot as serving a dual purpose, it either a) updates the version but doesn't re-render, b) re-renders because of pinning changes but doesn't update the version, or c) do both a) and b).

One could also consider having an acceptable "window" of smithy versions which don't require a re-render. For example all smithy versions which support py36 are ok, but everything else needs a re-render.

dougalsutherland:
So another possible thing that could help would be to improve the speed of doing multiple re-renderings. I don't really know how that works exactly, but my impression is that each time you call `conda smithy rerender`, the slow part is that it downloads and parses the index file for each platform. If the bot could cache those index files over the course of a run, re-rendering might be much faster, making this less of an issue. (Maybe I'm wrong about that though.)

jakirkham:
Do we already have an issue open on `conda-smithy` for re-rendering performance?

Ultimately index caching might be a `conda` thing, but am not sure. Please see [this comment]( https://github.com/conda-forge/conda-smithy/issues/241#issuecomment-250510239 ). Thoughts on that issue and possibly at `conda` welcome.

As to resource limitations, this will be a persistent problem until we move to a webservice ( https://github.com/regro/cf-scripts/issues/54 ). Happy for any optimizations you deem necessary in the short term. These will just be bonuses in the long term.

Long term being able to re-render for pinnings ( https://github.com/regro/cf-scripts/issues/44 ) and/or version update sounds great.

We will need to figure out a better versioning scheme for `conda-smithy` and possibly breakout things like the linter ( https://github.com/conda-forge/conda-smithy/issues/386 ) to really provide windows. Suggestions and work in this area certainly welcome.

scopatz:
The code here looks good to me. Any reason not to merge?

PR #116, opened 2018-04-02
write archived status into graph if PR fails with exact error message

CJ-Wright:
`bob` is causing issues again. In this case this is blocking progress for the bot backlog due to an outage over the weekend.


CJ-Wright:
@justcalamari

PR #125, opened 2018-04-05
Take first url if all contain Archive

### LINKED ISSUE #124: r-corpor breaks things
It seems that there are to urls with `Archive` in them. I guess we could just take the first one then.
https://github.com/conda-forge/r-corpcor-feedstock/blob/master/recipe/meta.yaml

justcalamari:
Closes #124

### DIFF
--- a/03-auto_tick.xsh
+++ b/03-auto_tick.xsh
@@ -174,6 +174,7 @@ def run(feedstock=None, protocol='ssh',
                 if 'Archive' not in url:
                     source_url = url
                     break
+        if isinstance(source_url, list):
             source_url = source_url[0]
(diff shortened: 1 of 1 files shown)


03-auto_tick.xsh
```diff
@@ -174,6 +174,8 @@ def run(feedstock=None, protocol='ssh',
                 if 'Archive' not in url:
                     source_url = url
                     break
+        if isinstance(source_url, list):
```
scopatz: This shoulde really check if it is a non-string instance of `collections.abc.Sequence`, rather than strictly being a list.

scopatz:
Thanks for the PR @justcalamari!

PR #146, opened 2018-05-11
Use conda-forge specific version badge in the PR template

nicoddemus:
This is a badge specific for conda-forge and in my experience it updates
immediately contrary to the anaconda badge which takes a day or to.

For illustration purposes here's the badge for pytest:

img


scopatz:
Thanks!

Issue #32, opened 2018-03-03
Overwrites {{ build }} jinja


CJ-Wright:
@tkelman commented on [Fri Feb 23 2018](https://github.com/regro/cf-graph/issues/26)

see https://github.com/conda-forge/awscli-feedstock/pull/29

---

@CJ-Wright commented on [Fri Feb 23 2018](https://github.com/regro/cf-graph/issues/26#issuecomment-368188923)

Thank you for reporting!

---

@CJ-Wright commented on [Fri Feb 23 2018](https://github.com/regro/cf-graph/issues/26#issuecomment-368190833)

This came about because you were able to update the version before the bot realized there is a new version. Normally that PR would have included a version bump.

@scopatz thoughts?
This is being overwritten because in the replacements a) there is no `{{ build ...}` in our replacement setup, b) there is also a replacement which overwrites the build setting. I don't suppose that there is a meaningful way to do this in the regex?

To be fair, having the build number as a jinja variable is a bit uncommon, especially as it is usually only used once, hence why we didn't include it in the initial setup.

---

@scopatz commented on [Mon Feb 26 2018](https://github.com/regro/cf-graph/issues/26#issuecomment-368615773)

Well, you could check and make sure that the `build: NUM` really is just a a simple number before overwriting it.  If it isn't you can punt.

---

@scopatz commented on [Mon Feb 26 2018](https://github.com/regro/cf-graph/issues/26#issuecomment-368616302)

Basically, this comes down to the fact that the ticker is assuming that there is a specific form of meta.yaml, and if it isn't in that form, it will fail

Issue #21, opened 2018-03-02
The jinja var for build number is ignored


CJ-Wright:
@ocefpaf commented on [Thu Mar 01 2018](https://github.com/regro/cf-graph/issues/44)

See https://github.com/conda-forge/dropbox-feedstock/pull/4#discussion_r171557374

IMO we should just drop the jinja variable for build number. It is silly to have a variable that s used in one place. If that action is taken there is nothing to do in cf-graph. If not we need to fix the bot :unamused: 

---

@dougalsutherland commented on [Thu Mar 01 2018](https://github.com/regro/cf-graph/issues/44#issuecomment-369619312)

The reason I've often used it is just because it makes it harder to forget to reset the build number to 0 when you're bumping the version yourself if the build number is specified right next to the version, instead of potentially 15 lines away. It makes just as much sense as an `sha256` variable that's also only used once (as you noted there)....

If both `sha256` and `build` variables are going away in general to make life easier for bots, that's fine in that the build will definitely break if you forget to update the checksum. :) (Alternatively, maybe the linter could check that the build number is 0 for package versions that don't have a build yet?)

---

@ocefpaf commented on [Thu Mar 01 2018](https://github.com/regro/cf-graph/issues/44#issuecomment-369620622)

> The reason I've often used it is just because it makes it harder to forget to reset the build number to 0 when you're bumping the version yourself if the build number is specified right next to the version, instead of potentially 15 lines away. It makes just as much sense as an sha256 variable that's also only used once (as you noted there)....

I disagree but I am more used to the conda recipe format than most people. Anyway, I don't oppose to use the jinja, but I would love to get rid of the excess, like the file extension for example :smile: 

> If both sha256 and build variables are going away in general to make life easier for bots, that's fine in that the build will definitely break if you forget to update the checksum. :)

Yeah, in like of the age of the bots all this may change :smile: 

---

@CJ-Wright commented on [Thu Mar 01 2018](https://github.com/regro/cf-graph/issues/44#issuecomment-369643317)

I think we can implement either way. I like the idea of using the bot to PR in basic maintenance to recipes (removing excess variables, fixing jinja variables (there was at least on instance of `{%set...` rather than `{% set...`).

At the end of the day though the bots serve the humans, so unless a change/feature is especially painful for the bot, I'd rather have things be human friendly than tailored to the bot.

---

@CJ-Wright commented on [Fri Mar 02 2018](https://github.com/regro/cf-graph/issues/44#issuecomment-369971779)

[...]

PR #148, opened 2018-05-15
Update jinja var for build if exists

### LINKED ISSUE #94: When build number is a variable, change the variable, not build number
For example, here, no change should happen to build number.
https://github.com/conda-forge/lighttpd-feedstock/pull/1

I guess this should be a general rule: if some value is provided via a variable, change the variable value.

justcalamari:
Closes #94, #21, #32

### DIFF
--- a/03-auto_tick.xsh
+++ b/03-auto_tick.xsh
@@ -75,7 +75,7 @@ DEFAULT_PATTERNS = (
     ('meta.yaml', '{%set version = ".*" %}', '{%set version = "$VERSION" %}'),
     # reset the build number to 0
     ('meta.yaml', '  number:\s*[0-9]+', '  number: 0'),
-    ('meta.yaml', '{%\s*set build_number\s*=\s*".*"\s*%}, {% set build_number = "0" %}),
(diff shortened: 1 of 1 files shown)


CJ-Wright:
@jakirkham I don't think this solves the issue you found completely (since it will overwrite both build numbers), but I think it gets part of the way there.

@justcalamari do you have a link to that issue?

CJ-Wright:
This LGTM, I'll run a dedicated run on travis to see if this works.

CJ-Wright:
@justcalamari thank you very much!

justcalamari:
@CJ-Wright which issue?

CJ-Wright:
There was one that recently came up on a feedstock PR the bot issued. There was a selector associated with the build number.

justcalamari:
Ok I see it. One way to handle this is to only replace the jinja definition if the variable is a number, so `build_number = 2` gets replaced, but not `build_number = build_number + 200`.

justcalamari:
Is there an issue or is it just the PR?

CJ-Wright:
The PR please

justcalamari:
https://github.com/conda-forge/vtk-feedstock/pull/58#discussion_r187773678

PR #151, opened 2018-05-20
WIP: Add nodes to the graph even if there is a problem

### LINKED ISSUE #150: Don't require a URL/eliminate bad
It might be good to eliminate (most of) bad. There are some recipes which don't have urls (eg toolchain) that it would be nice to have in the graph.

One way to accomplish this would be to reduce the requirements to make it into the graph (eg. if it has a meta.yaml we accept it).

justcalamari:
Closes #150

### DIFF
--- a/conda_forge_tick/make_graph.py
+++ b/conda_forge_tick/make_graph.py
@@ -41,11 +41,11 @@ def make_graph(names, gx=None):
                 logger.warn('Something odd happened when fetching recipe '
                             '{}: {}'.format(name, r.status_code))
                 bad.append(name)
-		sub_graph['bad'] = r.status_code
-            	if name in new_names:
-            	    gx.add_node(name, **sub_graph)
-            	else:
-            	    gx.nodes[name].update(**sub_graph)
+                sub_graph['bad'] = r.status_code
+                if name in new_names:
+                    gx.add_node(name, **sub_graph)
+                else:
+                    gx.nodes[name].update(**sub_graph)
                 continue
 
             text = r.content.decode('utf-8')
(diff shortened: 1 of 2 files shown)


conda_forge_tick/make_graph.py
```diff
@@ -36,6 +41,11 @@ def make_graph(names, gx=None):
                 logger.warn('Something odd happened when fetching recipe '
                             '{}: {}'.format(name, r.status_code))
                 bad.append(name)
+		sub_graph['bad'] = r.status_code
```
CJ-Wright: This seems to be missing a space? (It looks like the line above doesn't match)

conda_forge_tick/make_graph.py
```diff
@@ -28,6 +28,11 @@ def make_graph(names, gx=None):
 
     for i, name in enumerate(total_names):
         try:
+            sub_graph = {
```
CJ-Wright: Should we pull the generation of the dict into its own function? That would cut down on the `if name in new_names...` boilerplate (since the function can return a dict at various points).

scopatz:
LGTM too, other than the issues that @CJ-Wright pointed out

justcalamari:
Right now the code still writes to bad.txt even though bad information is stored in the graph. Should we get rid of bad.txt altogether?

CJ-Wright:
I'd keep the redundancy until we know that putting it into the graph works.

conda_forge_tick/make_graph.py
```diff
@@ -10,6 +10,68 @@
...
+
+    keys = [('source', 'url'), ('package', 'name'),
+            ('package', 'version')]
+    missing_keys = [k[1] for k in keys if k[1] not in
+                    yaml_dict.get(k[0], {})]
+    if missing_keys:
+        logger.warn("Recipe {} doesn't have a {}".format(name,
+                    ', '.join(missing_keys)))
+        bad.append(name)
+        sub_graph['bad'] = ', '.join(missing_keys)
```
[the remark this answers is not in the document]
justcalamari: Thanks, good catch.

PR #161, opened 2018-05-29
Find the longest path to each node from a starting node

justcalamari:


conda_forge_tick/path_lengths.py
```diff
@@ -0,0 +1,40 @@
+import networkx as nx
```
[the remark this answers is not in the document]
scopatz: Also, according to PEP8, Python standard library imports should come before 3rd part library imports. Please swap these two and add a blank line in between.

conda_forge_tick/path_lengths.py
```diff
@@ -0,0 +1,40 @@
...
+def get_levels(graph_file, source):
+    g = nx.read_gpickle(graph_file)
+    g2 = deepcopy(g)
+    desc = nx.algorithms.descendants(g, source)
+    for node in g.node:
+        if (node not in desc and node != source):
+            g2.remove_node(node)
+
+    dist = get_longest_paths(g2, source)
+    return {v: [k for k in dist if dist[k] == v] for v in dist.values()}
```
scopatz: While this is perhaps the most readable nested comprehension I have ever seen, there are a couple of more efficient methods. For example, with a defaultdict you can only have one loop.

```py
from collections import defaultdict

rtn = defaultdict(list)
for k, v in dist.items():
    rtn[v].append(k)
return rtn
```

Or you can dive into using `itertools.groupby()` but I am not sure it will make the code more readable

justcalamari:
@scopatz Ready for review

scopatz:
Looks great, I think it just needs tests

justcalamari:
@scopatz Ready for review

conda_forge_tick/path_lengths.py
```diff
@@ -0,0 +1,109 @@
...
+        for v in graph.neighbors(u):
+            if v in visited:
+                continue
+            if dist[v] < dist[u] + 1:
+                dist[v] = dist[u] + 1
+
+    return dist
+
+
+def get_levels(graph_file, source):
```
CJ-Wright: Should this take in a graph rather than a file?

justcalamari: Yeah that's probably a good idea.

PR #180, opened 2018-06-08
Add attrs attribute to Migrator

justcalamari:
The bot was running into a problem where it used `migrator.remote_branch()` before the `version` attribute was being set, causing an error. This should fix that.

Also I think it is useful to store the attributes of the node currently being migrated inside the migrator because other subclasses of Migrator might want to use other attributes of the node.

### DIFF
--- a/03-auto_tick.xsh
+++ b/03-auto_tick.xsh
@@ -21,7 +21,7 @@ def run(attrs, migrator, feedstock=None, protocol='ssh',
         pull_request=True, rerender=True, fork=True, gh=None,
         **kwargs):
     # get the repo
-    migrator.set_attrs(attrs)
+    migrator.attrs = attrs
     feedstock_dir, repo = get_repo(attrs, branch=migrator.remote_branch(),
                                    feedstock=feedstock,
                                    protocol=protocol,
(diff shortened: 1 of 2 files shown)


conda_forge_tick/migrators.xsh
```diff
@@ -44,6 +44,16 @@ class Migrator:
...
 
+    def set_attrs(self, attrs):
+        """Set the attributes of the node currently being migrated.
+
+        Parameters
+        ----------
+        attrs: dict
+            The attributes.
+        """
+        self.attrs = attrs
```
scopatz: This is an anit-pattern in python

03-auto_tick.xsh
```diff
@@ -21,6 +21,7 @@ def run(attrs, migrator, feedstock=None, protocol='ssh',
         pull_request=True, rerender=True, fork=True, gh=None,
         **kwargs):
     # get the repo
+    migrator.set_attrs(attrs)
```
scopatz: This should just be `migrator.attrs = attrs`, since attrs is public.

scopatz:
Thanks for putting this in @justcalamari!

justcalamari:
Ready for review @scopatz

PR #181, opened 2018-06-10
add coverage

CJ-Wright:


CJ-Wright:
@scopatz I guess coverage/codecov doesn't play nice with `.xsh` files? Or is there some flag I can set to cover those files?

run_tests.py
```diff
@@ -0,0 +1,17 @@
...
+
+if __name__ == '__main__':
+    # show output results from every test function
+    args = ['-v', 'tests']
+    # show the message output for skipped and expected failure tests
+    if len(sys.argv) > 1:
+        args.extend(sys.argv[1:])
+    print('pytest arguments: {}'.format(args))
+    # # compute coverage stats for xpdAcq
+    # args.extend(['--cov', 'xpdAcq'])
```
justcalamari: Can you remove references to xpdAcq?

scopatz:
@CJ-Wright - To the best of my knowledge, no.

scopatz:
But I have never really used coverage, so I am not super familiar

CJ-Wright:
It seems like there is a coverage plugin system so maybe we need a xonsh plugin.

scopatz:
Yeah that would be great!

PR #182, opened 2018-06-12
Handle conda-build 3 style multiple sources

justcalamari:


scopatz:
Thanks for putting this in @justcalamari! I like the test-to-code ratio :)

scopatz:
Looks like a test failed

justcalamari:
Ready for review @scopatz @CJ-Wright

conda_forge_tick/make_graph.py
```diff
@@ -48,10 +48,19 @@ def get_attrs(name, i, bad):
         req = set([x.split()[0] for x in req])
     sub_graph['req'] = req
 
-    keys = [('source', 'url'), ('package', 'name'),
-            ('package', 'version')]
+    keys = [('package', 'name'), ('package', 'version')]
     missing_keys = [k[1] for k in keys if k[1] not in
                     yaml_dict.get(k[0], {})]
+    source = yaml_dict.get('source', [])
+    if not isinstance(source, list):
```
scopatz: It would be preferable if this was `not isinstance(source, collections.abc.Sequence)` or `isinstance(source, str)`.  What are the types that source could come in as?

conda_forge_tick/migrators.xsh
```diff
@@ -110,31 +110,29 @@ class Version(Migrator):
...
     def find_urls(self, text):
         """Get the URLs and platforms in a meta.yaml."""
         urls = []
-        lines = text.splitlines()
-        for line in lines:
-            m = self.url_pat.match(line)
-            if m is not None:
-                urls.append((m.group(1), m.group(3)))
+        matches = self.url_pat.finditer(text)
+        for m in matches:
```
scopatz: This could/should be one line: `for m in self.url_pat.finditer(text)`

conda_forge_tick/migrators.xsh
```diff
@@ -110,31 +110,29 @@ class Version(Migrator):
...
+            urls.append((m.group(4), m.group(6), m.group()))
         matches = self.r_url_pat.finditer(text)
         for m in matches:
```
scopatz: same here, the above two lines could be combined.

conda_forge_tick/make_graph.py
msarahan: yes, isinstance is fragile.  In conda-build, I have a function called "ensure_list" that is a no-op for lists and tuples, but does this same enclosure of a string on lists.

https://github.com/conda/conda-build/blob/master/conda_build/utils.py#L978-L984

Not sure exactly how much better that is, but the hasattr might be less fragile than a purely isinstance-based approach.

conda_forge_tick/migrators.xsh
```diff
@@ -144,10 +142,17 @@ class Version(Migrator):
                         continue
             else:
                 hash = hash_url(url, hash_type)
-            p = '  {}:\s*[0-9A-Fa-f]+'.format(hash_type)
-            n = '  {}: {}'.format(hash_type, hash)
-            if platform is not None:
-                p += '\s*(#.*)\[{}\](?(1)[^\(\)]*)$'.format(platform)
+            m = re.search('\s*{}:(.+)'.format(hash_type), line)
+            if m is not None:
```
scopatz: It is generally eaiser to follow positive logical expressions.  I would recommend swapping this to be `if m is None`

conda_forge_tick/update_upstream_versions.py
```diff
@@ -142,7 +142,14 @@ def get_url(self, meta_yaml):
             for next_ver in next_version(current_ver):
                 new_content = content.replace(orig_ver, next_ver)
                 meta = parse_meta_yaml(new_content)
-                url = str(meta['source'].get('url', None))
+                source = meta['source']
+                if not isinstance(source, list):
```
scopatz: same type checking issue here.

scopatz:
This looks good @justcalamari - just a couple of stylistic changes that I think would improve readability.

conda_forge_tick/make_graph.py
justcalamari: Here source will either be a dict or a list, not a string. I don't think `ensure_list` will work in this case.

msarahan: In light of that, I think you want

```
if hasattr(source, "keys"):
```

scopatz: actually, I think in that case you want to check if it is a mapping:

`if isinstance(source, collections.abc.Mapping):`

msarahan: Yes, that's a stronger check.

justcalamari:
@scopatz Nope. Feel free to merge.

PR #187, opened 2018-06-17
Break up travis jobs

justcalamari:


conda_forge_tick/make_graph.py
```diff
@@ -57,8 +57,9 @@ def get_attrs(name, i, bad):
         source = [source]
     source_keys = set()
     for s in source:
-        if not sub_graph.get('url'):
-            sub_graph['url'] = s.get('url')
+        if s.get('url'):
+            sub_graph['url'] = s['url']
+            break
```
CJ-Wright: Ah is that the problem spot?

00.xsh
```diff
@@ -3,10 +3,8 @@ import time
 cd ../cf-graph
 $PATH.insert(0, '~/mc/bin')
 
-stages = [3, 0, 1, 2, 3]
+stage = 0
 start = time.time()
-for i in stages:
```
CJ-Wright: What was the issue with this loop?

conda_forge_tick/make_graph.py
justcalamari: Nope, that's actually a mistake. I thought this might help but then realized it would break other things and forgot to change it back.

00.xsh
justcalamari: Nothing, but I was setting it to run only one stage for each worker. Maybe I should leave the loop but change the `stages` list to only have one element in each script.

CJ-Wright: Yea is rather just cut the list down since this is hopefully a temporary debug change.

justcalamari: I think I will close this PR and open a new one with the changes since everything in this commit is going to be undone anyway. Let me know if you agree or if you think I should just add commits here.

justcalamari: Also I realized make-cf-graph runs 03.xsh and cf-auto-tick runs 01.xsh, but I think those should be swapped.

CJ-Wright: Whichever you prefer.

PR #188, opened 2018-06-17
Break up travis jobs

justcalamari:


justcalamari:
Is something going to run at 12? We probably don't want multiple workers running simultaneously.

PR #205, opened 2018-06-23
Remove bad and upstream_bad files plus some stylistic changes

### LINKED ISSUE #56: restart bad and bad upstream
As we make enhancements to the code we should consider cleaning out bad and bad upstream. That way things that we couldn't process we can get another shot at.

### LINKED ISSUE #176: Don't run update_upstream_versions on archived feedstocks
If a repo is archived we don't need to check it's upstream versions.

justcalamari:
Closes #56
Closes #176

### DIFF
--- a/conda_forge_tick/make_graph.py
+++ b/conda_forge_tick/make_graph.py
@@ -30,7 +30,7 @@ def get_attrs(name, i):
             "Something odd happened when fetching recipe "
             "{}: {}".format(name, r.status_code)
         )
-        sub_graph["bad"] = r.status_code
+        sub_graph["bad"] = "make_graph: {}".format(r.status_code)
         return sub_graph
 
     text = r.content.decode("utf-8")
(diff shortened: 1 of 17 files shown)


conda_forge_tick/make_graph.py
```diff
@@ -11,119 +11,110 @@
...
     if r.status_code != 200:
-        logger.warn('Something odd happened when fetching recipe '
-                    '{}: {}'.format(name, r.status_code))
-        bad.append(name)
-        sub_graph['bad'] = r.status_code
+        logger.warn(
+            "Something odd happened when fetching recipe "
+            "{}: {}".format(name, r.status_code)
+        )
+        sub_graph["bad"] = r.status_code
```
CJ-Wright: I liked what you did with the update bad, would it be possible to do something similar here? eg. `'make_graph: {}'.format(r.status_code)`

CJ-Wright:
Thank you @justcalamari

CJ-Wright:
I guess this means we can remove bad, upstream_bad, exceptions, et al.

PR #216, opened 2018-06-27
fix for tbb

CJ-Wright:
from @isuruf


scopatz:
This seems reasonable.

PR #217, opened 2018-06-28
Fix multi source upstream

justcalamari:


justcalamari:
Yes. The problem was that your last PR ignored that `meta_yaml["source"]` might be a list instead of a dict.

CJ-Wright:
Ok, I'll merge, although we could come back for the docs/tests.

PR #219, opened 2018-06-28
say which migrator is doing the migration

CJ-Wright:


justcalamari:
Can you either make the "IS MIGRATING" lowercase or the migrator name uppercase? Not a big deal, but I think it will look better.

CJ-Wright:
I know the caps locks is a bit shouty, but our logs are noisier than a rave inside a jet engine, so it might be ok.

PR #223, opened 2018-06-28
FIX: Use correct dicts when getting urls

justcalamari:


scopatz:
LGTM

scopatz:
Thanks @justcalamari

PR #224, opened 2018-06-28
Get raw meta yaml from graph

justcalamari:


scopatz:
Thanks @justcalamari!

PR #225, opened 2018-06-29
Update in topological order

justcalamari:
Closes #225


scopatz:
Thanks for putting this in @justcalamari! My main extra thought here is that it might be nice to have migrators select whether they need topo sorting or now.  However, for migrators that don't care, it certainly doesn't hurt to have them executed in topo order too.  So I think this is fine as is.

PR #231, opened 2018-06-29
Clean /tmp after each migration

justcalamari:


conda_forge_tick/auto_tick.xsh
```diff
@@ -184,6 +185,9 @@ def main(args=None):
                 rm -rf $REVER_DIR + '/*'
                 print(![pwd])
                 ![doctr deploy --token --built-docs . --deploy-repo regro/cf-graph --deploy-branch-name master .]
+                for f in g`/tmp/*`:
+                    if f not in temp:
+                        rm -rf f
```
CJ-Wright: I think the f here needs to be `@(f)`

PR #237, opened 2018-07-02
start noarch and have migrator based pr limits, also force rerender

CJ-Wright:
for noarch migration

### DIFF
--- a/conda_forge_tick/auto_tick.xsh
+++ b/conda_forge_tick/auto_tick.xsh
@@ -18,9 +18,9 @@ from .path_lengths import cyclic_topological_sort
 # https://travis-ci.org/regro/00-find-feedstocks/jobs/388387895#L1870
 from .migrators import *
 $MIGRATORS = [
-    # Version(),
-    # Compiler(),
-    Noarch(),
+    Version(pr_limit=0),
+    Compiler(pr_limit=0),
+    Noarch(pr_limit=1),
 ]
 
 def run(attrs, migrator, feedstock=None, protocol='ssh',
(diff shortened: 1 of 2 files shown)


CJ-Wright:
I'd hold off on this

conda_forge_tick/auto_tick.xsh
```diff
@@ -17,10 +17,9 @@ from .path_lengths import cyclic_topological_sort
 # TODO: move this back to the bot file as soon as the source issue is sorted
 # https://travis-ci.org/regro/00-find-feedstocks/jobs/388387895#L1870
 from .migrators import *
-$MIGRATORS = [Version(),
-              Compiler(),
-              #Noarch(),
-             ]
+$MIGRATORS = {Version(): 5,
```
scopatz: It would be nice if there was a comment above this line that says that value is the PR limit.  A better model would be if the instances took a PR limit, ie `Version(pr_limit=5)` and this returned to a list.

CJ-Wright: Do you want the migrators to do the PRing themselves? Currently the migrators don't have any PRing functionality.

CJ-Wright: I guess we could request the attribute during the loop.

scopatz: Yeah, that was the idea.  The migrators can store the data about how they are to be used.

CJ-Wright:
@scopatz I think this is ready for another round of review.

CJ-Wright:
Thank you!

PR #238, opened 2018-07-03
use futures to speed up upstream (maybe)

CJ-Wright:
@mariusvniekerk I took some inspiration from libcflib's use of futures to try to speed up upstream tracking.

### DIFF
--- a/conda_forge_tick/make_graph.py
+++ b/conda_forge_tick/make_graph.py
@@ -14,72 +14,67 @@ logger = logging.getLogger("conda_forge_tick.make_graph")
 
 
 def get_attrs(name, i):
-    try:
-        sub_graph = {
-            "time": time.time(),
-            "feedstock_name": name,
-            # All feedstocks start out as good
-            "bad": False,
-        }
+    sub_graph = {
+        "time": time.time(),
+        "feedstock_name": name,
+        # All feedstocks start out as good
+        "bad": False,
+    }
 
-        logger.info((i, name))
-        r = requests.get(
-            "https://raw.githubusercontent.com/"
-            "conda-forge/{}-feedstock/master/recipe/"
-            "meta.yaml".format(name)
+    logger.info((i, name))
+    r = requests.get(
+        "https://raw.githubusercontent.com/"
+        "conda-forge/{}-feedstock/master/recipe/"
+        "meta.yaml".format(name)
+    )
+    if r.status_code != 200:
+        logger.warn(
+            "Something odd happened when fetching recipe "
+            "{}: {}".format(name, r.status_code)
         )
-        if r.status_code != 200:
-            logger.warn(
-                "Something odd happened when fetching recipe "
-                "{}: {}".format(name, r.status_code)
-            )
-            sub_graph["bad"] = "make_graph: {}".format(r.status_code)
-            return sub_graph
+        sub_graph["bad"] = "make_graph: {}".format(r.status_code)
+        return sub_graph
 
-        text = r.content.decode("utf-8")
-        sub_graph["raw_meta_yaml"] = text
-        yaml_dict = parse_meta_yaml(text)
-        if not yaml_dict:
-            logger.warn("Something odd happened when parsing recipe " "{}".format(name))
-            sub_graph["bad"] = "make_graph: Could not parse"
-            return sub_graph
-        sub_graph["meta_yaml"] = yaml_dict
-        # TODO: Write schema for dict
-        req = yaml_dict.get("requirements", set())
-        if req:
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
Do you have some rough guess as to what speedup you got?

conda_forge_tick/update_upstream_versions.py
```diff
@@ -187,38 +188,45 @@ def get_version(self, url):
...
-            )
+    futures = []
+    with ThreadPoolExecutor(max_workers=20) as pool:
+        for node, attrs in gx.node.items():
+            if attrs.get("bad") or attrs.get("archived"):
+                attrs["new_version"] = False
+                continue
+            futures.append(pool.submit(get_latest_version(attrs, sources)))
+        for f in as_completed(futures):
+            node, attrs = f.result()
```
justcalamari: I think you can put the try here instead of inside the functions.

CJ-Wright: Will I have access to the name for the logger?

CJ-Wright:
No but it seems that we are stalling a little on the getting upstream versions so maybe the request is slow?

conda_forge_tick/make_graph.py
```diff
@@ -12,68 +14,72 @@
...
-        "https://raw.githubusercontent.com/"
-        "conda-forge/{}-feedstock/master/recipe/"
-        "meta.yaml".format(name)
-    )
-    if r.status_code != 200:
-        logger.warn(
-            "Something odd happened when fetching recipe "
-            "{}: {}".format(name, r.status_code)
+    try:
+        sub_graph = {
```
scopatz: try-blocks should be as minimal as possible. Could you maybe move this to a helper function please? Or become more targeted in where the try-excepts are. Thanks!

conda_forge_tick/update_upstream_versions.py
```diff
@@ -187,38 +188,45 @@ def get_version(self, url):
...
-        else:
-            meta_yaml["bad"] = "Upstream: Could not find version on {}".format(
-                source.name
-            )
-    if not meta_yaml.get("bad"):
-        meta_yaml["bad"] = "Upstream: unknown source"
-    return False
+def get_latest_version(name, meta_yaml, sources):
+    try:
+        for source in sources:
```
scopatz: Same comment about large try-blocks as above.

CJ-Wright:
ready for more review!

conda_forge_tick/make_graph.py
```diff
@@ -89,16 +90,20 @@ def make_graph(names, gx=None):
...
-        except Exception as e:
-            logger.warn("Error adding {} to the graph: {}".format(name, e))
-        else:
-            if name in new_names:
-                gx.add_node(name, **sub_graph)
+    with ThreadPoolExecutor(max_workers=20) as pool:
+        futures = {pool.submit(get_attrs(name, i)): name for i, name in enumerate(total_names)}
+
+        for f in as_completed(futures):
+            name = futures[f]
```
justcalamari: What is this line doing?

CJ-Wright: This is getting the name of the node who's future has been completed. This is needed for the logger/exception so that we know which node failed.

conda_forge_tick/update_upstream_versions.py
```diff
@@ -187,38 +187,46 @@ def get_version(self, url):
...
+    with ThreadPoolExecutor(max_workers=20) as pool:
+        for node, attrs in gx.node.items():
+            if attrs.get("bad") or attrs.get("archived"):
+                attrs["new_version"] = False
+                continue
+            futures.update({pool.submit(get_latest_version(attrs, sources)): (node, attrs)})
+        for f in as_completed(futures):
+            node, attrs = futures[f]
+            try:
+                f.result()
```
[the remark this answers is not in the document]
CJ-Wright: Actually we are getting the node and attrs from the futures dict, so I don't know if we need that here.
I borrowed the pattern from https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example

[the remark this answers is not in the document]
justcalamari: Ok. Then should `get_latest_version` be returning anything?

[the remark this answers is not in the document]
justcalamari: Maybe leave `get_latest_version` as returning the version and have it called by another function that updates the attrs.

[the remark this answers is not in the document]
CJ-Wright: We can have the update trigger in the future loop.

CJ-Wright:
Thank you!

PR #260, opened 2018-07-09
Remove version pinning for github3.py

justcalamari:
The bot failed on its most recent run, which I believe is due to pinning an old version of github3.py.


CJ-Wright:
Yes that was from a bad merge for that package

PR #256, opened 2018-07-09
Add more detailed logger setup

mariusvniekerk:

PR #267, opened 2018-07-10
Revert "Add more detailed logger setup"

CJ-Wright:
Reverts regro/cf-scripts#256
This is causing some issues with formatting https://circleci.com/gh/regro/circle_worker/189


scopatz:
I think it is better to fix than revert

CJ-Wright:
@mariusvniekerk asked for a revert, either way works for me!

PR #285, opened 2018-07-20
Remove pinning migrator

### LINKED ISSUE #283: perl migrator
Many old recipes that use `perl` have explicit `perl` version pinnings (because of an old `conda-build` bug), which causes problems with new recipes that try to use them: e.g. https://github.com/conda-forge/perl-xml-parser-feedstock/pull/1 / https://github.com/conda-forge/staged-recipes/pull/4858#discussion_r203149590. It'd be nice to migrate those to not do that.

(I haven't checked how many there are; maybe there's few enough that they should be done explicitly.)

justcalamari:
closes #283

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -7,6 +7,15 @@ from conda_build.config import Config
 from conda_build.metadata import parse
 
 
+class UniversalSet(set):
+    """The universal set, or identity of the set intersection operation."""
+
+    def __add__(self, other):
+        return other
+
+    def __radd__(self, other):
+        return other
+
 class NullUndefined(jinja2.Undefined):
     def __unicode__(self):
         return self._undefined_name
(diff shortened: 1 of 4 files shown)


conda_forge_tick/utils.py
```diff
@@ -7,6 +7,15 @@
 from conda_build.metadata import parse
 
 
+class UniversalSet(set):
```
CJ-Wright: Why is this needed? (Not saying it's wrong, I just can't see it yet)

justcalamari: This is the default value for requirements we remove pinnings on. If no specific packages are provided, just remove the pinnings on everything. Since the filter and migrator use set intersection, this is a nice way to make it a no op.

scopatz: Also, you should not subclass from set, but rather from collections.abc.Set

CJ-Wright:
Is this ok to go in @justcalamari ?

justcalamari:
Yes, @CJ-Wright

CJ-Wright:
I'll merge first thing tomorrow.

CJ-Wright:

PR #290, opened 2018-07-25
better pin filter

### LINKED ISSUE #289: Make pinning removal a bit more precise
Would it be possible to have the pinning filter remove things which aren't pinned? Otherwise we're going to run on a lot of things.
Attn: @justcalamari

CJ-Wright:
Closes #289

### DIFF
--- a/conda_forge_tick/migrators.xsh
+++ b/conda_forge_tick/migrators.xsh
@@ -601,9 +601,11 @@ class Pinning(Migrator):
                 m = p.match(line)
                 if m is not None:
                     lines[i] = lines[i].replace(m.group(1), "")
+                    removed_version = m.group(1).strip()
                     if not n:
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@justcalamari

justcalamari:
My problem with this is that now filter is doing the same stuff that migrate is doing, so for any feedstock with pinnings, we would be running the same code twice. It might be better to just check in migrate if there is a pinning and if not return False.

justcalamari:
Also we should probably bump the build number in migrate.

CJ-Wright:
@justcalamari I think this is ready for another round.

justcalamari:
Thanks, @CJ-Wright!

CJ-Wright:
Thank you!

PR #296, opened 2018-07-27
Add npm source

### LINKED ISSUE #291: [idea] Filtering out GitHub/npm prereleases
Greetings :robot: overlords! Thanks for all the helpful release updates!

On https://github.com/conda-forge/yarn-feedstock, we're getting a lot of false positives from https://github.com/yarnpkg/yarn. I assume the root cause is that project doesn't use distinguishing features like `rc` in their prereleases tags, which i guess is common in npm land :man_shrugging: . I actually merged the bot's `1.8.0` PR, which was also a dev release, as I didn't think to check. I'd like to avoid that in the future. Nobody has complained yet, I guess, so I guess we don't need to take it down!

> Note: Dev issues aside, we're presently doing a rather non-standard thing in the feedstock, just copying in the release contents directly into `bin` and `lib`, which is actually kinda broken on older windows `conda`s, but that aside, it would be nice to do the right thing to get it sorted out properly.

So the question: is there a...
- URL scheme we could use that would already carry the metadata?
- way to specify our install that would trigger using npm (if this, then we'll just do it!)
- new work to try and track this down

Here are some leads i tracked down:
- github:
  - i assume github api calls are precious. the best summary of approaches i've seen is something [along these lines](https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c), including some HTTP options that tracks the `302` of `https://github.com/yarnpkg/yarn/releases/latest`
- libraries.io
  - `https://libraries.io/api/github/yarnpkg/yarn/projects` says the `latest_stable_release` is `1.9.2`, as does the atom feed. haven't PRd there, as they definitely do distinguish between stable and pre, but i had trouble getting the dev env up (not a ruby guy)
- npm
  - `http://registry.npmjs.org/yarn` reports `"dist-tags":{"latest":"1.7.0","rc":"1.9.2"}`
  - `http://registry.npmjs.org/yarn/latest`

Thoughts? Happy to follow up one of those leads, if they sound fruitful...

bollwyvl:
Fixes #291.

This adds a simple, yet [not very widely applicable](https://github.com/search?q=org%3Aconda-forge+registry.npmjs.org&type=Code), source to, if the feedstock url is from `registry.npmjs.org`, make a call to `registry.npmjs.org` and return the `dist-tag` which is `latest` (the default). Like the PyPI source, it also will not return pre-release-y values in the name of the tag, as that also happens as the `dist-tags` feature is fairly buried. 

Not doing prereleases might be introducing an unwelcome opinion... thoughts? Both of the major upstream tools that use the namespace, `npm` and `yarn`, ignore them, so it seems reasonable enough.

### DIFF
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -114,6 +114,26 @@ class PyPI:
         return r.json()["info"]["version"].strip()
 
 
+class NPM:
+    name = "npm"
+
+    def get_url(self, meta_yaml):
+        if "registry.npmjs.org" in meta_yaml["url"]:
+            return None
+        # might be namespaced
+        pkg = meta_yaml["url"].split("/")[3:-2]
+        return "https://registry.npmjs.org/{}".format("/".join(pkg))
+
+    def get_version(self, url):
+        r = requests.get(url)
+        latest = r.json()["dist-tags"].get("latest", "").strip()
+        # If it is a pre-release don't give back the pre-release version
+        if len(latest) and parse_version(latest).is_prerelease:
+            return False
+
+        return latest
+
+
 class CRAN(LibrariesIO):
     name = "cran"
     url_contains = "cran.r-project.org/src/contrib/Archive"
@@ -204,7 +224,7 @@ def get_latest_version(meta_yaml, sources):
     return False
 
 
-def update_upstream_versions(gx, sources=(PyPI(), CRAN(), RawURL(), Github())):
+def update_upstream_versions(gx, sources=(PyPI(), NPM(), CRAN(), RawURL(), Github())):
     futures = {}
     with ProcessPoolExecutor(max_workers=20) as pool:
         for node, attrs in gx.node.items():
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -118,7 +118,7 @@ class NPM:
     name = "npm"
 
     def get_url(self, meta_yaml):
-        if "registry.npmjs.org" in meta_yaml["url"]:
+        if "registry.npmjs.org" not in meta_yaml["url"]:
             return None
         # might be namespaced
         pkg = meta_yaml["url"].split("/")[3:-2]
@@ -126,9 +126,11 @@ class NPM:
 
     def get_version(self, url):
         r = requests.get(url)
+        if not r.ok:
+            return False
         latest = r.json()["dist-tags"].get("latest", "").strip()
         # If it is a pre-release don't give back the pre-release version
-        if len(latest) and parse_version(latest).is_prerelease:
+        if not len(latest) or parse_version(latest).is_prerelease:
             return False
 
         return latest
--- a/tests/test_upstream_versions.py
+++ b/tests/test_upstream_versions.py
@@ -1,0 +1,162 @@
+import pytest
+
+from conda_forge_tick.update_upstream_versions import NPM, get_latest_version
+from conda_forge_tick.utils import parse_meta_yaml
+
+
+sample_npm = """
+{% set name = "configurable-http-proxy" %}
+{% set version = "3.1.0" %}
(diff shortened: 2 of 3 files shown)


conda_forge_tick/update_upstream_versions.py
```diff
@@ -114,6 +114,26 @@ def get_version(self, url):
         return r.json()["info"]["version"].strip()
 
 
+class NPM:
+    name = "npm"
+
+    def get_url(self, meta_yaml):
+        if "registry.npmjs.org" in meta_yaml["url"]:
```
justcalamari: Should this be `not in`?

conda_forge_tick/update_upstream_versions.py
```diff
@@ -114,6 +114,26 @@ def get_version(self, url):
...
+            return None
+        # might be namespaced
+        pkg = meta_yaml["url"].split("/")[3:-2]
+        return "https://registry.npmjs.org/{}".format("/".join(pkg))
+
+    def get_version(self, url):
+        r = requests.get(url)
+        latest = r.json()["dist-tags"].get("latest", "").strip()
```
justcalamari: Can you please check `r.ok` before trying to get the json?

justcalamari:
Thanks for putting this in, @bollwyvl!

isuruf:
Instead of checking the url to have `registry.npmjs.org` you can check whether the meta.yaml file has `- nodejs` in it. `meta_yaml['raw_meta_yaml']` gives the raw file

conda_forge_tick/update_upstream_versions.py
bollwyvl: ack! good catch!

conda_forge_tick/update_upstream_versions.py
bollwyvl: fixed!

bollwyvl:
> Instead of checking the url to have registry.npmjs.org you can check whether the meta.yaml file has - nodejs in it. meta_yaml['raw_meta_yaml'] gives the raw file

Hmmm... I guess the namespace thing is what trips me up from taking that approach. A separate PR could look to doing something about the github release filtering as well, but i have no data outside of a bit of looking as to which projects are aggressively using that particular ui feature... i guess if they _ever_ used a pre-release, then you'd only want the stable releases.

scopatz:
Thanks for putting this in @bollwyvl! Generally LGTM

scopatz:
It would be nice to have tests for this, but we don't have any other tests for related classes, so I don't feel comfortable putting that burden on this PR. But it would make a nice addition :wink:

bollwyvl:
Oh no worries, tests sound good!

CJ-Wright:
I'm :+1: on these changes. Tests would be spectacular!

bollwyvl:
Added a simple test with requests-mock.

I kinda phoned it in on the `attrs`, which probably need to be fixed.

tests/test_upstream_versions.py
```diff
@@ -0,0 +1,162 @@
...
+        sample_npm, "3.1.1", NPM(), {
+            "https://registry.npmjs.org/configurable-http-proxy": sample_npm_response
+        }
+    ),
+]
+
+
+@pytest.mark.parametrize("inp, ver, source, urls", latest_url_test_list)
+def test_latest_version(inp, ver, source, urls, requests_mock):
+    pmy = parse_meta_yaml(inp)["source"]
```
bollwyvl: What's the right way to transform the package into the thing that `get_latest_version` is expecting?

CJ-Wright: You could use `get_attrs` https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/make_graph.py#L24 (although this currently pulls the `meta.yaml` itself from github. I guess at some point we should separate the request from the rest of it for these kinds of things.

bollwyvl: Hm, guess I can mock that, too...

CJ-Wright:
@bollwyvl thank you very much!

PR #297, opened 2018-07-27
Set PRed for recipes with no pinning

justcalamari:


CJ-Wright:
I think we should revisit if we should filter these no pinning recipes out. We seem to be patching infrastructure to patch infrastructure because we may not be doing the right thing.

PR #309, opened 2018-08-02
Log exception info in auto_tick

justcalamari:


conda_forge_tick/auto_tick.xsh
```diff
@@ -186,8 +186,7 @@ def main(args=None):
                     if is_github_api_limit_reached(e, gh):
                         break
             except Exception as e:
-                logger.critical('NON GITHUB ERROR')
-                logger.critical(str(e))
+                logger.exception('NON GITHUB ERROR')
```
scopatz: aren't you missing the actual error message here?

justcalamari: My understanding of the `exception` method is that it includes the traceback in the logger message.

CJ-Wright: Will this also include the standard error for CLI tools?

scopatz: Oh you are right!

CJ-Wright:
@justcalamari would it be possible to add the standard error to this? I don't expect the `update-cb3` command to report it's traceback inside the xonsh process so this won't capture it. I guess we could use the python side and not the CLI?

PR #310, opened 2018-08-02
print update-cb3 errors

CJ-Wright:


justcalamari:
I just made a PR to switch syntax migration to use python functions in #311. Should be accomplishing the same thing, but makes it more consistent with other migrators.

PR #331, opened 2018-08-22
use frozendict for migrator_uid.

CJ-Wright:
This simplifies a bunch of really annoying syntax and allows us to get
the correct PRed_json every time

### DIFF
--- a/conda_forge_tick/migrators.xsh
+++ b/conda_forge_tick/migrators.xsh
@@ -286,7 +286,7 @@ class Version(Migrator):
 
     def migrator_uid(self, attrs):
         n = super().migrator_uid(attrs)
-        n = n.copy(**{'version': attrs["new_version"]})
+        n = n.copy(version=attrs["new_version"])
         return n
 
     def _extract_version_from_hash(self, h):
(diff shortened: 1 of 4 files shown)


CJ-Wright:
Need to move from update to `.copy`

scopatz:
Code looks fine but it is failing CI for some reason

CJ-Wright:
I need to use copy rather than update because frozen dict Arya immutable. Skidoo I need to coordinate this PR with changes to the graph so I'll halt ops while that happens.

conda_forge_tick/migrators.xsh
```diff
@@ -282,7 +286,7 @@ class Version(Migrator):
 
     def migrator_uid(self, attrs):
         n = super().migrator_uid(attrs)
-        n.update({'version': attrs["new_version"]})
+        n = n.copy(**{'version': attrs["new_version"]})
```
scopatz: Wouldn't it be cleaner to do `n.copy(version=attrs["new_version"])`

conda_forge_tick/migrators.xsh
```diff
@@ -612,7 +616,7 @@ class Rebuild(Migrator):
 
     def migrator_uid(self, attrs):
         n = super().migrator_uid(attrs)
-        n.update({'name': self.name})
+        n = n.copy(**{'name': self.name})
```
scopatz: same here, isn't this just `n.copy(name=self.name)`

CJ-Wright:
ping @scopatz

CJ-Wright:
Thank you

PR #332, opened 2018-08-22
lazy json so we can store all the json elsewhere but still interface

CJ-Wright:
with it like normal

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -1,6 +1,5 @@
 import os
 from collections import defaultdict
-from typing import _KT, Iterator, _T_co
 
 from collections.abc import Set, MutableMapping
 import logging
@@ -8,8 +7,6 @@ import itertools
 import json
 
 import jinja2
-from conda_build.config import Config
-from conda_build.metadata import parse
 
 
 class UniversalSet(Set):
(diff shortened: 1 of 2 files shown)


conda_forge_tick/utils.py
```diff
@@ -42,6 +42,56 @@ def __getitem__(self, name):
...
+                self.data = json.load(f)
+
+    def _dump(self):
+        self._load()
+        with open(self.file_name, "w") as f:
+            json.dump(self.data, f)
+
+    def __getitem__(self, item):
+        self._load()
+        return self.data.__getitem__(item)
```
scopatz: This should be `return self.data[item]`

conda_forge_tick/utils.py
```diff
@@ -42,6 +42,56 @@ def __getitem__(self, name):
...
+
+    def __setitem__(self, key, value):
+        self._load()
+        self.data.__setitem__(key, value)
```
scopatz: Should be `self.data[key] = value`

CJ-Wright:
ping @scopatz

conda_forge_tick/utils.py
```diff
@@ -42,6 +42,56 @@ def __getitem__(self, name):
...
+                json.dump({}, f)
+        self.data = None
+
+    def __len__(self) -> int:
+        self._load()
+        return len(self.data)
+
+    def __iter__(self):
+        self._load()
+        return self.data.__iter__
```
scopatz: Shouldn't this be a `yield from self.data`?

CJ-Wright:
one more round? @scopatz

conda_forge_tick/utils.py
```diff
@@ -42,6 +42,56 @@ def __getitem__(self, name):
...
+                json.dump({}, f)
+        self.data = None
+
+    def __len__(self) -> int:
+        self._load()
+        return len(self.data)
+
+    def __iter__(self):
+        self._load()
+        yield from self.data.__iter__
```
scopatz: This should just be `yield from self.data`

CJ-Wright:
Thank you!

PR #341, opened 2018-09-04
better graph for rebuild based of host

CJ-Wright:


CJ-Wright:

PR #344, opened 2018-09-05
split requirements from pins and use host

CJ-Wright:


scopatz:
Seems good to me, feel free to merge after CI

PR #347, opened 2018-09-07
use chaindb to clobber os requirements together

CJ-Wright:


CJ-Wright:
ping @regro/auto-tick

CJ-Wright:
@scopatz thoughts?

conda_forge_tick/chained_db.py
```diff
@@ -0,0 +1,72 @@
...
+
+    __inst = None
+
+    def __new__(cls):
+        if ChainDBSingleton.__inst is None:
+            ChainDBSingleton.__inst = object.__new__(cls)
+        return ChainDBSingleton.__inst
+
+
+Singleton = ChainDBSingleton()
```
scopatz: I don't like calling this a Singleton.  I would recommend following what None and NoneType do.  So this would be something like:

```
ChainDBDefault = ChainDBDefaultType()
```

conda_forge_tick/chained_db.py
```diff
@@ -0,0 +1,72 @@
...
+Copyright 2015-2016, the xonsh developers
+"""
+
+import itertools
+
+from collections import ChainMap
+from collections.abc import MutableMapping
+
+
+class ChainDBSingleton(object):
```
scopatz: And calling this ChainDBDefaultType

conda_forge_tick/chained_db.py
```diff
@@ -0,0 +1,72 @@
...
+        for mapping in self.maps:
+            results.append(mapping.get(key, Singleton))
+        # if all the results are mapping create a ChainDB
+        if all([isinstance(result, MutableMapping) for result in results]):
+            for result in results:
+                if res is None:
+                    res = ChainDB(result)
+                else:
+                    res.maps.append(result)
+        elif all([isinstance(result, list) for result in results]):
```
scopatz: Should this really be list, and not non-string sequence?

conda_forge_tick/chained_db.py
scopatz: Also, is this something that should go into `xonsh.lib`?

conda_forge_tick/make_graph.py
```diff
@@ -45,12 +46,15 @@ def get_attrs(name, i):
 
     text = r.content.decode("utf-8")
     sub_graph["raw_meta_yaml"] = text
-    yaml_dict = parse_meta_yaml(text)
+    yaml_dict = ChainDB(*[parse_meta_yaml(text, arch=arch) for arch in
+                       ['win', 'osx', 'linux']])
+    # yaml_dict = parse_meta_yaml(text)
```
scopatz: please remove commented out code

conda_forge_tick/chained_db.py
CJ-Wright: This is code vendorized from regolith. It could go into `xonsh.lib`

conda_forge_tick/chained_db.py
CJ-Wright: I think the main issue with this is that tuples might not make sense when extended.

conda_forge_tick/chained_db.py
scopatz: Right, so rather than copy it up the stack, we could copy it down :)

conda_forge_tick/chained_db.py
scopatz: Ahh I see ok. So really, you could test against MutableSequence and be fine (since strings are immutable)

conda_forge_tick/chained_db.py
CJ-Wright: xref: https://github.com/xonsh/xonsh/pull/2809

CJ-Wright:
@scopatz can I get another round?

conda_forge_tick/make_graph.py
```diff
@@ -7,12 +7,17 @@
...
+    ProcessPoolExecutor,
+    as_completed,
+    ThreadPoolExecutor,
+)
 
 import github3
 import networkx as nx
 import requests
 
+from xonsh.lib import ChainDB, _convert_to_dict
```
scopatz: shouldn't this be from `xonsh.lib.collections`?

conda_forge_tick/make_graph.py
```diff
@@ -45,20 +50,40 @@ def get_attrs(name, i):
 
     text = r.content.decode("utf-8")
     sub_graph["raw_meta_yaml"] = text
-    yaml_dict = parse_meta_yaml(text)
+    yaml_dict = ChainDB(
+        *[
+            parse_meta_yaml(text, arch=arch)
+            for arch in [
+                # 'win',
```
scopatz: Please remove dead code

CJ-Wright:
I think needs to wait while I get branch deletion online

PR #351, opened 2018-09-12
delete only on merge

CJ-Wright:


conda_forge_tick/git_utils.xsh
```diff
@@ -155,7 +155,7 @@ def refresh_pr(pr_json: LazyJson, gh=None):
         pr_obj.refresh()
         pr_obj_d = pr_obj.as_dict()
         # if state passed from opened to closed delete the branch
-        if pr_obj_d['state'] == 'closed':
+        if pr_obj_d['state'] == 'closed' and pr_obj_d.get('merged_at', None):
```
scopatz: Wouldn't this make more sense as False, rather than None?

CJ-Wright: Done

PR #353, opened 2018-09-13
run rebuild migrations in number of decendents order

CJ-Wright:
ping @regro/auto-tick


mariusvniekerk:
I think before we can consider changing this we need to have a comparison of the build order.

This may have some fairly negative impact for chains that could be finished fairly quickly.

CJ-Wright:
tl;dr This PR would cause packages with more dependencies to be rebuilt first, assuming that the package was ready to be rebuilt. This would allow us to get deeper into the graph faster, but would leave some packages (the ones with less children) waiting longer.

If we had download numbers we could do it by downloads :smile:

mariusvniekerk:
yeah seems okay to me since the toposort for the current layer is effectively arbitrary since they are all flat

PR #362, opened 2018-09-16
Revert "add run and test deps to graph solving"

CJ-Wright:
Reverts regro/cf-scripts#356


mariusvniekerk:
why?

CJ-Wright:
It caused keras to be built. It thought keras was in a cycle.

CJ-Wright:
Need to remove cycle deps from requirements prior to edge re-computation

PR #365, opened 2018-09-22
Add build and run to requirements again?

mariusvniekerk:
cc @scopatz


scopatz:
Do we need this to find cycles?

scopatz:
hmmm it couldn't find `graph.pkl`

mariusvniekerk:
Looks pretty good for cycles.  hyperspy has bad self cycles in those packages.

PR #366, opened 2018-09-23
use total graph for measuring decendents

CJ-Wright:
This properly sorts by the number of decendents.
This took 5 mins to write, don't worry I'm still studying.


CJ-Wright:
@scopatz @mariusvniekerk

PR #370, opened 2018-10-02
Update bump_build_number to have a class defined callback.

mariusvniekerk:
This allows us to easily control the behavior of build_number bumping

### DIFF
--- a/conda_forge_tick/migrators.xsh
+++ b/conda_forge_tick/migrators.xsh
@@ -33,6 +33,14 @@ class Migrator:
 
     migrator_version = 0
 
+    build_patterns = ((re.compile('(\s*?)number:\s*([0-9]+)'),
+                       'number: {}'),
+                      (re.compile('(\s*?){%\s*set build_number\s*=\s*"?([0-9]+)"?\s*%}'),
+                       '{{% set build_number = {} %}}'),
+                      (re.compile('(\s*?){%\s*set build\s*=\s*"?([0-9]+)"?\s*%}'),
+                       '{{% set build = {} %}}')
+                     )
+
     def __init__(self, pr_limit=0):
         self.pr_limit = pr_limit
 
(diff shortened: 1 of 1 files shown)


conda_forge_tick/migrators.xsh
```diff
@@ -321,6 +365,13 @@ class Version(Migrator):
     def _extract_version_from_hash(self, h):
         return h.get('version', '0.0.0')
 
+    @classmethod
+    def new_build_number(cls, old_build_number: int):
+        if old_build_number > 1000:
+            return 1000
```
ocefpaf: :+1:

conda_forge_tick/migrators.xsh
```diff
@@ -724,7 +744,7 @@ class Pinning(Migrator):
         upd = "\n".join(lines) + "\n"
         with open(os.path.join(recipe_dir, "meta.yaml"), "w") as f:
             f.write(upd)
-        Rebuild.bump_build_number(os.path.join(recipe_dir, "meta.yaml"))
+        self.set_build_number(os.path.join(recipe_dir, "meta.yaml"))
```
CJ-Wright: I think this will have the wrong behavior. This will set the build number to 1000 for the rebuilds, when it needs to be set to the current build number + 1000

scopatz: Oh you are right, this has an extra `cls.bump_number` in it.  @mariusvniekerk - would you mind putting in a fix please?

mariusvniekerk: Err that should not set it to 1000, the set to 1000 bit only applies to the Version migrator

conda_forge_tick/migrators.xsh
```diff
@@ -151,6 +159,47 @@ class Migrator:
...
+                if m is not None:
+                    old_build_number = int(m.group(2))
+                    new_build_number = cls.new_build_number(old_build_number)
+                    lines[i] = m.group(1) + n.format(new_build_number)
+            upd = '\n'.join(lines) + '\n'
+            with open(filename, 'w') as f:
+                f.write(upd)
+
+    @classmethod
+    def new_build_number(cls, old_number: int):
```
mariusvniekerk: @CJ-Wright this is the default behavior

CJ-Wright: Ah ok, sorry for the confusion

scopatz: It is OK, I understand you haven't been coding for a while :smirk:

PR #382, opened 2018-10-19
request review from feedstock maintainers if checks have updated and none are pending

CJ-Wright:
at least for the bot's PRs. As a maintainer it can be difficult to rememe to come back to a PR when the CI is done. This helps by notifying maintainers. This might also accelerate the compiler migration as PRs may be closed faster than waiting for maintainers to remember they exist.

Closes #325

Attn: @regro/auto-tick

### DIFF
--- a/conda_forge_tick/git_utils.xsh
+++ b/conda_forge_tick/git_utils.xsh
@@ -12,6 +12,7 @@ import github3
 import github3.pulls
 
 import networkx as nx
+import requests
 from doctr.travis import run_command_hiding_token as doctr_run
 from pkg_resources import parse_version
 from rever.tools import (eval_version, hash_url, replace_in_file)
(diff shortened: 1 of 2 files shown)


conda_forge_tick/git_utils.xsh
```diff
@@ -160,6 +161,24 @@ def refresh_pr(pr_json: LazyJson, gh=None):
...
+        r = requests.get(pr_json['statuses_url'],
+                         auth=($USERNAME, $PASSWORD))
+        statuses_json = r.json()
+        current_status = {i['context']: (i['state'], i['id']) for i in
+             sorted(statuses_json, key=lambda x: x['updated_at'])}
+        cached_status = LazyJson(pr_json['id'] + 'status.json')
+        # If the status has been updated, and none are pending
+        if current_status != cached_status and all(i['state'] != 'pending' for i in current_status):
+            pr_obj = github3.pulls.PullRequest(pr_json, gh)
+            if pr_obj.create_review_requests(team_reviewers=['conda-forge/' + pr_json['head']['repo']['name']]):
```
scopatz: I don't think this does what you want because the team itself as the reviewers, ie `@conda-forge/pkg-maintainers`.  Rather I think you want to add the list of maintainers here, not the team.

scopatz:
Thanks for putting this in @CJ-Wright!

CJ-Wright:
Actually this is a moot point, we need to go with a comment because only users with write access can request review. I also need to change the lazy json path so that it gets it's own folder.

CJ-Wright:
@regro/auto-tick ready for another round!

PR #390, opened 2018-11-05
add remote

CJ-Wright:
It seems we can't reset onto a url so we add the upstream remote

### DIFF
--- a/conda_forge_tick/git_utils.xsh
+++ b/conda_forge_tick/git_utils.xsh
@@ -135,7 +135,7 @@ def get_repo(attrs, branch, feedstock=None, protocol='ssh',
         # git pull @(upstream) master -s recursive -X theirs --no-edit
         # always run upstream master
         git fetch @(upstream)
-        git remote add ups @(upstream)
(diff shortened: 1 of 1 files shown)


conda_forge_tick/git_utils.xsh
```diff
@@ -135,7 +135,8 @@ def get_repo(attrs, branch, feedstock=None, protocol='ssh',
         # git pull @(upstream) master -s recursive -X theirs --no-edit
         # always run upstream master
         git fetch @(upstream)
-        git reset --hard @(upstream)/master
+        git remote add ups @(upstream)
```
scopatz: Please call this "upstream" for clarity.  I am all for abbreviations, on personal machines, but no need for shortcuts here

CJ-Wright: I didn't want to confuse it with `upstream` which is a python variable (it might be a bit jarring for someone new to xonsh)

scopatz: boh

PR #397, opened 2018-11-14
don't exclude bad upstreams from rebuilds

CJ-Wright:
This solves issues with `r-rcpp`.


CJ-Wright:
@regro/auto-tick

CJ-Wright:
Ready for review!

PR #400, opened 2018-11-16
Use regex to update path for any R license.

jdblischak:
I created a regex that will update the file path for any R license file that is shipped with r-base. It can also handle any number of spaces between the line and the selector in case this was done manually and used something other than 2 spaces.

```
# Fix path to licenses
import re
regex_unix = re.compile('license_file: \'{{ environ\["PREFIX"\] }}/lib/R/share/licenses/(.+)\'\\s+# \[unix\]')
regex_win = re.compile('license_file: \'{{ environ\["PREFIX"\] }}\\\R\\\share\\\licenses\\\(.+)\'\\s+# \[win\]')

lines = ['license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3\'  # [unix]',
         'license_file: \'{{ environ["PREFIX"] }}\R\share\licenses\GPL-3\'  # [win]',
         'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-2\'  # [unix]',
         'license_file: \'{{ environ["PREFIX"] }}\R\share\licenses\GPL-2\'  # [win]',
         'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/MIT\'  # [unix]',
         'license_file: \'{{ environ["PREFIX"] }}\R\share\licenses\MIT\'  # [win]',
         'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/BSD_3_clause\'  # [unix]',
         'license_file: \'{{ environ["PREFIX"] }}\R\share\licenses\BSD_3_clause\'  # [win]']

for i, line in enumerate(lines):
    if regex_unix.match(line.strip()):
        lines[i] = regex_unix.sub('license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/\\1\'',
				  line)
    if regex_win.match(line.strip()):
        lines[i] = ''
lines
# ['license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3\'',
# '',
# 'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-2\'',
# '',
# 'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/MIT\'',
# '',
# 'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/BSD_3_clause\'',
# '']
```

I added a new test and it passes on my local machine. However, I get a [strange error on Travis](https://travis-ci.org/jdblischak/cf-scripts/builds/456155898#L896):

```
        if pmy:
            pmy["version"] = pmy["package"]["version"]
            pmy["req"] = set()
            for k in ["build", "host", "run"]:
>               pmy["req"] |= set(pmy.get("requirements", {}).get(k, set()))
E               TypeError: 'NoneType' object is not iterable
tests/test_migrators.py:1663: TypeError
```

cc: @isuruf @CJ-Wright 

xref: https://github.com/regro/cf-scripts/pull/396/commits/2c41055cc80e78e7632700b5daefdb3e950c06a2#r234064627

### DIFF
--- a/conda_forge_tick/migrators.xsh
+++ b/conda_forge_tick/migrators.xsh
@@ -680,18 +680,18 @@ class Rebuild(Migrator):
                     if s > 0:
                         spacing = s
                     lines[index] = lines[index] + " "*spacing + "noarch: generic\n"
+                regex_unix = re.compile('license_file: \'{{ environ\["PREFIX"\] }}/lib/R/share/licenses/(.+)\'\s+# \[unix\]')
+                regex_win = re.compile('license_file: \'{{ environ\["PREFIX"\] }}\\\R\\\share\\\licenses\\\(.+)\'\s+# \[win\]')
                 for i, line in enumerate(lines_stripped):
                     if noarch and line.lower().strip().startswith("skip: true"):
                         lines[i] = ""
-                    # Fix path to GPL licenses
-                    if line.strip() == 'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-2\'  # [unix]':
-                        lines[i] = '  license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-2\'\n'
-                    if line.strip() == 'license_file: \'{{ environ["PREFIX"] }}\\R\\share\\licenses\\GPL-2\'  # [win]':
+                    # Fix path to licenses
+                    if regex_unix.match(line.strip()):
+                        lines[i] = regex_unix.sub('license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/\\1\'',
+                                                  lines[i])
+                    if regex_win.match(line.strip()):
                         lines[i] = ''
-                    if line.strip() == 'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3\'  # [unix]':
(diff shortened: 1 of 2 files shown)


CJ-Wright:
It seems that because the build is running on linux the zip build requirement doesn't exist. This causes the build section to report `None` as its value (rather than a list of requirements). This most likely means that we need to have some logic to handle when the section reports a `None`.

jdblischak:
@CJ-Wright I updated the tests to focus on the feature of fixing the license file paths. This avoids the edge case about empty requirements. The build passes now.

CJ-Wright:
Great! This seems reasonable to me (albeit with some cheeky tests with multiple licence entries). I'll merge in the morning tomorrow unless anyone has any objections

jdblischak:
@CJ-Wright Thanks! And some more context for the use of the strange tests. My original plan was to find a CRAN package on conda-forge for each of the licenses to test, but my original test with r-xfun for the MIT license failed due to an unrelated issue, preventing the fix from being merged. Then I realized that the other R recipes being tested were contrived too (i.e. adding `{{ compiler('c') }}` to the recipe for r-stabledist, which doesn't actually need to be compiled). Thus including multiple licenses in the recipe for r-stabledist (which I knew wouldn't have any unrelated problems) seemed to be the most straightforward test of this new functionality.

PR #403, opened 2018-11-20
remove archived from status

### LINKED ISSUE #380: Drop archived repos from status page
Looking at the status page, things like `bob.*` show up, which are all archived IIRC. Would be good to drop these archived repos from dependencies to avoid making things seem more important than they are.

screen shot 2018-10-16 at 12 39 45

CJ-Wright:
closes #380


scopatz:
LGTM! Feel free to merge when done

PR #413, opened 2018-11-30
Update R license even if the path contains extra backslashes.

jdblischak:
This is a follow to PRs #396 and #400. In some R recipes, the path to the license file has extraneous backslashes in it. Thus they did not match the regex and needed to be manually fixed ([example](https://github.com/conda-forge/r-mefa-feedstock/pull/2#discussion_r237879766)).

This PR adds support to catch these edge cases using the following regex:

```
import re
regex_unix2 = re.compile('license_file: \'{{ environ\["PREFIX"\] }}\\\/lib\\\/R\\\/share\\\/licenses\\\/(.+)\'\\s+# \[unix\]')

lines = ['license_file: \'{{ environ["PREFIX"] }}\/lib\/R\/share\/licenses\/GPL-2\'  # [unix]',
         'license_file: \'{{ environ["PREFIX"] }}\/lib\/R\/share\/licenses\/BSD_3_clause\'  # [unix]']

for i, line in enumerate(lines):
    if regex_unix2.match(line.strip()):
        lines[i] = regex_unix2.sub('license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/\\1\'',
				  line)				  
lines
# ['license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-2\'',
#  'license_file: \'{{ environ["PREFIX"] }}/lib/R/share/licenses/BSD_3_clause\'']

```

cc: @bgruening

PR #415, opened 2018-11-30
Add instructions for setting up local test environment.

jdblischak:
xref: #413 

cc: @CJ-Wright


CJ-Wright:
Thank you very much!

PR #440, opened 2019-01-18
restart bot but with gcc7 as default now

CJ-Wright:


conda_forge_tick/migrators.xsh
```diff
@@ -738,18 +738,9 @@ class Rebuild(Migrator):
 
 
 class CompilerRebuild(Rebuild):
-    bump_number = 1000
     migrator_version = 1
```
scopatz: Should you bump the migrator version?

CJ-Wright: I'm not certain, I think if we do that it will invalidate the current cache of PRs it has issued causing it to PR against everything again.

CJ-Wright: This is because we don't check if the current compiler is gcc7 (I'm not certain if there is a way for us to access that)

scopatz: Ahhh ok!

PR #454, opened 2019-02-13
Add openssl migrator

soapy1:


conda_forge_tick/auto_tick.xsh
```diff
@@ -22,7 +22,7 @@ logger = logging.getLogger("conda_forge_tick.auto_tick")
 # https://travis-ci.org/regro/00-find-feedstocks/jobs/388387895#L1870
 from .migrators import *
 $MIGRATORS = [
-   Version(pr_limit=7),
+   OpenSSLRebuild(pr_limit=7),
```
scopatz: I think we want both Version and OpenSSLRebuild here

conda_forge_tick/auto_tick.xsh
```diff
@@ -270,7 +310,7 @@ def initialize_migrators(do_rebuild=False):
     # TODO: reenable once graph order is correct
     if do_rebuild:
         add_rebuild($MIGRATORS, gx)
-    add_arch_migrate($MIGRATORS,gx)
+    add_rebuild_openssl($MIGRATORS, gx)
```
scopatz: again we probably want both

conda_forge_tick/cli.xsh
```diff
@@ -3,11 +3,11 @@ import time
...
-from .all_feedstocks import main as main0
-from .make_graph import main as main1
-from .update_upstream_versions import main as main2
-from .auto_tick import main as main3
-from .status_report import main as main4
+from .all_feedstocks import main as main_all_feedstocks
+from .make_graph import main as main_make_graph
+from .update_upstream_versions import main as main_update_upstream_versions
+from .auto_tick import main as main_auto_tick
+from .status_report import main as main_status_report
```
scopatz: Yes, thank you finally!!!!

scopatz:
Thanks a million for putting this in @soapy1!

conda_forge_tick/migrators.xsh
```diff
@@ -822,6 +822,27 @@ class CompilerRebuild(Rebuild):
                             self.graph.out_edges($PROJECT))[:5]])))
         return body
 
+class OpenSSLRebuild(Rebuild):
+    migrator_version = 1
+
+    def migrate(self, recipe_dir, attrs, **kwargs):
```
CJ-Wright: Why is this needed?

scopatz: hmm yeah, it just seems to be calling the super

CJ-Wright: We might not need the class. I think the message might be the stock rebuild message?

scopatz: Well the message probably should be customized

soapy1: Ya, I copy-pasted this from the CompilerRebuild class. I figure that they are logically different rebuilds so they should get different classes. Any ideas on what the rebuild message should be? I was thinking to add something at the beginning of the message like "This PR has been triggered in an effort to update openSSL."

scopatz: yeah, I think that is good enough.  And I agree they are logically different and should get different version numbers

CJ-Wright: I think this should be different from the `CompilerRebuild` but maybe not from the `Rebuild` general class. To some extent this may be a moot point for our current level of development, but I'm not certain that we want a bespoke class per rebuild migration. The compiler and arch got their own because they need to poke other parts of the CF infrastructure (changing the yaml files) than just re-rendering and bumping the build number.

scopatz: Hmmm, I think it is useful to track different package ABI breakages separately.  Right now, that is done by having a minimal subclass with its own version number.  Also, this gives the option of heavily customizing the rebuild if needed.

CJ-Wright:
I'm happy to merge this as is, so take this with a mountainous grain of salt. 

I think we should discuss (maybe in an issue?) what this will look like going forward too. I'm a bit suspicious of having a new class being written for each migration. My hope is that most migrations have the same machinery and that the edge cases which don't will get their own subclass (eg we need to poke some more advanced configuration). How would this work when we move to automate the rebuild system will something produce the classes, do we need a metaclass? We might consider allowing PR messages to be passed into the class optionally so we can have more customization on that front.

CJ-Wright:
@soapy1 thank you very much!

scopatz:
> We might also consider giving runtime version numbers to the rebuild objects so they have a class level version number

Yeah, I had that thought too.

PR #461, opened 2019-02-18
urls can be lists

CJ-Wright:
This fixes a long standing issue when checking if there are newer version of r packages

### DIFF
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -21,7 +21,11 @@ def urls_from_meta(meta_yaml):
     urls = set()
     for s in source:
         if "url" in s:
-            urls.add(s["url"])
+            # if it is a list for instance
+            if not isinstance(s["url"], str):
+                urls.update(set(s['url']))
+            else:
+                urls.add(s["url"])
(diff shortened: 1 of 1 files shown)


CJ-Wright:
@regro/auto-tick

conda_forge_tick/update_upstream_versions.py
```diff
@@ -21,7 +21,11 @@ def urls_from_meta(meta_yaml):
     urls = set()
     for s in source:
         if "url" in s:
-            urls.add(s["url"])
+            # if it is a list for instance
+            if not isinstance(s["url"], str):
+                urls.update(set(s['url']))
```
mariusvniekerk: technically you don't really need to bother with casting to a set here, because set will do this already

CJ-Wright: Fair point

CJ-Wright:
```suggestion
                urls.update(s['url'])
```

CJ-Wright:
I'll merge when the ci is finished

CJ-Wright:
Unleash the R versions!

PR #465, opened 2019-02-22
put migrator and object versions into branches

CJ-Wright:


CJ-Wright:
@regro/auto-tick This fixes an issue with the issuing of new Rebuilds against the same class and name(for instance if we updated the migrator or the object itself) by appending the migrator version and the object version to the branch.

conda_forge_tick/auto_tick.xsh
```diff
@@ -410,7 +410,10 @@ def main(args=None):
                 break
             $PROJECT = attrs['feedstock_name']
             $NODE = node
-            logger.info('%s IS MIGRATING %s', migrator.__class__.__name__.upper(),
+            m_name = migrator.__class__.__name__.upper()
+            if m_name == 'REBUILD':
+                m_name = migrator.name.upper()
```
scopatz: Maybe all migrators classes should have a name attr/property, which just defaults to `__class__.__name_`_, so that you don't have to put this logic everywhere

scopatz: And can just use migrator.name.

CJ-Wright: This would break our current name for the Arch rebuild

scopatz: But is a better design overall.

CJ-Wright: Can we put that into its own PR? It seems like that might be a breaking change, or at the very least require a description kwarg added to the migrator.

scopatz: It seems like there should be a name property, which takes care of this

CJ-Wright: The ``name`` is used alongside the ``migrator_name`` (aka class name) in the PRed_json keys.

conda_forge_tick/migrators.xsh
```diff
@@ -798,7 +798,8 @@ class Rebuild(Migrator):
         return $USERNAME + ':' + self.remote_branch()
 
     def remote_branch(self):
-        return 'rebuild'+self.name.lower().replace(' ', '_')
+        s_obj = str(self.obj_version) if self.obj_version else ''
+        return 'rebuild'+self.name.lower().replace(' ', '_')+str(self.migrator_version)+s_obj
```
mariusvniekerk:
```suggestion
        return 'rebuild' + self.name.lower().replace(' ', '_') + str(self.migrator_version) + s_obj
```

conda_forge_tick/auto_tick.xsh
scopatz: Maybe there should be a different attribute / property for this then

CJ-Wright: Maybe, or a migration of the graph itself.

CJ-Wright:
I think this is ready for another round of review @regro/auto-tick

CJ-Wright:
This will help make sure our branch names are unique across class level and object level versions

PR #466, opened 2019-02-23
delete branches of PRs closed for a day or more

CJ-Wright:

PR #467, opened 2019-02-25
now branches are removed if they are closed and old

CJ-Wright:
This fixes a bug from #466 which doesn't trigger because the PRs are already
closed, causing it to not trip the if statement.


mariusvniekerk:
🤣

PR #478, opened 2019-03-13
don't use archived packages as deps for migration

CJ-Wright:


CJ-Wright:
@regro/auto-tick this should unblock conda-build

scopatz:
seems reasonable

CJ-Wright:
thank you!

PR #479, opened 2019-03-14
Add libprotobuf migrator

xhochy:
Inspired by the OpenSSL migrator. I have roughly understood how this work but I did not understand how it determines which packages are already rebuild. Does it compare the pinnings of the package with the pinnings in `conda-forge-pinnings`? I.e. will this also work for future `libprotobuf` repins?

### DIFF
--- a/conda_forge_tick/auto_tick.xsh
+++ b/conda_forge_tick/auto_tick.xsh
@@ -298,7 +298,7 @@ def add_rebuild_libprotobuf(migrators, gx):
     migrators.append(
         Rebuild(graph=total_graph,
                 pr_limit=5,
-                name='libprotobuf',
+                name='libprotobuf-3.7',
                 top_level=top_level,
                 cycles=cycles, obj_version=3))
 
--- a/conda_forge_tick/auto_tick.xsh
+++ b/conda_forge_tick/auto_tick.xsh
@@ -261,6 +261,48 @@ def add_rebuild_openssl(migrators, gx):
                 cycles=cycles, obj_version=3))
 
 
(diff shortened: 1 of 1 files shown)


conda_forge_tick/auto_tick.xsh
```diff
@@ -261,6 +261,48 @@ def add_rebuild_openssl(migrators, gx):
...
+    top_level = {node for node in gx.successors("libprotobuf") if
+                 (node in total_graph) and
+                 len(list(total_graph.predecessors(node))) == 0}
+    cycles = list(nx.simple_cycles(total_graph))
+    # print('cycles are here:', cycles)
+
+    migrators.append(
+        Rebuild(graph=total_graph,
+                pr_limit=5,
+                name='libprotobuf',
```
CJ-Wright:
```suggestion
                name='libprotobuf-3.7',
```

CJ-Wright:
> Does it compare the pinnings of the package with the pinnings in conda-forge-pinnings? 

No it just rebuild the things in order, keeping track of what has been merged/closed, what is in PR and what has yet to be done.

> I.e. will this also work for future libprotobuf repins?

I think if we add the version number to this then we can reuse it for future ones by changing the name

xhochy:
This doesn't seem to work as expected: https://github.com/conda-forge/libhdfs3-feedstock/pull/25#pullrequestreview-214941122

xhochy:
Ah, it was merged too early, now it works. We should probably leave enough time between merging the migrator and the pinning changes so that the bot already has the new pinning.

PR #481, opened 2019-03-18
Fixed issue where a single bad version could break the whole bot

mariusvniekerk:
This fixes issues like those in https://circleci.com/gh/regro/circle_worker/6281

This does NOT fix the underlying issue, but prevents it from wreaking havoc.


CJ-Wright:
Travis is complaining about a suntan syntax error

PR #486, opened 2019-03-20
remove compiler rebuild

CJ-Wright:


CJ-Wright:
This is to make certain that the auto status page doesn't list the compiler migration
attn: @regro/auto-tick

PR #496, opened 2019-04-11
move PR data to outermost of dict

CJ-Wright:


CJ-Wright:
@ocefpaf This should fix the issues we're seeing with the migration and the `bot-rerun` things.

PR #502, opened 2019-04-19
end proj4 6.0.0 migration

ocefpaf:
Proj4 6.0.0 was a success IMO, only two left and they are already in PR\*. 

@scopatz and @mariusvniekerk I'm not sure if this is the right way to finish a migration, or even if we have to?

\* QGIS is quite complicated and may need patching to work with proj4 6 and postgis seems to have a conflict when installing gdal/proj4/postgresql that I'm investigating but it may take a while due to the high number of dependencies involved.

proj4-migration


scopatz:
If you are happy with this, then I am :)

PR #503, opened 2019-04-25
migrate to qt 5.9.7

ocefpaf:
Follow up on https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/221

Note to self: once this is over we may want to drop the pin on `pyqt` and let `conda-build 3` and `run_exports` that care of that one, and start a new migration to the `icu` with `run_exports`.


scopatz:
This LGTM.  Feel free to merge when CI is done

PR #507, opened 2019-05-14
end zeromq and lz4 migration

ocefpaf:
Follow up on the bot sub-group meeting today. We will keep `openssl` and `qt` open for a while. (Qt may be closed as soon as I fix the `qwtpolar` PR.)


scopatz:
Thanks!

PR #520, opened 2019-06-12
Add pyarrow to aarch64 rebuild

xhochy:
I'll also have a look at fixing the dependencies that pop up along the way.


scopatz:
Thanks!

PR #527, opened 2019-06-25
end blas2, libprotobuf, boos-cpp and re-enable-boost

ocefpaf:
As discussed in today's meeting.

Note that the `boost-cpp` will be covered by the `boost` migration we are re-starting here. All others reached a dead end like `shogun`, `sage`, or `audi`, which require more work to get the feedstocks building again.

### DIFF
--- a/conda_forge_tick/auto_tick.xsh
+++ b/conda_forge_tick/auto_tick.xsh
@@ -446,11 +446,8 @@ def initialize_migrators(do_rebuild=False):
 
     add_arch_migrate($MIGRATORS,gx)
     add_rebuild_openssl($MIGRATORS, gx)
-    add_rebuild_libprotobuf($MIGRATORS, gx)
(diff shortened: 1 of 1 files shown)


conda_forge_tick/auto_tick.xsh
```diff
@@ -446,11 +446,8 @@ def initialize_migrators(do_rebuild=False):
 
     add_arch_migrate($MIGRATORS,gx)
     add_rebuild_openssl($MIGRATORS, gx)
-    add_rebuild_libprotobuf($MIGRATORS, gx)
-    add_rebuild_blas($MIGRATORS, gx)
```
ocefpaf: There are code bits associated with these migrations. Should we remove them too or should we leave them there for future re-use?

scopatz: We should keep them around for now

PR #532, opened 2019-06-28
Handle manually PRed migrators

### LINKED ISSUE #531: Migrator status not correct due to split reliance on PRed and PRed_json
We currently list a lot packages in the `awaiting-parents` section although they have been rebuilt. This is caused by https://github.com/regro/cf-scripts/blob/e7ce9c61a9101025a2910e7b130fffdbeead4542/conda_forge_tick/auto_tick.xsh#L501-L505 using `PRed_json` to check for the status whereas https://github.com/regro/cf-scripts/blob/e7ce9c61a9101025a2910e7b130fffdbeead4542/conda_forge_tick/migrators.xsh#L62 is using `PRed` as the base. As these values can differ, this causes the listing of packages in the wrongly sections.

An example for this is the `pkgconfig` package. Here we have manually done the Arch migration in https://github.com/conda-forge/pkgconfig-feedstock/pull/10 but a PR from the bot was never issued. The code for listing the status of the migration considers them as TODO while the migrator itself checks in the sources whether the migration was already done and will not issue a PR.

xhochy:
Fixes #531

### DIFF
--- a/conda_forge_tick/auto_tick.xsh
+++ b/conda_forge_tick/auto_tick.xsh
@@ -503,9 +503,16 @@ def migrator_status(migrator: Migrator, gx):
                 break
         else:
             pr_json = None
+
+        # No PR was ever issued but the migration was performed.
+        # This is only the case when the migration was done manually before the bot could issue any PR.
+        manually_done = pr_json is None and frozen_to_json_friendly(migrator.migrator_uid(attrs)) in attrs.get('PRed', [])
+
         buildable = not migrator.filter(attrs)
 
-        if pr_json is None:
+        if manually_done:
+            out['done'].add(node)
+        elif pr_json is None:
             if buildable:
                 out['awaiting-pr'].add(node)
             else:


scopatz:
Thanks for putting this in @xhochy!  I am a little confused how the original PR that is put in gets noticed by the migrator and ends up in "PRed".

xhochy:
The original/manual PR never gets noticed. It is rather that the migrator is run on the package and no migration PR is issued as the Migrator object detects that no code change is necessary. Still, the result of the `migrate` call in `cf-scripts` is recorded as content in `PRed`. 

One of the things that really confused me while reading the code was that I assumed that `PRed` and `PRed_json` would point to the same information and were just a different representation of each other. This is not the case as `PRed_json` are all PRs issued to a feedstock and `PRed` are all migrators run on a feedstock (it seems that `len(PRed_json) <= len(PRed)` holds, in the cases this PR tackles, it is always a strong `<`).

CJ-Wright:
I would be interested in having `PRed_json` subsume `PRed` in this case, we could create some dummy metadata which is overridden if a PR is actually issued.
The `PRed` key comes from a time when we didn't track the actual PR itself (when we were only running version bumps) but needed to know if we issued a PR. Now we need to track the PR status, so `PRed_json` should be more descriptive.

xhochy:
@CJ-Wright Would it be ok for you to merge this PR as-is and do the `PRed` cleanup separately? This is my first interaction with this codebase besides adding simple migrators and thus would take me quite some time to understand everything to make the change.

CJ-Wright:
I don't have merge rights at the moment, but this seems reasonable to me.

mariusvniekerk:
Lets go!

PR #534, opened 2019-07-02
end boost pyqt zstd hdf5 libprotobuf migrations

ocefpaf:
`zstd` actually complete :tada: 
All others either issued all the possible PRs or have shogun/shogun-cpp awaiting parents.


scopatz:
Thanks @ocefpaf!

PR #549, opened 2019-07-19
can we have 25 over the weekend?

bgruening:
Hi,

could we have 25 builds over the weekend? I guess could finish the migration then quickly.

Thanks,
Bjoern


scopatz:
Let's try it and see if it falls over

bgruening:
Merci, it already helped a lot!!!

PR #561, opened 2019-08-09
end r-base migration

bgruening:


ocefpaf:
Just to keep a record here the two open PRs are:

https://github.com/conda-forge/r-openmx-feedstock/pull/10
https://github.com/conda-forge/r-fst-feedstock/pull/5

and the two awaiting parents are `r-mbess` and `r-userfriendlyscience`.

PR #580, opened 2019-08-31
lazy json all the way to the nodes

CJ-Wright:
This needs to be checked over, since I'd like to stop the
bot when we go to run this migration of the graph itself.

Overall this should reduce git thrashing a lot, 
since we will only touch the files we need to
and it will make the graph.json smaller.

Additionally this will make the state of individual
nodes much easier to read since they can be read as
json themselves.


CJ-Wright:
So this didn't work as I wanted. The node attributes needs to be a dict, I can't just have it be a lazyjson object.

PR #585, opened 2019-09-01
Use lazyjson payload for feedstock info

CJ-Wright:
This will put the feedstock data into their own files
making it easier to read/write to them via third party tools.
Note that this needs a migration of our existing graph into this structure.

### DIFF
--- a/conda_forge_tick/auto_tick.xsh
+++ b/conda_forge_tick/auto_tick.xsh
@@ -414,8 +414,7 @@ def add_arch_migrate(migrators, gx):
     total_graph = copy.deepcopy(gx)
 
     for node, node_attrs in gx.node.items():
-        attrs = node_attrs
-        # attrs = node_attrs['payload']
+        attrs = node_attrs['payload']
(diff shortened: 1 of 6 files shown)


conda_forge_tick/auto_tick.xsh
```diff
@@ -368,6 +371,7 @@ def add_rebuild_blas(migrators, gx):
     total_graph = copy.deepcopy(gx)
 
     for node, attrs in gx.node.items():
+        attrs = attrs['payload']
```
scopatz: Maybe this should be `attrs = node_attrs["payload"]` and similar cases below.

scopatz:
Looks reasonable, but I am unsure of migration path

CJ-Wright:
Same as the other graph migration.

CJ-Wright:
The migration will use the https://github.com/regro/cf-scripts/pull/585/files#diff-09a07bc2e2c9570b3861d6755dff6a95 script to move everything over to the payload system.

CJ-Wright:
Failing because upstream graph is not migrated yet.

PR #588, opened 2019-09-02
add bot_rerun to the PR data so that if a rerun is needed we don't delete the PR info, we just make it not match the muid

CJ-Wright:


scopatz:
LGTM but is failing CI

PR #590, opened 2019-09-03
MNT end gfortran 7 for OSX migration

beckermr:
This PR ends the gfortran 7 for OSX migration. We have issued all of the PRs we possibly can. The remaining ~10-13 feedstocks that need migrating have complicated build issues or actual package test failures that need the attention of those maintainers.


beckermr:
@isuruf for viz

PR #598, opened 2019-09-10
retain old version if bad

CJ-Wright:


conda_forge_tick/update_upstream_versions.py
```diff
@@ -296,15 +296,15 @@ def _update_upstream_versions_process_pool(gx, sources):
...
-                    attrs["new_version"] = f.result()
+                    new_version = f.result()
+                    attrs["new_version"] = new_version or attrs['new_version']
                 except Exception as e:
                     try:
                         se = str(e)
                     except Exception as ee:
                         se = "Bad exception string: {}".format(ee)
                     logger.warn("Error getting uptream version of {}: {}".format(node, se))
                     attrs["bad"] = "Upstream: Error getting upstream version"
```
mariusvniekerk: We should make this consistent with the other uses of bad which use dicts

PR #606, opened 2019-09-11
write out status for unmaintained feedstocks

CJ-Wright:


CJ-Wright:
@regro/auto-tick

conda_forge_tick/status_report.py
```diff
@@ -6,25 +6,37 @@
...
+            if len(
+                [
+                    z
+                    for z in v.get("payload", {}).get("PRed", [])
+                    if z.get("PR", {}).get("state", "closed") == "open"
+                    and z.get("data", {}).get("migrator_name", "") == "Version"
+                ]
+            )
+            >= 3
+        ]
```
scopatz: l's definition should not be in the with block

PR #618, opened 2019-09-24
fix bad comparisons, which looked at the PR objects which are unique

CJ-Wright:
Don't merge unless you can support/fix. Although review would be great!


scopatz:
LGTM!

Issue #624, opened 2019-09-27
allow migration yamls to blacklist packages from graph construction


CJ-Wright:
Currently some migrations may require the addition of keys which should not be included in the graph of packages to be migrated. For instance, adding a python pin requires updating the associated compilers, since they are related via zip keys. This may cause the bot to think it needs to migrate all packages which inherit from those packages (if they exist) and will cause a mass migration for many non-python things. I think the solution here is to add metadata to the migration yaml itself, or the bot itself, which blacklists these from the packages that are used as the parent nodes in the graph.

Thoughts @regro/auto-tick

PR #626, opened 2019-09-30
support exclusions and don't migrate noarch packages

CJ-Wright:
Closes https://github.com/regro/cf-scripts/issues/624
Noarch packages don't have ABI incompatibilities since they don't link so they
never need to be in a rebuild. This removes noarch pythons from our python rebuild.

Also support excluding things from the rebuild, in the case of python
the c compilers


CJ-Wright:
@regro/auto-tick I think this is ready for review. Just to double check, we won't ever need to perform a pin migration on `noarch` packages, right?

scopatz:
I think this is generally true. Though there may be some truly weird corner cases we can ignore here and do manually

CJ-Wright:
Thank you!

PR #633, opened 2019-10-04
Run some integration tests

CJ-Wright:
This way we can know if things aren't working.

This may need some integration on the Travis side as well.


CJ-Wright:
As suspected this needs keys for the bot.

CJ-Wright:
I've added the missing keys. I think this should work now. We need to make certain that we don't push on things. Also we don't want to blow up the GH api.

CJ-Wright:
use dry-run here

beckermr:
closing as this PR is too old to be merged

PR #634, opened 2019-10-05
Get cran upstream versions directly from cran v2

dbast:
context: https://github.com/regro/cf-scripts/pull/630

The second commit fixes the previous seen issues with dask, by not keeping the cran index inside the cran class, which get serialized for every dask job and causes memeory issues already on job submission.

This time fully locally tested without any additonal changes :) 

@CJ-Wright Can we give this another try?

Thanks!


CJ-Wright:
Thank you very much! Sorry for the delay.

PR #638, opened 2019-10-08
check zip vs tarball when url fails

CJ-Wright:


CJ-Wright:
@regro/auto-tick I think this is ready for review and should cut down on some of our version blind spots.

CJ-Wright:
This could use a test, but I think this should wait for the license bot to go through.

PR #645, opened 2019-10-10
more migration docs

CJ-Wright:
Thoughts @jakirkham?


jakirkham:
Can we call out `migration_ts` explicitly?

CJ-Wright:
@jakirkham is this enough to go off of for an arrow migration PR?

PR #659, opened 2019-10-16
handle multi output recipes better by linking them back to the parent feedstock

CJ-Wright:
@ocefpaf


ocefpaf:
That means I no longer have an excuse to stall the static library splits :stuck_out_tongue_winking_eye:

CJ-Wright:

CJ-Wright:
I'd give others some time. I think most of the damage on the existing migrations has been done.

PR #671, opened 2019-11-01
make unmaintained name friendlier

CJ-Wright:
This way when people ask where they can help we can point to this list


CJ-Wright:
@regro/auto-tick

PR #682, opened 2019-11-03
Batch requests in order to avoid rate limiting failures

soapy1:


conda_forge_tick/make_graph.py
```diff
@@ -26,7 +27,8 @@
 
 
 NUM_GITHUB_THREADS = 4
-
+BATCH_SIZE = 10
+SLEEP_TIME = 60
```
soapy1: Maybe this is too large a unit of time. https://developer.github.com/v3/rate_limit/

mariusvniekerk: I'd probably go for some light exponential back off here capping at 60s

mariusvniekerk: Odds are we should just put this block into something that we just allocate a fixed maximal time for, eg if we've been updating for 10 minutes just stop.

conda_forge_tick/make_graph.py
```diff
@@ -154,14 +156,18 @@ def make_graph(names, gx=None):
     env = builtins.__xonsh__.env
     debug = env.get("CONDA_FORGE_TICK_DEBUG", False)
     builder = _build_graph_sequential if debug else _build_graph_process_pool
-    builder(gx, total_names, new_names)
+
+    for i in range(0, ceil(len(total_names)/BATCH_SIZE)):
+        names = total_names[i*BATCH_SIZE:(i+1)*BATCH_SIZE]
+        time.sleep(SLEEP_TIME)
+        builder(gx, names, new_names)
```
CJ-Wright: I think this part is usually fine, since we aren't actually using the gh API. The problem usually comes from the updating of the prs here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/make_graph.py#L181

conda_forge_tick/git_utils.xsh
```diff
@@ -238,7 +245,9 @@ def push_repo(feedstock_dir, body, repo, title, head, branch,
     ljpr.update(**pr_dict)
     return ljpr
 
-
+@backoff.on_exception(backoff.expo,
+  (requests.exceptions.RequestException, requests.exceptions.Timeout),
+  max_time=60)
```
CJ-Wright: Can this set the `max_time` from a global or an env var? (I'm imagining that this may need tuning)

conda_forge_tick/git_utils.xsh
```diff
@@ -20,11 +20,14 @@ from pkg_resources import parse_version
 from rever.tools import (eval_version, hash_url, replace_in_file)
 from xonsh.lib.os import indir
 
+import backoff
+from request.exceptions import Timeout, RequestException
```
CJ-Wright:
```suggestion
from requests.exceptions import Timeout, RequestException
```

CJ-Wright:
This looks good! I'll merge when I have a few minutes to watch the bot.

PR #684, opened 2019-11-03
add piggy back migrators to migration yaml migrator

CJ-Wright:
@ocefpaf


mariusvniekerk:
Yeah that seems like a great idea.

ocefpaf:
Thanks!

PR #685, opened 2019-11-04
Add several HEP packages to migration

henryiii:


CJ-Wright:
I'm going to let the python rebuild slow down a bit before merging this if that's ok.

CJ-Wright:
Thank you very much!

PR #689, opened 2019-11-05
Make MigrationYaml migrator take __migration config

soapy1:


conda_forge_tick/migrators.xsh
```diff
@@ -1345,6 +1344,7 @@ class MigrationYaml(Migrator):
             self.pr_limit = 2
         elif number_pred < 7:
             self.pr_limit = 5
+        self.bump_number = yaml_contents.get("build_number", 0)
```
CJ-Wright: In the future this should default to one (but not for this PR)

PR #700, opened 2019-11-13
Add pytest benchmark

henryiii:
Needed for testing a formula (boost-histogram)


CJ-Wright:
Formally this should be handled by the migrator itself

henryiii:
https://github.com/conda-forge/boost-histogram-feedstock/pull/2#issuecomment-553609819

CJ-Wright:
Hmm, that's fair. Although I think we should handle the circular deps at some point.

PR #711, opened 2019-12-06
Extract version starting at the first digit

### LINKED ISSUE #707: spaCy model updates do not get picked up by the bot
One of the biggest differences for these, comparing with other libraries, is that [`spaCy models`](https://github.com/explosion/spacy-models) does not contain an `__init__.py` or `setup.py` that the bot can query for updates. Instead, the version updates are pushed using the json files in the `meta` directory, and are then available as releases.

What are alternatives that can be done to fix this behavior and set the bot to pick up the updates automatically?

Related: https://github.com/conda-forge/spacy-model-en_core_web-feedstock/issues/3

setu4993:
Issue:
- Version number from the `spacy-models` repository contains characters that are not a part of the version. Because of this, the updates do not get picked by the bot.

Changes:
- Perform a regex match to extract version number from the `ver` string such that everything starting at the first digit gets picked up in the pattern.

All of the expected version patterns expected by [`VersionOrder`](https://github.com/conda/conda/blob/da8b5a812bb17d83012f3a0bb63ef7158f1e846e/conda/models/version.py#L47) start with a numeric digit as well.

This fixes #707.

### DIFF
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -72,6 +72,8 @@ class VersionFromFeed:
                     ver = ver[len(prefix) :]
             if any(s in ver for s in self.dev_vers):
                 continue
+            # Extract vesion number starting at the first digit.
+            ver = re.search(r"(\d+[^\s]*)", ver).group(0)
             vers.append(ver)
         if vers:
             return max(vers, key=lambda x: VersionOrder(x.replace("-", ".")))


conda_forge_tick/update_upstream_versions.py
```diff
@@ -72,6 +72,8 @@ def get_version(self, url):
                     ver = ver[len(prefix) :]
             if any(s in ver for s in self.dev_vers):
                 continue
+            # Extract vesion number starting at the first digit.
+            ver = re.search(r"(\d+.*)", ver).group(0)
```
msarahan:
```suggestion
            ver = re.search(r"(\d+[^\s]*)", ver).group(0)
```

might be a tiny bit safer?

setu4993:
Good suggestion, @msarahan. Accepted the change, rebased to `master` and force-pushed.

PR #721, opened 2019-12-17
pull arch rebuild packages from pinnings text file

CJ-Wright:


CJ-Wright:
I think this is failing because the file doesn't exit yet (pending https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/356)

CJ-Wright:
@regro/auto-tick

ocefpaf:
> I think this is failing because the file doesn't exit yet (pending [conda-forge/conda-forge-pinning-feedstock#356](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/356))

Any reason to not do both in a single PR?

CJ-Wright:
> Any reason to not do both in a single PR?

They are in two different repos, can I do multi-repo/multi-org PRs?

ocefpaf:
> They are in two different repos, can I do multi-repo/multi-org PRs?

Sorry. I clicked to early on the comment button. What I meant is on the edit above.

CJ-Wright:
I think we've been pretty good about exposing migrations to people via the pinning repo. I'd like to continue that trend so we can get better engagement. Also I like having the "what I want" and "how does that get done" split between this and pinnings.

CJ-Wright:
This is failing because the file layout of travis is a bit different than the file layout in production. The tests are run from their own dir rather than inside the cf-graph-countyfair dir.

scopatz:
Does that mean that it is OK to merge?

CJ-Wright:
maybe? I'll merge tomorrow and see how it goes.

scopatz:
This has conflicts now

PR #724, opened 2019-12-23
fix cran check

CJ-Wright:


CJ-Wright:
xref: @723
The issue here was that R packages had `cran_mirror` in the raw URL meaning the bot didn't know to fix the version from conda-compatible to R-compatible. This caused a bunch of feedstocks to not get PRs which in turn caused other feedstocks to become blocked as we ran out of time.

This PR should fix that issue with R recipes getting PRs again which should unblock other PRs

Thanks to @andrewAnnex and @pgunn for bringing this issue up!

PR #725, opened 2019-12-23
support swapping v and not v version numbers

CJ-Wright:


CJ-Wright:

CJ-Wright:
I think this will get rolled into Marius' mega PR and then we can think about maintainability. @ocefpaf thoughts?

ocefpaf:
> I think this will get rolled into Marius' mega PR and then we can think about maintainability. @ocefpaf thoughts?

LGTM ideally we should investigate if @jjhelmus way to do this is more robust and port that here. But that should not stop us from merging this.

PR #726, opened 2019-12-23
increase number of version bump PRsto deal with backlog

CJ-Wright:
Currently we do 30 and it takes 20 mins so 60 should put us at 40mins, which should leave some breathing room for the other parts to be within the 1 hour time limit.


CJ-Wright:
@regro/auto-tick

PR #740, opened 2020-01-04
Add occt to the arch migrator

hmaarrfk:
I got a request for the python version of this by email. so this is at least a start.


CJ-Wright:
Thank you!

PR #742, opened 2020-01-06
meta_yaml is down one level

CJ-Wright:


CJ-Wright:
@regro/auto-tick ready for review

PR #751, opened 2020-01-13
strong exports support (and fix gdal/outputs)

CJ-Wright:
This needs work, but is ready for a look over

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -25,7 +25,6 @@ from .path_lengths import cyclic_topological_sort
 from .utils import (
     setup_logger,
     pluck,
-    get_requirements,
     load_graph,
     dump_graph,
     LazyJson,
@@ -204,16 +203,6 @@ def run(
     return migrate_return, ljpr
 
 
-def _requirement_names(reqlist: Optional[Sequence[Optional[str]]]) -> List[str]:
-    """Parse requirement names from a list ignoring `None`
-    """
-    if reqlist is None:
-        return []
-    else:
-        return [r.split()[0] for r in reqlist if r is not None]
-
-
-# TODO: unify this with get_requirements
 def _host_run_test_dependencies(meta_yaml: "MetaYamlTypedDict") -> Set["PackageName"]:
     """Parse the host/run/test dependencies of a recipe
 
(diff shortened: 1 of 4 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -222,27 +211,8 @@ def _host_run_test_dependencies(meta_yaml: "MetaYamlTypedDict") -> Set["PackageN
...
-            rq.update(_requirement_names(req.get("build", []) or []))
-        rq.update(_requirement_names(req.get("run", []) or []))
-
-        # add testing dependencies
-        test: "TestTypedDict" = block.get("test", {})
-        rq.update(_requirement_names(test.get("requirements")))
-        rq.update(_requirement_names(test.get("requires")))
-
+    _ = meta_yaml["requirements"]
+    rq = (_["host"] or _["build"]) | _["run"] | _["test"]
```
beckermr: Whoa. This underscore thing threw me for a loop but I see it now. I wonder if a named var is better?

conda_forge_tick/auto_tick.py
```diff
@@ -272,15 +242,11 @@ def add_replacement_migrator(
...
+        rq = set().union([requirements[k] for k in ["build", "host", "run", "test"]])
         pkgs = {old_pkg}
-        old_pkg_c = len(pkgs.intersection(get_requirements(meta_yaml))) > 0
+        old_pkg_c = pkgs.intersection(rq)
 
-        rq = _host_run_test_dependencies(meta_yaml)
-
-        for e in list(total_graph.in_edges(node)):
-            if e[0] not in rq:
-                total_graph.remove_edge(*e)
```
beckermr: I see this block is removed here but not in the rebuild migrator below. Why is this?

conda_forge_tick/auto_tick.py
CJ-Wright: Most likely it will be I was coding this without an IDE, so some code could use some work.

conda_forge_tick/auto_tick.py
CJ-Wright: This motif is removing edges that are part of the wrong requirements section. In the rebuild we don't care about build, only host, run and test (assuming all feedstocks have the build/host split). So we don't want any edges from build, since that would cause weird rebuilds of build tooling that we don't need. This isn't needed in the replace package chunk because nominally all parts of the requirements are fair game, so no edges need removal. This is also why the arm rebuild doesn't have that section, it doesn't care if you are a build tool or not.

CJ-Wright: Nominally we should replace these by making a multi graph with each edge labeled by the section it is in, but I didn't want to do that in this PR

CJ-Wright:
This is failing because the graph is structured differently.

scopatz:
LGTM

CJ-Wright:
I think this needs a fix in the travis.yaml to undo the thing I did

CJ-Wright:
actually this needs work still

CJ-Wright:
@scopatz

PR #754, opened 2020-01-16
ENH add bot-automerge slug

### LINKED ISSUE #752: Add slug for PR merging bot
We need to add a slug to the PRs so they can be identified and parsed by the merging bot. Maybe something in the initial comment with the information in the "data" key from the PR. 

@beckermr

beckermr:
This PR adds the `[bot-automerge]` slug to PR titles for version bumps.

@CJ-Wright for code review

Closes #752

### DIFF
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -721,7 +721,12 @@ class Version(Migrator):
 
     def pr_title(self, feedstock_ctx: FeedstockContext) -> str:
         assert isinstance(feedstock_ctx.attrs["new_version"], str)
-        return feedstock_ctx.package_name + " v" + feedstock_ctx.attrs["new_version"]
+        return (
+            "[bot-automerge] "
+            + feedstock_ctx.package_name
+            + " v"
+            + feedstock_ctx.attrs["new_version"]
+        )
 
     def remote_branch(self, feedstock_ctx: FeedstockContext) -> str:
         assert isinstance(feedstock_ctx.attrs["new_version"], str)
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -721,8 +721,7 @@ class Version(Migrator):
 
     def pr_title(self, feedstock_ctx: FeedstockContext) -> str:
         assert isinstance(feedstock_ctx.attrs["new_version"], str)
-        # TODO remove this later when we want to automerge everything
-         # TODO: turn False to True when we default to automerge
+        # TODO: turn False to True when we default to automerge
         if feedstock_ctx.attrs.get('conda-forge.yml', {}).get('bot', {}).get('automerge', False):
             add_slug = "[bot-automerge] "
         else:
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -721,8 +721,14 @@ class Version(Migrator):
 
     def pr_title(self, feedstock_ctx: FeedstockContext) -> str:
         assert isinstance(feedstock_ctx.attrs["new_version"], str)
+        # TODO remove this later when we want to automerge everything
+        if feedstock_name.feedstock_name == 'cf-autotick-bot-test-package':
+            add_slug = "[bot-automerge] "
+        else:
+            add_slug = ""
+        
         return (
-            "[bot-automerge] "
+            add_slug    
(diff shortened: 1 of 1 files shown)


CJ-Wright:
Should we put a note in the PR body discussing why the slug is there? People might get confused/concerned (since the slug would exist even if they haven't enabled the feature).

CJ-Wright:
CC @scopatz

beckermr:
Yup will do.

beckermr:
Actually, let’s just scope this to the testing feedstock for now. We can address deployment for everything later.

beckermr:
OK @CJ-Wright take a look again. One question I have is what the feedstock name is set to. Is it 
`cf-autotick-bot-test-package` or `cf-autotick-bot-test-package-feedstock`?

conda_forge_tick/migrators/core.py
```diff
@@ -721,7 +721,18 @@ def commit_message(self, feedstock_ctx: FeedstockContext) -> str:
 
     def pr_title(self, feedstock_ctx: FeedstockContext) -> str:
         assert isinstance(feedstock_ctx.attrs["new_version"], str)
-        return feedstock_ctx.package_name + " v" + feedstock_ctx.attrs["new_version"]
+        # TODO remove this later when we want to automerge everything
+        if feedstock_name.feedstock_name == 'cf-autotick-bot-test-package':
```
CJ-Wright: This is correct, we don't have `'feedstock'` in the `feedstock_name` since everything would have that name.

CJ-Wright: We also have access to the `conda-forge.yaml` if you want to use that to check if automerge is enabled.

beckermr: Interesting. I think it makes sense to leave that check in the action, not here. That way if someone changes the automerge setting on master, old PRs will still have the slug and so can be merged.

beckermr: But we can work this out later. I want to simply turn things on for the testing feedstock for now.

conda_forge_tick/migrators/core.py
```diff
@@ -721,7 +721,18 @@ def commit_message(self, feedstock_ctx: FeedstockContext) -> str:
 
     def pr_title(self, feedstock_ctx: FeedstockContext) -> str:
         assert isinstance(feedstock_ctx.attrs["new_version"], str)
-        return feedstock_ctx.package_name + " v" + feedstock_ctx.attrs["new_version"]
+        # TODO remove this later when we want to automerge everything
+        if feedstock_ctx.feedstock_name == 'cf-autotick-bot-test-package':
```
CJ-Wright:
```suggestion
         # TODO: turn False to True when we default to automerge
        if feedstock_ctx.attrs.get('conda-forge.yml', {}).get('bot', {}).get('automerge', False):
```

CJ-Wright: This way we don't actually need to do anything as automerge is enabled

beckermr: Right. See my above comment about keeping this logic only in the automerge bot.

CJ-Wright: Ok

beckermr: So without this we can still turn it on globally. We simply switch the default here: https://github.com/regro/cf-autotick-bot-action/blob/master/conda_forge_tick_action/automerge.py#L29 and then redeploy the bot.

beckermr: We could use the ternary logic and bump the switch in both spots.

CJ-Wright: ah ok

beckermr: Sorry for being terse earlier. You are totally right. What I want to avoid is us having to go through each feedstock and rerender it. As long as we can change the default on the backend, we will be fine. This change allows that. Plus having the switch in two spots is probably a good idea.

CJ-Wright: Would it be worthwhile going the other way. Label PRs which can't be automerged?
That way the automerge is set to the feedstock wide default unless a particular PR wants not to be automerged.

beckermr: Hmmmm. I prefer adding the slug for automerge as opposed to adding a slug to not automerge for now. I think this change might make sense later as this gets further into deployment?

PR #761, opened 2020-01-21
all feedstocks seems to be failing due to timeout errors so maybe this will fix it

CJ-Wright:


scopatz:
Is the failure here relevant?

CJ-Wright:
Seems to be a bad syntax, but I think we can close for now.

PR #769, opened 2020-01-28
quiet hashing, so we don't run over our character limits

CJ-Wright:


CJ-Wright:
@regro/auto-tick it seems that the graph is broken?

CJ-Wright:
I think I've fixed this

PR #773, opened 2020-01-29
Add URL to status graph nodes

bollwyvl:
Love the new graph!

This adds the PR URL (if available) to each node that has one, which becomes clickable when the SVG is not shown as an `img` tag. While this sets a bunch of empty strings, they just get removed. I suppose a fallback value could be just to the feedstock URL itself (wasn't sure where that would come from in hte data). This does make the generated SVG a fair amount chunkier, but a good amount richer.

I couldn't tell how to test the generated SVG locally, but have used this feature a fair amount in the past.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -556,6 +556,7 @@ def migrator_status(
                 fillcolor=fc,
                 style="filled",
                 fontcolor=fntc,
+                URL=pr_json.get("PR", {}).get("html_url", "")
             )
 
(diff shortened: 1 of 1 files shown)


CJ-Wright:
tested on my local

CJ-Wright:
Thank you very much!

bollwyvl:
And they're up! It's pretty cool. Would having the links to the non-PR'd feedstocks be of value? It would be lovely to have a `TypedDict` of what's in those gx2 entries...

PR #774, opened 2020-01-29
more compact grpah

CJ-Wright:


CJ-Wright:
@regro/auto-tick

mbargull:
The graphs would still be rather wide and not fit well with the columnar viewport on the status page, right?

Maybe a simpler alternative:
```diff
-    gv = graphviz.Digraph()
+    gv = graphviz.Digraph(graph_attr={"packmode": "node"})
```
at https://github.com/regro/cf-scripts/blob/c2f6f7ca38522f9364438cb3d7961ffb093ef8ea/conda_forge_tick/auto_tick.py#L505.

If that's "too busy", `"packmode": f"array_{n}"` (with `n=3` maybe) could work.

Or (IDK if that works): Acc. to http://www.graphviz.org/doc/info/attrs.html#d:margin you could set `margin` for "clusters" to increase the space between connected components if you put the connected components in their own "cluster" subgraphs and use `"packmode": "cluster"`.

CJ-Wright:
PRs welcome!

mbargull:
> PRs welcome!

I kinda expected (not hoped) for that answer ;P.

I only need to clone this one, countyfair, and pinning repos to try it out, right?

CJ-Wright:
Sorry!

yes, countyfair and pinning need to live at the same level
```
top_dir
-- countyfair
-- pinning
```

PR #780, opened 2020-02-02
PR JSON on DynamoDB

scopatz:


conda_forge_tick/dynamo_models.py
```diff
@@ -0,0 +1,33 @@
...
+    def dump(cls, pr_json: dict):
+        """Dumps a PR JSON object to DynamoDB"""
+        attrs = dict(
+            id=pr_json["id"],
+            state=pr_json["state"],
+            merged_at=pr_json.get("merged_at"),
+            ETag=pr_json.get("ETag"),
+        )
+        if not isinstance(attrs["id"], int):
+            attrs["id"] = -uuid.UUID(attrs["id"]).int
```
beckermr: Why does this line have a minus sign?

CJ-Wright: This is for PRs we don't actually issue

beckermr: Huh?

scopatz: Because they are not valid github ids :(

beckermr: Right, but why are we tracking PRs we did not issue?

CJ-Wright: We didn't issue them because they were done manually before the bot got to them. So we need to know not to issue them. So we pretend that we did issue them so the bot passes it over.

beckermr: Ahhhhhh ok.

PR #784, opened 2020-02-05
fix doc on merge criteria

dougalsutherland:
I was pretty surprised by this line in the docs, implying that if e.g. the Mac build fails but Linux passes then it'll still get merged – but that's [not what the code does](https://github.com/regro/cf-autotick-bot-action/blob/master/conda_forge_tick_action/automerge.py#L73).


beckermr:
Thank you! Sorry yes, you are right!

There is a gate that a PR must have at least one passing status beyond the linter so that we don't merge PRs where the CI provider has not even reported anything yet.

beckermr:
Ideally, we would know which statuses we need and wait for them, but this is not implemented.

PR #800, opened 2020-02-14
ENH version migrator refactor

### LINKED ISSUE #430: Check why the bot failed to handle multiple hashes with pre-processor selectors
The bot failed to update the hashes properly in https://github.com/conda-forge/antlr-python-runtime-feedstock/pull/2

For more info see https://github.com/conda-forge/antlr-python-runtime-feedstock/pull/2#discussion_r243265221

### LINKED ISSUE #500: Incorect SHA256
The update pull request (https://github.com/conda-forge/geant4-feedstock/pull/1) generated an incorrect SHA256. The hash used by the bot does however match with what I get locally using `wget`:

```bash
$ wget https://geant4-data.web.cern.ch/geant4-data/releases/geant4.10.05.01.tar.gz
$ openssl sha256 geant4.10.05.p01.tar.gz
SHA256(geant4.10.05.p01.tar.gz)= f4a292220500fad17e0167ce3153e96e3410ecbe96284e572dc707f63523bdff
```

I'm not sure why `conda-build` is getting something different, any ideas?

### LINKED ISSUE #529: regro-cf-autotick-bot not pulling pypi sha
I'm not sure whether this is intended behavior or a bug, but it would be nice (at least for me) if the autotick-bot took the sha of the new version, for example:

https://github.com/conda-forge/issho-feedstock/pull/5/commits/fb6ef967685240e6f88ba55d306d1d457b748441 was pushed by the autotick-bot, but did not update the sha available at https://pypi.io/packages/source/i/issho/issho-0.5.1.tar.gz, the source url defined by the rendered meta.yaml.

### LINKED ISSUE #762: bot fails to update hashes for multiple source recipes on version updates
The bot failed to update hashes for Git 2.25 in [git-feedstock](https://github.com/conda-forge/git-feedstock/pull/73)

The same thing happened for Git 2.24, which I mentioned here:
https://github.com/regro/cf-scripts/issues/529#issuecomment-558418920

### LINKED ISSUE #781: Bot not updating hash in libprotobuf recipe with multiple sources
The bot recently make a PR to the libprotobuf feedstock, conda-forge/libprotobuf-feedstock#45, which did not update the hash of the source file.  In the past the bot was able to do this, e.g. conda-forge/libprotobuf-feedstock#44.

Note that this recipe uses [multiple sources](https://github.com/conda-forge/libprotobuf-feedstock/blob/master/recipe/meta.yaml#L7-L19) which may be the cause of the issue.  Only the hash of the first tarball needs to be updated.

### LINKED ISSUE #794: fix cb3multi version test
We were accidentally skipping a test for version migrations for cb3multi recipes. This needs fixing. Discovered in https://github.com/regro/cf-scripts/pull/793

beckermr:
This PR has a refactor of the version migrator using the parsing from #795. This approach should allow us to close out various version migration issues for complicated recipes.

Closes #794 
Closes #762 
Closes #781 
Closes #500
Closes #529 
Closes #430


tests/test_version_migrator.py
```diff
@@ -16,8 +16,7 @@
     ('multisrc', '2.4.1'),
     ('jinja2sha', '2.4.1'),
     ('r', '1.3_2'),
-    # FIXME - this test was being skipped by accident - now fails
-    # ('cb3multi', '6.0.0'),
+    ('cb3multi', '6.0.0'),
```
beckermr: And now this test passes!

tests/test_yaml/version_cb3multi.yaml
```diff
@@ -1,4 +1,3 @@
-{# cb3_multi #}
```
beckermr: ruamel.yaml cannot deal with jinja2 comments...

beckermr:
Test failures are ruamel.yaml being picky when it outputs yaml.

conda_forge_tick/migrators/version.py
```diff
@@ -1,136 +1,281 @@
...
+        if CONDA_SELECTOR in key:
+            selectors.append(key.split(CONDA_SELECTOR)[1])
+    for key in src:
+        if CONDA_SELECTOR in key:
+            selectors.append(key.split(CONDA_SELECTOR)[1])
+    return set(selectors)
+
+
+def _try_url_and_hash_it(url: str, hash_type: str):
+    print("    downloading url:", url)
```
mariusvniekerk: Probably want to change these out for logger.debug

tests/test_recipe_parser.py
```diff
@@ -0,0 +1,239 @@
...
+
+    # check selectors
+    assert cm.meta['source']['sha256__###conda-selector###__py2k'] == 1
+    assert cm.meta['source']['sha256__###conda-selector###__py3k and win'] == 5
+
+    # check other keys
+    assert cm.meta['build']['number'] == 10
+    assert cm.meta['package']['name'] == '{{ name|lower }}'
+    assert cm.meta['source']['url'] == 'foobar'
+
```
mariusvniekerk: I'd add a round trip test here as well

tests/test_yaml/version_jinja2sha.yaml
```diff
@@ -21,7 +21,7 @@ test:
 about:
   home: https://git-lfs.github.com/
   license: MIT
-  license_file: '{{ environ["RECIPE_DIR"] }}/LICENSE.md'
```
mariusvniekerk: These do appear in the wild like this though...

beckermr: Right. They are not an issue on input. When ruamel.yaml outputs this recipe it will remove the quotes since they are not needed. I change the input recipe but this is not needed. We use exact string matching on output so the change on the output version was needed for the tests.

conda_forge_tick/migrators/version.py
beckermr: done. ruamel.yaml cleans lint so I added a linted version of the input yaml

beckermr: done.

tests/test_recipe_parser.py
beckermr: ruamel.yaml cleans lint so I had to added a linted version of the input.

tests/test_yaml/version_jinja2sha.yaml
beckermr: I added some quotes back to the input for a few recipes.

PR #820, opened 2020-02-18
ENH check for upstream extensions, fix openssh

beckermr:
Reverts regro/cf-scripts#819

More WIP here


beckermr:
I reverted the commit to the graph from the latest run.

PR #848, opened 2020-02-26
Add mini-migrator to fix extra jinja2 variables

### LINKED ISSUE #847: make minimigrator to remove hash_type jinja2 keys
These recipes need fixing: https://github.com/search?q=org%3Aconda-forge+hash_type+filename%3Ameta.yaml&type=Code

djhoese:
Closes #847 

I kept going back and forth with how to do this (regular expressions, for loop looking at all keys at the same time, etc) and settled on this approach. I'm not 100% happy with it, but it works. I ended up spending too much time just trying to figure out how things work and thought it'd be better to get feedback before going any further. Let me know if you don't want this many jinja variables removed or if you want more tests.

CC @beckermr

### DIFF
--- a/conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
+++ b/conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
@@ -5,45 +5,63 @@ from ruamel.yaml import YAML
 
 from conda_forge_tick.xonsh_utils import indir
 from conda_forge_tick.migrators.core import MiniMigrator
+from conda_forge_tick.recipe_parser import CondaMetaYAML
 
 if typing.TYPE_CHECKING:
     from ..migrators_types import AttrsTypedDict
 
 
 class ExtraJinja2KeysCleanup(MiniMigrator):
-    # if you add a key here, you need to add it to the set of
-    # keys kept in the graph in make_graph.py
-    keys_to_remove = [
-        'min_r_ver',
-        'max_r_ver',
-        'min_py_ver',
-        'max_py_ver',
-        'compiler_stack',
-    ]
+    """Remove extra Jinja2 keys that make things hard for the version migrator.
+
+    Variables removed:
+
+    - ``hash_type``: This is sometimes used as a key in the ``source`` section
+      to map the hash_type (ex. sha256) to the hash.
+    - ``hash_value``: This is unnecessary since it is typically only used once
+      in a YAML recipe.
+    - ``file_ext``: This is unnecessary and just makes things harder to read.
+      It may also be causing trouble for the version migrator.
+
+    """
 
     def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
-        """only remove the keys if they are there"""
-        print(attrs, not_bad_str_start)
+        """Check the raw YAML to find jinja variables to deal with."""
+        raw_yaml = attrs['raw_meta_yaml']
+        if '{% set file_ext' in raw_yaml or '{% set hash_type' in raw_yaml:
+            return False
         return True
-        # cfy = attrs.get("conda-forge.yml", {})
-        # if any(k in cfy for k in self.keys_to_remove):
-        #     return False
-        # else:
-        #     return True
+
+    def _replace_jinja_key(self, key_name, lines):
+        """Replace any usage of and the definition of a jinja2 variable."""
+        key_val = None
+        var_use = "{{ " + key_name + " }}"
+        var_def = "{% set " + key_name
+        for line in lines:
+            if line.startswith(var_def):
+                quote1_idx = line.find('"')
+                quote2_idx = line.rfind('"')
+                key_val = line[quote1_idx+1:quote2_idx]
+                continue
+            if key_val is not None:
+                # in a string literal
+                line = line.replace("'" + var_use + "'", key_val)
+                line = line.replace(var_use, key_val)
+            yield line
 
     def migrate(self, recipe_dir: str, attrs: "AttrsTypedDict", **kwargs: Any) -> None:
-        print(recipe_dir, attrs, kwargs)
-        # with indir(recipe_dir):
-        #     cfg_path = os.path.join('..', 'conda-forge.yml')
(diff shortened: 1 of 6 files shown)


djhoese:
Main description updated. Thanks for your help @beckermr.

conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
```diff
@@ -0,0 +1,67 @@
...
+    """Remove extra Jinja2 keys that make things hard for the version migrator.
+
+    Variables removed:
+
+    - ``hash_type``: This is sometimes used as a key in the ``source`` section
+      to map the hash_type (ex. sha256) to the hash.
+    - ``hash_value``: This is unnecessary since it is typically only used once
+      in a YAML recipe.
+    - ``file_ext``: This is unnecessary and just makes things harder to read.
+      It may also be causing trouble for the version migrator.
```
scopatz: Should this be done as a migrator or something that the linter complains about?

scopatz: You have the migrator so it is fine :) But should we also prevent this from happening in the future with linter warnings?

conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
```diff
@@ -0,0 +1,67 @@
...
+
+    """
+
+    def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
+        """Check the raw YAML to find jinja variables to deal with."""
+        raw_yaml = attrs['raw_meta_yaml']
+        if '{% set file_ext' in raw_yaml or '{% set hash_type' in raw_yaml:
```
beckermr: Do you want to check for hash_value here too?

conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
```diff
@@ -0,0 +1,67 @@
...
+            with open('meta.yaml', 'r') as fp:
+                lines = fp.readlines()
+
+            lines = list(self._replace_jinja_key('file_ext', lines))
+            lines = list(self._replace_jinja_key('hash_type', lines))
+            lines = list(self._replace_jinja_key('hash_value', lines))
+
+            yaml = YAML(typ='jinja2')
+            yaml.indent(mapping=2, sequence=4, offset=2)
+            yaml.width = 120
```
beckermr: This yaml object is not used.

conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def _replace_jinja_key(self, key_name, lines):
+        """Replace any usage of and the definition of a jinja2 variable."""
+        key_val = None
+        var_use = "{{ " + key_name + " }}"
+        var_def = "{% set " + key_name
+        for line in lines:
+            if line.startswith(var_def):
+                quote1_idx = line.find('"')
+                quote2_idx = line.rfind('"')
```
beckermr: What about single quotes here?

conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
```diff
@@ -0,0 +1,67 @@
...
+                key_val = line[quote1_idx+1:quote2_idx]
+                continue
+            if key_val is not None:
+                # in a string literal
+                line = line.replace("'" + var_use + "'", key_val)
```
beckermr: I am nervous about the extra quotes here given that a string could be enclosed in double quotes. jinja2 only does string replace so I would remove the quotes.

conda_forge_tick/migrators/extra_jinj2a_keys_cleanup.py
djhoese: @beckermr and I had talked about making the auto tick bot create an issue if it fails to update a recipe. This was happening for some of my recipes where a new version came out but the feedstock showed nothing (no PR, no issue) from the bots.

The main unknown for me on linter warnings is I'm not exactly sure *why* this was a problem for the version migrator, just that it was. I think not allowing jinja2 variables as YAML keys would be an easy lint check to do. The hash_value and file_ext are just extra.

scopatz: Sounds good!

djhoese:
Given the single versus double quote stuff maybe I'll switch to regular expressions to avoid having to bruce force all the cases.

As for not using YAML tools for parsing the YAML, I agree but couldn't think of a good way to do it. I originally was using the ruamel YAML object but as mentioned in the related issue there is no way to remove jinja variables from there. I then switched to following the migrator in pip_check.py but noticed all the custom functions it had for parsing and changing selectors. If I parsed the YAML with ruamel, then iterated over the keys, I'd still have to do a string replace wouldn't I? And I'd have to do string comparisons to detect the jinja variable definition lines to not include, right?

djhoese:
So I've updated this to use a regular expression to find the definition of the jinja2 variable which simplifies the code a bit and handles the various cases of quote/no quote. I've also removed the "redundant" variable usage replacement (one with quotes, one without quotes around it). This should now be handled implicitly by ruamel when it decides whether or not something needs to be quoted.

This still doesn't use any YAML parsing package, but given the need to remove the jinja variable definition I'm not sure of a "clean" solution. The solutions I've thought of...

And of course now as I'm writing this comment I find the `_replace_jinja2_vars` function in [the parser code](https://github.com/regro/cf-scripts/blob/faa07556b0766ac2b2c3451d1be475e3ab3f31bf/conda_forge_tick/recipe_parser/_parser.py#L207). Maybe I could add the functionality to allow variables to be deleted here. Eh having to replace the usage of the variable later on in the file though wouldn't be much prettier though. Thoughts?

beckermr:
This looks good to me. Once we merge, we should babysit the bot a bit to make sure it is still working. I will do this tomorrow.

PR #849, opened 2020-02-27
REF run the piggyback migrations first

beckermr:
This PR runs the piggyback migrations first. It will help us address #847 and #791.


djhoese:
Small question: why the `swap_piggyback` in the tests?

beckermr:
I am removing this but the github api is being flaky

beckermr:
there is a commit already pushed where I split into pre and post migrators - it turns out the license minimigrator relies on the source of the package and so it has to be run after the version is updated. I did swap the default to run the minimigrators before the main one

beckermr:
there is a commit already pushed where I split into pre and post migrators - it turns out the license minimigrator relies on the source of the package and so it has to be run after the version is updated. I did swap the default to run the minimigrators before the main one

scopatz:
This LGTM!

beckermr:
It is not showing my latest commit.

PR #853, opened 2020-03-04
Consider package output names when doing migrations (#852)

### LINKED ISSUE #852: Packages from recipes with multple outputs cannot be migrated
As shown in `root_base` wasn't being migrated. https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/415

It issue is that `root_base` isn't in `gx.nodes` so it gets removed here:

https://github.com/regro/cf-scripts/blob/077c79ecadbd77891f2e7d2e3ee1480e159ddd06/conda_forge_tick/auto_tick.py#L437-L440

I guess this is because it is one of the outputs in the ROOT recipe `root`

cc @beckermr @duncanmmacleod

chrisburr:
Fixes #852 

I started adding a test but I didn't have enough time to finish but I've build the migrations locally and they now work as expected.


beckermr:
Let's just delete the yaml file and push the change. If you don't have time, then I don't think it makes sense to keep it.

conda_forge_tick/auto_tick.py
```diff
@@ -428,16 +428,29 @@ def migration_factory(
...
 
         migrator_config = loaded_yaml.get("__migrator", {})
         exclude_packages = set(migrator_config.get("exclude", []))
+
         package_names = (
             (set(loaded_yaml) | {l.replace("_", "-") for l in loaded_yaml})
-            & set(gx.nodes)
+            & all_package_names
         ) - exclude_packages
+        package_names = {p if p in gx.nodes else output_to_recipe for p in package_names}
```
beckermr: output_to_recipe is a dict here? What does this line do?

beckermr: Doesn't it need to be `output_to_recipe[p]`?

beckermr: Also we should subtract excluded stuff again.

beckermr:
```suggestion
        package_names = {p if p in gx.nodes else output_to_recipe[p] for p in package_names} - exclude_packages
```

PR #854, opened 2020-03-05
Better match package output names when doing migrations (#852)

chrisburr:
I made a mistake when testing #853 which prevents it from properly working (it requires **both** the output and the top level recipe name to be requirement instead of either).

Assuming I've tested it properly this time, this PR fixes the issue.

Before:
```
aarch64 and ppc64le addition graph size: 644
grpc127 graph size: 1
python38 graph size: 2025
root620000 graph size: 1
boost172 graph size: 165
matplotlib-to-matplotlib-base graph size: 228
```
After:
```
aarch64 and ppc64le addition graph size: 644
grpc127 graph size: 1
python38 graph size: 2025
root620000 graph size: 7  # <- 6 more packages found
boost172 graph size: 165
matplotlib-to-matplotlib-base graph size: 228
```

Finishes fixing #852

cc @beckermr @duncanmmacleod

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -37,6 +37,7 @@ from typing import (
     Tuple,
     Dict,
     Set,
+    Mapping,
     MutableMapping,
     Union,
 )
@@ -338,6 +339,7 @@ def add_rebuild_migration_yaml(
     migrators: MutableSequence[Migrator],
     gx: nx.DiGraph,
     package_names: Sequence[str],
+    output_to_recipe: Mapping[str, str],
     migration_yaml: str,
     config: dict = {},
     migration_name: str = "",
@@ -393,6 +395,11 @@ def add_rebuild_migration_yaml(
 
     # post plucking we can have several strange cases, lets remove all selfloops
     total_graph.remove_edges_from(nx.selfloop_edges(total_graph))
+
+    package_names = {
+        p if p in gx.nodes else output_to_recipe[p]
+        for p in package_names
+    }
 
     top_level = {
         node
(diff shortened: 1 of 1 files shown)


beckermr:
LOL I just figured this out myself. :)

chrisburr:
😆

Thanks for all the help getting this fixed! (and all the other things you've been doing recently)

conda_forge_tick/auto_tick.py
```diff
@@ -394,6 +396,11 @@ def add_rebuild_migration_yaml(
     # post plucking we can have several strange cases, lets remove all selfloops
     total_graph.remove_edges_from(nx.selfloop_edges(total_graph))
 
+    package_names = {
+        p if p in gx.nodes else output_to_recipe[p]
+        for p in package_names
+    }
```
CJ-Wright: What happened to exclude packages? That was important for the python migration?

chrisburr: It got dropped as I edited a outdated branch when debugging why #853 didn't work properly and the rebased badly.

The exclusions are still applied on line 459 so it still behaves in almost the same way. The only change is that if `exclude_packages` contains a package with multiple `outputs`, the output name will also have to be added to `exclude_packages`. For example if `matplotlib` and `matplotlib_base` both depend on `python` both would have to be excluded explicitly.

Thinking about it more, I guess this is probably just annoyingly verbose and `exclude_packages` is actually `exclude_feedstocks`? If you agree I'll make a PR.

CJ-Wright: That's an interesting point. I would go with `exclude_feedstocks`. Its usage is rare (eg to remove compilers from the python graph) so I think we can make the assumption that you want to remove the entire feedstock, since we don't have the capacity to issue PRs on individual outputs.

chrisburr: Done in https://github.com/regro/cf-scripts/pull/855 😃

PR #865, opened 2020-03-12
ENH pinning migrator 

### LINKED ISSUE #665: auto pin PRs
It would be nice if the bot issued a migration PR if it issued a version bump PR for a pinned package that caused the version to move out of the `run_exports` supported range. This PR then could be accepted or rejected, without requiring a human to make the PR. This would also help shift control of the ABI compatibility ranges to the maintainers of the feedstocks themselves, as changes to the `run_exports` and versions would automatically result in a migration proposal. Finally this would help with pins that we are missing, for example `arrow-cpp` was missing a pin until recently despite declaring a `run_exports`.

Proposed design:
1. When doing a version bump regex the `raw_meta_yaml` for `max_pin=...`, extract the max_pin.
2. Determine if the new version is outside the max_pin range.
3. If so produce a migration PR and issue that PR into the pinning feedstock (after issuing the version bump PR). Include in this PR a link to the Version bump PR and the number of packages impacted/plot of the graph for the migration.

Attn: @regro/auto-tick @wolfv @beckermr

### LINKED ISSUE #861: Use the installed CFP package instead of a git checkout

beckermr:
This PR has code from @CJ-Wright's one but finishes a few extra odds and ends.

closes #861 
closes #858 
closes #665

### DIFF
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -159,6 +159,7 @@ class MigrationYamlCreator(Migrator):
         self.new_pin_version = new_pin_version
         self.package_name = package_name
         self.bump_number = bump_number
+        self.name = package_name + " pinning"
 
     def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
         if attrs["name"] == "conda-forge-pinning":
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -174,12 +174,17 @@ class MigrationYamlCreator(Migrator):
             "migrator_ts": f"{time.time():.0f}",
         }
         with indir(os.path.join(recipe_dir, "migrations")):
-            with open(f"{self.package_name}{self.new_pin_version}.yaml", "w") as f:
-                yaml.dump(migration_yaml_dict, f)
+            mig_fname = "%s%s.yaml" % (
+                self.package_name,
+                self.new_pin_version.replace(".", "")
+            )
+            with open(mig_fname, "w") as f:
+                yaml.dump(
+                    migration_yaml_dict,
+                    f,
+                    default_flow_style=False,
+                )
             eval_xonsh("git add .")
-
-        with indir(recipe_dir):
-            self.set_build_number("meta.yaml")
 
         with indir(recipe_dir):
             with open("meta.yaml", "r") as fp:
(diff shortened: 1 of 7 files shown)


conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -127,3 +133,117 @@ def order(
...
+        with indir(recipe_dir):
+            self.set_build_number("meta.yaml")
+
+        with indir(recipe_dir):
+            with open("meta.yaml", "r") as fp:
+                new_lines = []
+                found_ver = False
+                for line in fp.readlines():
+                    if line.startswith("{% set version = "):
+                        now = datetime.datetime.now().strftime("%Y.%m.%d")
```
CJ-Wright: what if we make two PRs in the same day?

beckermr: Then the people who merge those PRs work that out, just like they do with the normal migration bot.

beckermr: Once the first one is merged, the second will get a conflict due to the build number bump.

beckermr:
This one is ready to merge @CJ-Wright. I added a test.

beckermr:
The parsing first extracting pins leaves a dictionary where there used to be a string in the list of requirements. This change breaks the whole code base which assumes a string.

CJ-Wright:
What parts use the run exports metadata?

beckermr:
You didn’t just chnage run exports. The pin compatible and pin sub package jinja functions are used in normal host build and run sections too.

beckermr:
It is the dict in those sections that is th e issue.

CJ-Wright:
Ahh ok sorry I totally forgot those are used elsewhere.

beckermr:
No worries! Good to merge?

conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -127,3 +133,142 @@ def order(
...
         )
+
+
+class MigrationYamlCreator(Migrator):
+    """Migrator creating migration yaml files."""
+
+    migrator_version = 0
+    rerender = False
+
+    # TODO: add a description kwarg for the status page at some point.
```
CJ-Wright: IDK if this needs to land on the status page since it will only ever have one PR issued.

conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -127,3 +133,142 @@ def order(
...
+    # TODO: make yaml_contents an arg?
```
CJ-Wright: Or just push all the kwargs into the migration

CJ-Wright:
Thank you very much!

PR #867, opened 2020-03-13
BUG prefer node names over outputs

### LINKED ISSUE #866: aarch pypy circular dep
It appears the pypy and pypy-meta nodes have a circular dep in the graph that shouldn't be there.

```
In [6]: with gx.nodes['pypy3.6']['payload'] as attrs: 
   ...:     print(m[0].filter(attrs)) 
   ...:                                                                                                                                                                   
missing: pypy-meta
parents
parent
True

In [7]: with gx.nodes['pypy-meta']['payload'] as attrs: 
   ...:     print(m[0].filter(attrs)) 
   ...:      
   ...:                                                                                                                                                                   
missing: pypy3.6
parents
parent
True
```

beckermr:
This change fixes the pyp36 -> pypy-meta -> pypy36 circular dep.

IDK if it should be general though. 

closes #866 

cc @isuruf @CJ-Wright

### DIFF
--- a/conda_forge_tick/make_graph.py
+++ b/conda_forge_tick/make_graph.py
@@ -300,7 +300,7 @@ def make_graph(names: List[str], gx: Optional[nx.DiGraph] = None) -> nx.DiGraph:
             # replace output package names with feedstock names via LUT
             deps = set(
                 map(
-                    lambda x: outputs_lut.get(x, x),
+                    lambda x: x if x in gx else outputs_lut.get(x, x),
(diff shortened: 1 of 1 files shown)


beckermr:
In fact it shouldn't. The problem here is that the metadata is not expressive enough. Unfortunately, with pypy-meta outputting python, a large portion of the graph will depend on it incorrectly.

beckermr:
OK closing. The right thing here is to map outputs to more than one feedstock and then require either one to be done in order to migrate the node.

PR #868, opened 2020-03-13
BUG check each dep instead of nodes

### LINKED ISSUE #866: aarch pypy circular dep

beckermr:
This fixes #866

### DIFF
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -547,6 +547,14 @@ class GraphMigrator(Migrator):
             else:
                 return False
 
+            if (
+                node not in self.graph.nodes and
+                node not in self.graph.predecessors(attrs["feedstock_name"])
+            ):
+                # this is ok because we would not have looked at it before
+                # anyways
+                continue
+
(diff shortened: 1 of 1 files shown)


conda_forge_tick/migrators/core.py
```diff
@@ -511,13 +511,51 @@ def __init__(
             self.graph = nx.DiGraph()
         else:
             self.graph = graph
+
+        # IDK if this will be there so I am going to make it if needed
```
CJ-Wright: This is made in make graph

PR #894, opened 2020-03-22
More verbose yaml creator

CJ-Wright:


beckermr:
@CJ-Wright do you want to compute the migration size here? I simple loop over the graph looking for the dep should do the trick to get a rough size

CJ-Wright:
I don't really want to do that in the PR body, since it doesn't have access to the graph.

CJ-Wright:
Maybe at that point we should, pull the `add_rebuild_migration_yaml` inner loop into a function that expects a loaded yaml file, pass the graph into the `Create...` migrator init, and then build the appropriate migrator and then do a filter. It's more work but if we're going to go through the effort of increasing the scope of the migrator to hold onto the graph then we might as well get it correct.

beckermr:
Sure or we can pass an approximate count as an argument to the constructor like you did here with the feedstock name.

CJ-Wright:
To be fair that approximate count could be quite removed from the actual count without filtering, for instance the number of packages that list python as a dep is far removed from the number of packages that actually need to be rebuilt.

beckermr:
How so?

CJ-Wright:
Many packages that have a python dep are noarch, which we don't rebuild when we rebuild python.

beckermr:
noarch is a special case

CJ-Wright:
Also packages that are only depended on as a run dependency, not build or host.
Pretty much this entire block of code: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/auto_tick.py#L392-L418

Most likely that block should be lifted into its own function and used.

beckermr:
Sure. That sounds like a good plan for this.

beckermr:
We can call that block to get an approximate count and then feed it in as an argument to go in the pr text.

CJ-Wright:
We still need to have the graph in hand. That wouldn't be an *approximate* count, that would be *the* count, it is what the migrator thinks needs to be migrated.

beckermr:
Right. We have the graph in auto_tick.py when we make the migrators. I’ll make a pr after this one.

CJ-Wright:
Great! I think this will be a good opportunity to pay down technical debt and provide the correct information where possible. The refactor work is most likely out of scope for this PR anyway.
I'll try to put up some refactors around the outputs to feedstock lookup, which should be pulled from the graph where possible (that way we keep everything consistent).

CJ-Wright:
Also in that case, would you like to merge?

beckermr:
Merge Away!

PR #897, opened 2020-03-24
requirements:run :same config removed

Prodyte:
extra 'requests' removed

### DIFF
--- a/requirements/run
+++ b/requirements/run
@@ -8,7 +8,6 @@ doctr
 rever
 xonsh
 conda-smithy
-requests
 feedparser
 conda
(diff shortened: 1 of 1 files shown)


requirements/run
```diff
@@ -8,7 +8,6 @@ doctr
 rever
 xonsh
 conda-smithy
-requests
```
beckermr: I only see one listing of requests here. Why are you removing this?

Prodyte: it's on line 6

beckermr: Omg I am so sorry! Thank you and my apologies!

PR #898, opened 2020-03-25
BUG if dep is self that is ok

beckermr:
This will fix the root migration


beckermr:
cc @CJ-Wright @chrisburr

PR #926, opened 2020-03-31
don't check diffed_files if PR into pinning, since rerender does nothing

CJ-Wright:


CJ-Wright:
we need to clean up the graph as well

conda_forge_tick/auto_tick.py
```diff
@@ -207,7 +207,7 @@ def run(
 
     # TODO: Better annotation here
     pr_json: typing.Union[MutableMapping, None, bool]
-    if isinstance(migrator, MigrationYaml) and not diffed_files:
+    if isinstance(migrator, MigrationYaml) and not diffed_files and feedstock_ctx.attrs['name'] != 'conda-forge-pinning':
```
beckermr: Do we need to limit this just to the MigrationYaml and not the other migrators?

CJ-Wright: Yes, rerender might not do anything for version updates or package replacements and that would be ok.

beckermr:
LGTM!

PR #934, opened 2020-03-31
use the correct directionality for the graph and then use the graph only for pinning

CJ-Wright:
This needs a fix to the graph (namely the PRs "issued" against the pinning)

### DIFF
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -568,10 +568,14 @@ class GraphMigrator(Migrator):
 
             return True
 
+        # Special case pinning so we don't need to add it as a dep for every feedstock
+        if attrs.get('name', '') == 'conda-forge-pinning':
+            all_deps = self.graph.predecessors('conda-forge-pinning')
         # check deps directly instead of using the graph
         # this breaks cycles in cases where a dep comes from more than
         # one package
-        all_deps = set().union(*attrs.get("requirements", {}).values())
+        else:
+            all_deps = set().union(*attrs.get("requirements", {}).values())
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@regro/auto-tick

beckermr:
LGTM. I guess we have to fix the graph too.

CJ-Wright:
For some reason this is still putting out a node for the pinning.

CJ-Wright:
I'll debug tomorrow

CJ-Wright:
@beckermr

conda_forge_tick/migrators/core.py
```diff
@@ -568,10 +568,14 @@ def _test_node_built(node):
 
             return True
 
+        # Special case pinning so we don't need to add it as a dep for every feedstock
```
beckermr, line 571: This comment is wrong. I think you mean to say we don't need to add every feedstock to its deps.

beckermr:
One comment here, but otherwise LGTM.

conda_forge_tick/migrators/core.py
CJ-Wright, line 571:
```suggestion
        # Special case pinning so we don't need to add every feedstock as a dep
```

CJ-Wright:
Ready for another round

beckermr:
what is your plan for fixing the graph? bot re-run labels? adjust the json by hand?

CJ-Wright:
For the PRs that we actually issued re-runs should be ok. For those we should have issued but didn't (eg. root) I will go in and delete the offending metadata.

PR #944, opened 2020-04-04
only pin down to the level of the spec

CJ-Wright:


CJ-Wright:
@beckermr @isuruf

beckermr:
I;m confused. The pin spec here ix x.x so the output yaml should have 1.99 in it. Right now it is wrong

https://github.com/regro/cf-scripts/pull/944/files/d9b00d9c5ada2a46c77e10cd78645f9f6cb793b4..c9d91101ba0f3f73d197ab3847ca5cbe727d1ce2#diff-fa45d9211b4ce58833ac70a9c0a032c5R118

beckermr:
Oh. you enforce the spec before it gets into the migrator. We should move that code to within the migrator so we can test that it works maybe. Or write a test for the code on the outside.

CJ-Wright:
@beckermr ready for another round

PR #935, opened 2020-03-31
list the archived packages for ease of debug

CJ-Wright:

Issue #599, opened 2019-09-10
We should make this consistent with the other uses of bad which use dicts.  Out of scope for this PR


mariusvniekerk:
We should make this consistent with the other uses of bad which use dicts.  Out of scope for this PR

_Originally posted by @mariusvniekerk in https://github.com/regro/cf-scripts/pull/598_

PR #955, opened 2020-04-08
ENH pull the active feedstocks from admin-migrations

### LINKED ISSUE #937: Archived holiday
We could run the `make_graph` update steps on archived packages once in a while (maybe every day?). This would slow us down a little but would make certain we catch feedstocks that were archived and have been revived or mistakenly archived.

CC: @duncanmmacleod

beckermr:
cc @CJ-Wright 

closes #937 
closes #935

### DIFF
--- a/conda_forge_tick/all_feedstocks.py
+++ b/conda_forge_tick/all_feedstocks.py
@@ -1,8 +1,6 @@
-import datetime
-import os
 from typing import Any, List
 
-import github3
+import requests
 import logging
 
 from .utils import setup_logger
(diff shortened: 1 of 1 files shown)


beckermr:
ready for review @CJ-Wright !

CJ-Wright:
Does this include archived feedstocks?

beckermr:
So the json slug has two entries, "active" and "archived". So the names here are all active feedstocks. This PR does resurrect archived feedstocks.

CJ-Wright:
How does it resurrect archived feedstocks?

beckermr:
Ack. I wrote too fast. It does NOT resurrect them. Sorry!

CJ-Wright:
LGTM

CJ-Wright:
So I'm happy to merge as is, but we could also take this PR farther if you want. At this point you might be able to get rid of that step entirely, plumbing the list int https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/make_graph.py#L464 directly

Do you want to do that?

I also like the direction this is taking the bot, moving away from the bot creating data and being more on the consuming side.

beckermr:
Yeah this had crossed my mind and also marking nodes as archived or not. 

Let's hold off on merging and we can tackle both of those things here. I gotta run for now.

CJ-Wright:
The archived question is a little problematic at the moment, since we use `archived: True` for a bunch of things, including stub packages (eg. `cxx_compiler_stub`). Most likely we should move archived under `bad` since they play similar roles, we'll just need to be more careful about only deleting parts of bad when remaking the graph.

beckermr:
Ah. Interesting. More explicit data attributes that don't carry more than one meaning in more than one spot would be helpful to make the code more understandable.

CJ-Wright:
Sure, hence the move to bad as a dict, with meaningful keys (and values).

```
bad = {'archived': True, 'stub': True, 'auto_tick': 'error message', 'make_graph': 'error message'}
```
(This is hypothetical, almost any bad entry would disqualify a package from further processing, except for a bad upstream in the case of a migration)

CJ-Wright:
xref: https://github.com/regro/cf-scripts/issues/599

beckermr:
Let’s hold off on this. GitHub is returning dashes in names when there should be underscores and I cannot tell why.

PR #961, opened 2020-04-14
use mamba to test if package req are solvable before sending PR

CJ-Wright:


isuruf:
This might be problematic. For example in the pypy migration we had cpython only packages that we closed the PR for. When the dependent package got a PR, I added a condition to make the dependency `cpython` only.

CJ-Wright:
Hmm, ok. How did you make the condition?

CJ-Wright:
In the short term we might make it something you can select on a per migration basis. It is needed for some version bumps (eg the boto stack) and for the R migration with autotick so we don't knowingly open bad PRs.

conda_forge_tick/migrators/core.py
```diff
@@ -287,6 +287,19 @@ def parse_already_pred() -> bool:
         if bad_attr:
             LOGGER.debug("%s: bnad attr" % __name)
 
+        # Check that all upstreams are installable
+        arch_packages = {}
+        # TODO: global variable this somewhere and add aarch and ppc
```
scopatz, line 292: Full list is actually:

```py
    platforms = [
        "noarch",
        "linux-64",
        "osx-64",
        "win-32",
        "win-64",
        "linux-armv7l",
        "linux-ppc64le",
        "linux-aarch64",
    ]
```

FWIW

CJ-Wright: That's what I needed (although mamba seems to know to add noarch to everything)

conda_forge_tick/migrators/core.py
```diff
@@ -287,6 +287,19 @@ def parse_already_pred() -> bool:
...
+        for arch in ['win-64', 'osx-64', 'linux-64']:
+            platform = arch.split('-')[0]
+            if attrs[platform].get('build', {}).get('skip', False):
+                continue
+            req = extract_requirements(attrs[platform])
+            # TODO: maybe throw in test?
+            arch_packages[arch] = set().union(req['host'], req['run'])
```
scopatz, line 299: build, host, run, and test are all solved independently, right? I don't mind unioning to get something working. But if we merge it might be good to make an issue.

conda_forge_tick/migrators/core.py
CJ-Wright: Also do we build win-32 anymore?

conda_forge_tick/utils.py
```diff
@@ -543,3 +544,27 @@ def as_iterable(iterable_or_scalar):
         return iterable_or_scalar
     else:
         return (iterable_or_scalar,)
+
+
+def mamba_cant_solve(packages, os_arch):
+
+    subprocess_list = ['mamba', 'install'] + packages + ['-c', f'https://conda.anaconda.org/conda-forge/{os_arch}', '--override-channels']
```
scopatz, line 551: There is a different way to set arches via env vars, but this is probably easier

conda_forge_tick/utils.py
```diff
@@ -543,3 +544,27 @@ def as_iterable(iterable_or_scalar):
         return iterable_or_scalar
     else:
         return (iterable_or_scalar,)
+
+
+def mamba_cant_solve(packages, os_arch):
+
```
scopatz, line 550: Is this suppossed to have a docstring?

conda_forge_tick/migrators/core.py
CJ-Wright: host and run should be solved together, and then when tests are to be run host run and test should be together I think?

(my understanding was that to run the code you needed host+run and then test was on top of that)

conda_forge_tick/utils.py
```diff
@@ -543,3 +544,27 @@ def as_iterable(iterable_or_scalar):
...
+    subprocess_list = ['mamba', 'install'] + packages + ['-c', f'https://conda.anaconda.org/conda-forge/{os_arch}', '--override-channels']
+    results = subprocess.check_output(subprocess_list)
```
scopatz, line 552: I think if you pass in `universal_newlines=True` it comes back as a string rather than bytes

conda_forge_tick/utils.py
CJ-Wright: I was looking for that in the docs, couldn't find it :(

conda_forge_tick/utils.py
```diff
@@ -543,3 +544,27 @@ def as_iterable(iterable_or_scalar):
...
+    return 'Problem' in results.decode('utf-8')
+
+
+def extract_requirements(meta_yaml):
```
scopatz, line 556: This should probably also have a docstring

conda_forge_tick/migrators/core.py
beckermr: We do not build win-32 iiuic.

CJ-Wright: Also how does this interact with selectors? (We don't have linux-64 as a selector, we have just linux).

conda_forge_tick/make_graph.py
```diff
@@ -119,10 +119,10 @@ def populate_feedstock_attributes(
                 "bot",
             }
         }
-
-    yaml_dict = ChainDB(
-        *[parse_meta_yaml(meta_yaml, platform=plat) for plat in ["win", "osx", "linux"]]
-    )
+    arches = ["win", "osx", "linux"]
```
CJ-Wright, line 122: This needs platform and arch

conda_forge_tick/migrators/core.py
CJ-Wright: Also do we build `armv7l`?

conda_forge_tick/auto_tick.py
```diff
@@ -450,7 +450,6 @@ def migration_factory(
         migrator_config = loaded_yaml.get("__migrator", {})
         excluded_feedstocks = set(migrator_config.get("exclude", []))
         pr_limit = min(migrator_config.pop("pr_limit", pr_limit), pr_limit)
-        check_solvability = migrator_config.get('check_solvability', True)
```
isuruf: Why remove this? We want this for pypy

CJ-Wright:
The current implementation might take too much time, I may need to use the mamba API directly to speed this up a bit.

conda_forge_tick/make_graph.py
```diff
@@ -119,10 +119,14 @@ def populate_feedstock_attributes(
                 "bot",
             }
         }
-
-    yaml_dict = ChainDB(
-        *[parse_meta_yaml(meta_yaml, platform=plat) for plat in ["win", "osx", "linux"]]
-    )
+    build_subdirs = ALL_BUILD_SUBDIRS if 'provider' in sub_graph['conda-forge.yml'] else MAIN_SUBDIRS
```
beckermr, line 122: This check is not quite right as people can list provider in feedstocks but still have no aarch packages. Also, some packages do not build windows.

conda_forge_tick/migrators/core.py
```diff
@@ -282,16 +284,30 @@ def parse_already_pred() -> bool:
 
         if attrs.get("archived", False):
             LOGGER.debug("%s: archived" % __name)
+            return True
 
-        bad_attr = _parse_bad_attr(attrs, not_bad_str_start)
-        if bad_attr:
-            LOGGER.debug("%s: bnad attr" % __name)
+        if _parse_bad_attr(attrs, not_bad_str_start):
+            LOGGER.debug("%s: bad attr" % __name)
```
beckermr, line 290: Thank you for the debug statements!

tests/test_extra_jinja2_keys_migrator.py
```diff
@@ -8,7 +8,7 @@
 
 from test_migrators import run_test_migration
 
-VERSION_CF = Version(piggy_back_migrations=[ExtraJinja2KeysCleanup()])
+VERSION_CF = Version(piggy_back_migrations=[ExtraJinja2KeysCleanup()], check_solvability=False)
```
beckermr, line 11: Can we use a default keyword to avoid having to put this everywhere?

CJ-Wright: The default is true. It's turned to false because these tests don't use real feedstocks

conda_forge_tick/make_graph.py
CJ-Wright: Can you give a case where the provider is used but a package is not built? The windows case is handled below with a skip check

tests/test_extra_jinja2_keys_migrator.py
beckermr: Ohhhh. I had assumed we would set the default to False to maintain backwards compat and then set it to true where we wanted. I'm happy either way though.

CJ-Wright: I thought we'd want to use it in more cases than not. It's set to false be cause most tests aren't backed by a real package. I can flip it if you want.

conda_forge_tick/auto_tick.py
CJ-Wright: We load the config in here: https://github.com/regro/cf-scripts/pull/961/files#diff-2be32f968602f542022009339eaac6afL422. The config dict will provide the `check_solvability` key for the pypy migration

conda_forge_tick/make_graph.py
beckermr: Any feedstock with provider: win: azure.

tests/test_extra_jinja2_keys_migrator.py
beckermr: nope it is fine as is.

conda_forge_tick/make_graph.py
beckermr: It's not so much if they exist but that they could exist and this would be totally fine.

CJ-Wright: Ah ok, so I need to do a bit more thorough check of `provider`. We might consider adding a provider for each build_subdir for good measure.

conda_forge_tick/utils.py
```diff
@@ -543,3 +544,27 @@ def as_iterable(iterable_or_scalar):
         return iterable_or_scalar
     else:
         return (iterable_or_scalar,)
+
+
+def mamba_cant_solve(packages, os_arch):
```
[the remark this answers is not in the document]
beckermr: We need to render the recipe with the new meta yaml for this to work, right?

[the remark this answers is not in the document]
beckermr: I mean the new cbc with the migration

CJ-Wright:
I'm going to hold off on this for now. I have it on good authority that the mamba API might support this better soon.

wolfv:
Hi CJ-Wright, 

as promised here is a small example on how to use the same repodata with mamba to query multiple times for different packages:

https://gist.github.com/wolfv/cd12bd4a448c77ff02368e97ffdf495a

Do you think that will work for you?

beckermr:
Thanks Wolf!

I think this pr needs call this code for each ci support script after rerendering. 

This way the new constraints and pins will show up properly when we try to solve.

wolfv:
If you want, you can set / remove the cache directory every time so that you can be assured that you download fresh repodata. 

Is that what you mean?

beckermr:
I think this means we need to feed the host and run specs after a call to conda build to the mamba solver as a list of strings. Is that right @wolfv?

beckermr:
Not quite. So when we run a migration, conda smithy combines the migration yaml with the pinnings file to produce a new conda build config in the .ci_support directory of the feedstock. It is this conda build config that is actually used to build the recipe by smithy. 

To check for solvability, we need to run mamba with the pins from this new config. 

Thus I think we need to render the recipe with conda build, extract the requirements lists, and then feed those to mamba to make sure the env can be made.

beckermr:
This pr is looking at requirements before we rerender and so I am not sure it is checking for the new pins.

CJ-Wright:
@wolfv this looks awesome! Thank you very much!

I think @beckermr is right we need to put in some more tooling around this to make it work properly.
I think there are two main use cases here: boto-core and R.

In boto-core most of the direct downstream libraries (boto3, etc.) have jinjia2 syntax to tightly couple their versions. To properly check if we are ready to bump the version of a downstream we need to perform the migration on the downstream, updating the downstream version number so that it get the correct pin to boto-core. This wouldn't formally require a full rerender.

The R case is a bit more complex, since to get the correct value for the R version we need to rerender with the latest pins. The rerender could be a costly process. Also I'm certain how we plug the pins into the actual requirements such that we get versions attached to packages.

CJ-Wright:
Also we most likely need to create a solver per build_subdir

beckermr:
@CJ-Wright let's put the check here

https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/auto_tick.py#L206 

within the context block. Let me push what I am thinking in a PR today for you all to look at.

CJ-Wright:
Won't that be a bit late? The filtering step would have already occurred.

wolfv:
> I think this means we need to feed the host and run specs after a call to conda build to the mamba solver as a list of strings. Is that right @wolfv?

yep a list of strings. you can normalize them by passing them through a conda MatchSpec and get the `.conda_build_form()`

E.g.:

from conda.models.match_spec import MatchSpec

normalized_spec = MatchSpec(my_string).conda_build_form()
```

CJ-Wright:
To be fair, we can do this as a last resort on the filter, so only nodes that are on the frontier according to the status of all the PRs would be checked for solvability. Although the filter doesn't have access to the entire repo like migrate does.

beckermr:
Sure. We’d still need to rerender. And then we’d rerender again after the filter call. Might as well only do that once.

CJ-Wright:
Part of me wonders if we should rerender anyway and push that up to the bot's fork and only open the PR when ready. That way we don't need to keep doing this.

beckermr:
If we want to do that we can make an api call to have GitHub actions do a rerender as opposed to the bot. Seems complicated though.

CJ-Wright:
Sorry that wasn't what I meant. More that, when we rerender the package it isn't wrong, just premature. We could do everything short of opening the PR then once the node is ready just open the PR. That way we don't waste cycles recomputing. I've also been thinking of if the bot would work well with a central queue of work to be done.

CJ-Wright:
Although we'd need to make sure there were no changes to the packages in the intervening time, which could be rough

beckermr:
> Although we'd need to make sure there were no changes to the packages in the intervening time, which could be rough

Yep. :/

PR #982, opened 2020-04-29
add msys2 into the mamba solve

CJ-Wright:


CJ-Wright:
cycle for CI

CJ-Wright:
@beckermr I think this will be needed for R

beckermr:
Yup. We should check the rest of the default channels while we are at it and make sure we have them. We might pull them from their source instead of hard coding things.

CJ-Wright:
pushes to this branch welcome!

PR #992, opened 2020-05-01
ENH more fixes for R

beckermr:
This PR has three things

1. We now check that the graph can be loaded before we deploy it. If it cannot, we deploy to a branch for debugging and stop the bot.

2. I added the ability to pause migrations.

3. I added the ability to patch the conda-forge.yml from the migrator via a syntax like

```yaml
conda_forge_yml_patches:
  provider.linux_ppc64le: azure
```

This will allow us to set this only for the R migration.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -477,9 +477,11 @@ def migration_factory(
     )
     for yaml_file, yaml_contents in migration_yamls:
         loaded_yaml = yaml.safe_load(yaml_contents)
-        print(os.path.splitext(yaml_file)[0])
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
So one minor comment.  Shouldn't 

```yaml
conda_forge_yml_patches:
  provider.linux_ppc64le: azure
```

rather be
```
conda_forge_yml_patches:
  provider:
     linux_ppc64le: 
        azure
```

beckermr:
@mariusvniekerk see the docs on the function. If you specify a dict like the second thing, it will replace the entire provider section. Instead, you use the `.` notation to replace just a single value.

PR #1025, opened 2020-05-30
remove the pinning node from the status

CJ-Wright:
@beckermr


CJ-Wright:
@beckermr I think this is ready for review

PR #1028, opened 2020-06-02
ENH compute archived from admin-migrations

beckermr:
This PR pulls a list of non-archived feedstocks from the admin-migrations repo. It then uses that list and correctly sets archived on each feedstock.


beckermr:
@CJ-Wright this is ready for a good look. I tested it locally and it works great

PR #1034, opened 2020-06-13
Run audits of grayskull

CJ-Wright:


conda_forge_tick/audit.py
```diff
@@ -36,41 +38,78 @@ def audit_feedstock(fctx: FeedstockContext, ctx: MigratorSessionContext):
     return deps
 
 
+def grayskull_audit_feedstock(fctx: FeedstockContext, ctx: MigratorSessionContext):
+    """Uses grayskull to audit the requirements for a python package
+    """
+    # TODO: come back to this, since CF <-> PyPI is not one-to-one and onto
```
beckermr, line 44: This is a tricky issue. I think we need to render the recipe and extract the url. Then we can pull out the pypi name.

CJ-Wright: maybe? I would be interested in actually having metadata that states what the name of the package is in other package managers

conda_forge_tick/audit.py
```diff
@@ -36,41 +38,78 @@ def audit_feedstock(fctx: FeedstockContext, ctx: MigratorSessionContext):
...
+    pkg_name = fctx.attrs["name"]
+    pkg_version = fctx.attrs["version"]
+    recipe = GrayskullFactory.create_recipe(
```
marcelotrevisani, line 47: If you want to , you can just generate the data without dumping it to a file
there are some examples in the file ``test_pypi.py``
```python
from grayskull.pypi import PyPi

recipe = PyPi(pkg_name, pkg_version)
```

and you can access all the information there, such as
```python
for host_req in recipe["requirements"]["host"]:
    print(host_req)

for run_req in recipe["requirements"]["run"]:
    print(run_req)
```

CJ-Wright: I kinda wanted the full `meta.yaml`, since grayskull would do more than replace the deps, right?

marcelotrevisani: Yes, but everything should be available to be accessed through that object
for example
```
print(recipe["about"]["home"])
print(recipe["about"]["license"])
print(recipe["extra"]["recipe-maintainers"])
```

But for the first analysis the file generation should be interesting to see :)

PR #1039, opened 2020-06-14
Fix the arch migration

CJ-Wright:
Use the arch requirements to build the graph


CJ-Wright:
this needs a bunch of testing

CJ-Wright:
I tested this on my local copy of the graph, it seems to do the correct thing.

PR #1041, opened 2020-06-17
run pre-commit as a GHA

ocefpaf:
This adds a GitHub Action as a CI service to check for the pre-commit hooks. It should help when reviewing style (black) and other hooks present in the `.pre-commit-config.yaml` file. Note that, at the moment, there are massive failures with the `mypy` and `check-yaml` hooks. Not sure they are real failures though.

PS: I can send a new commit fixing the style but I wanted to ask if that is OK first b/c it will be a massive commit.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,20 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+    - name: set PY
+      run: echo "::set-env name=PY::$(python --version --version | sha256sum | cut -d' ' -f1)"
+    - uses: actions/cache@v1
+      with:
+        path: ~/.cache/pre-commit
+        key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+    - uses: pre-commit/action@v1.0.0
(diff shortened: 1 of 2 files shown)


beckermr:
I’d vote to turn off mypy. Our type implementation and use is not complete. I don’t personally want to maintain it.

CJ-Wright:
Also does this mean we should remove the black check from travis?

beckermr:
I’d leave it. It costs us nothing.

CJ-Wright:
My understanding is that GHA will be triggered on merge to master, which is after the travis black check will have failed. (maybe I don't understand what is going on)

beckermr:
Ahhh. Well then turn it off?

ocefpaf:
> My understanding is that GHA will be triggered on merge to master,

Not sure I follow. Do you mean this one I added? If so, I believe not.

> which is after the travis black check will have failed. (maybe I don't understand what is going on)

This one works as a check for all pre-commit, not only black. The mains advantage IMO is that we can easily guide contributors to get this test passing my running using pre-commit. If they don't want to, this GHA will act as a CI to guide them to what needs fixing.

PS: feel free to close it if you don't want this. It should help reviews but you should do whatever you feel it is easier for you here.

beckermr:
Ahhh this duplicates the black check. Then we should leave both. 

@ocefpaf did you remove mypy from precommit here? We should do that

ocefpaf:
> Ahhh this duplicates the black check. Then we should leave both.

This is a bit cleaner b/c it is easier to navigate the logs, specially if we remove the yaml checks and mypy b/c they are failing filling up the logs at the moment.

We can also add a matrix of jobs in Travis-CI to make the black check easier to read.
 
> @ocefpaf did you remove mypy from precommit here? We should do that

Just did.

beckermr:
Thanks! I’m good with whatever here.

PR #1048, opened 2020-06-28
Bug correction of `new_update_versions`

viniciusdc:
Some bad things were happening on Circle, this is the first attempt to solve some of the bugs. Let's see what happens next.

### DIFF
--- a/conda_forge_tick/new_update_versions.py
+++ b/conda_forge_tick/new_update_versions.py
@@ -107,13 +107,14 @@ def new_update_upstream_versions(
                 logger.info(
                     f"# {node_count:<5} - {node} - {attrs.get('version')} - {attrs.get('new_version')}",
                 )
-            up_to[f"{node}"] = {
-                "bad": attrs.get("bad"),
-                "new_version": attrs.get("new_version"),
-            }
+
             logger.debug("writing out file")
             with open(f"versions/{node}.json", "w") as outfile:
-                json.dump(to_update, outfile)
+                up_to[f"{node}"] = {
+                    "bad": attrs.get("bad"),
+                    "new_version": attrs.get("new_version"),
+                }
+                json.dump(up_to, outfile)
             node_count += 1
     return up_to
 
--- a/conda_forge_tick/new_update_versions.py
+++ b/conda_forge_tick/new_update_versions.py
@@ -85,9 +85,6 @@ def new_update_upstream_versions(
         with node_attrs as attrs:
             up_to = {}
 
-            # verify the actual situation of the package;
-            actual_ver = str(attrs.get("version"))
-
             # avoid
             if node == "ca-policy-lcg":
                 up_to["ca-policy-lcg"] = {"bad": attrs.get("bad"), "new_version": False}
--- a/conda_forge_tick/new_update_versions.py
+++ b/conda_forge_tick/new_update_versions.py
@@ -70,15 +70,14 @@ def new_update_upstream_versions(
 
     # Inspection the graph object and node update:
     # print(f"Number of nodes: {len(gx.nodes)}")
-    Node_count = 0
-
+    node_count = 0
     to_update = []
     for node, node_attrs in tqdm.tqdm(_all_nodes):
-        attrs = node_attrs["payload"]
-        if attrs.get("bad") or attrs.get("archived"):
-            attrs["new_version"] = False
-            continue
-        to_update.append((node, attrs))
+        with node_attrs["payload"] as attrs:
+            if attrs.get("bad") or attrs.get("archived"):
+                attrs["new_version"] = False
+                continue
+            to_update.append((node, attrs))
 
     for node, node_attrs in to_update:
         # checking each node
@@ -88,7 +87,7 @@ def new_update_upstream_versions(
             # avoid
             if node == "ca-policy-lcg":
                 up_to["ca-policy-lcg"] = {"bad": attrs.get("bad"), "new_version": False}
-                Node_count += 1
+                node_count += 1
                 continue
 
             # New version request
@@ -106,7 +105,7 @@ def new_update_upstream_versions(
                 attrs["bad"] = "Upstream: Error getting upstream version"
             else:
                 logger.info(
-                    f"# {Node_count:<5} - {node} - {attrs.get('version')} - {attrs.get('new_version')}",
+                    f"# {node_count:<5} - {node} - {attrs.get('version')} - {attrs.get('new_version')}",
                 )
             up_to[f"{node}"] = {
                 "bad": attrs.get("bad"),
@@ -115,7 +114,7 @@ def new_update_upstream_versions(
             logger.debug("writing out file")
             with open(f"versions/{node}.json", "w") as outfile:
                 json.dump(to_update, outfile)
-            Node_count += 1
+            node_count += 1
     return up_to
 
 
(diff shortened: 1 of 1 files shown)


conda_forge_tick/new_update_versions.py
```diff
@@ -45,10 +45,10 @@ def get_latest_version(
                 return ver
             else:
                 logger.debug(f"Upstream: Could not find version on {source.name}")
-                meta["bad"] = f"Upstream: Could not find version on {source.name}"
+                meta_yaml["bad"] = f"Upstream: Could not find version on {source.name}"
```
CJ-Wright, line 48: This is writing to the nodes, please don't do that.

conda_forge_tick/new_update_versions.py
```diff
@@ -70,29 +70,24 @@ def new_update_upstream_versions(
...
     to_update = []
     for node, node_attrs in tqdm.tqdm(_all_nodes):
-        attrs = node_attrs["payload"]
-        if attrs.get("bad") or attrs.get("archived"):
-            attrs["new_version"] = False
-            continue
-        to_update.append((node, attrs))
+        with node_attrs["payload"] as attrs:
+            if attrs.get("bad") or attrs.get("archived"):
+                attrs["new_version"] = False
```
CJ-Wright, line 78: This will also write to the node

conda_forge_tick/new_update_versions.py
```diff
@@ -110,17 +105,17 @@ def new_update_upstream_versions(
...
-                "bad": attrs.get("bad"),
-                "new_version": attrs.get("new_version"),
-            }
+
             logger.debug("writing out file")
             with open(f"versions/{node}.json", "w") as outfile:
-                json.dump(to_update, outfile)
-            Node_count += 1
-    return up_to
+                up_to[f"{node}"] = {
```
CJ-Wright, line 113: why does `up_to` need to have the node name as a key?

viniciusdc:
Requested changes were made :+1:

conda_forge_tick/new_update_versions.py
```diff
@@ -70,35 +73,34 @@ def new_update_upstream_versions(
...
     to_update = []
     for node, node_attrs in tqdm.tqdm(_all_nodes):
-        attrs = node_attrs["payload"]
-        if attrs.get("bad") or attrs.get("archived"):
-            attrs["new_version"] = False
-            continue
-        to_update.append((node, attrs))
+        with node_attrs["payload"] as attrs:
+            if attrs.get("bad") or attrs.get("archived"):
+                # "new_version" = False
```
CJ-Wright, line 81: please remove unused code

conda_forge_tick/new_update_versions.py
```diff
@@ -30,7 +30,10 @@
 
 
 def get_latest_version(
-    name: str, payload_meta_yaml: Any, sources: Iterable[AbstractSource],
+    name: str,
+    payload_meta_yaml: Any,
+    sources: Iterable[AbstractSource],
+    versions: dict,
```
CJ-Wright, line 36: you might consider having this return a dict rather than mutating one.

conda_forge_tick/new_update_versions.py
```diff
@@ -70,35 +73,34 @@ def new_update_upstream_versions(
...
 
-            # verify the actual situation of the package;
-            actual_ver = str(attrs.get("version"))
-
-            # rude exception
+            # avoid
             if node == "ca-policy-lcg":
-                up_to["ca-policy-lcg"] = {"bad": attrs.get("bad"), "new_version": False}
-                Node_count += 1
+                up_to["new_version"] = False
```
CJ-Wright, line 92: I'm not certain if this is the most descriptive variable name

conda_forge_tick/new_update_versions.py
```diff
@@ -70,35 +73,34 @@ def new_update_upstream_versions(
...
                 continue
 
             # New version request
             try:
-                new_version = get_latest_version(node, attrs, sources)
-                attrs["new_version"] = new_version or attrs["new_version"]
+                # check for latest version
+                new_version = (
+                    get_latest_version(node, attrs, sources, up_to)
+                    or attrs["new_version"]
```
CJ-Wright, line 101: idk if we want to use the existing `new_version` if we don't get a version then the file should say so.

viniciusdc: Could it be replaced by False ?

conda_forge_tick/new_update_versions.py
viniciusdc: I leave it there to serve as a comment for what's happening, but it's better to remove

conda_forge_tick/new_update_versions.py
viniciusdc: This way I can attach the `bad` string when it occurs, but I will try another approach

conda_forge_tick/new_update_versions.py
viniciusdc: hha sorry about that, what about `version_data` ?

conda_forge_tick/new_update_versions.py
CJ-Wright: I would have it return an empty dict or not write out the file at all (if it would be empty).

conda_forge_tick/new_update_versions.py
CJ-Wright: sure

conda_forge_tick/new_update_versions.py
CJ-Wright: If you want to leave a descriptive comment then I would use prose rather than commented out code.

conda_forge_tick/new_update_versions.py
viniciusdc: I've added a dict containing the `ver` output and `bad` error string (when it occurs) inside `get_latest_version`. Then update using `version_data.update(get_latest_version(args))`

viniciusdc:
BTW why codecov/path always fail ?

conda_forge_tick/new_update_versions.py
```diff
@@ -78,29 +81,24 @@ def new_update_upstream_versions(
...
             to_update.append((node, attrs))
 
     for node, node_attrs in to_update:
         # checking each node
         with node_attrs as attrs:
-            up_to = {}
+            version_data = {}
 
             # avoid
             if node == "ca-policy-lcg":
```
CJ-Wright, line 93: you might consider putting this clause in the `get_latest_version` function.

CJ-Wright:
are your changes tested? If not then that is why it fails

conda_forge_tick/new_update_versions.py
viniciusdc: the `ca-policy-lcg` ?

viniciusdc:
> are your changes tested? If not then that is why it fails

There is a specific way to test it?

conda_forge_tick/new_update_versions.py
CJ-Wright: Yes, that carve out is not part of the for loop but a special case that should be in the `get_latest_version` function

viniciusdc: Oh, now I see. I will commit it later on the pool PR updates

PR #1049, opened 2020-06-29
Attempt to Update upstream version process pool 

viniciusdc:
Due to the expensive time consuming usage of the actual `update upstream version` I've inserted the `pool` mode, as also, implements the function to choose the mode accordingly to the current `CONDA_FORGE_TICK_DEBUG` status.
See as references the issue [#842 ] and PR's [#1048 , #1043 ].

Issue #842, opened 2020-02-25
separate version updates to their own microservice


beckermr:
It is time to split out the version update cron job to its own service. 

- For now, we will reuse the LazyJson infrastructure to build out a new version update table stored in a new cf-graph-version repo. (I had mentioned putting it in the cf-graph-countyfair repo, but then we have to deal with git pull and push issues, which we can punt on using a separate repo.)
- Eventually, this will be moved into dynamodb in another table.
- The JSON model will be a list of version blobs per feedstock.
- This list will be flattened by version string in the final dynamodb table.
- We will run it at half past the hour with the hope that it finishes by the time the bot starts, but it doesn't have to. 

cc @CJ-Wright @mariusvniekerk @scopatz

beckermr:
The entire script is here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/update_upstream_versions.py

The key is to move the data structures to another repo with a separate process.

viniciusdc:
I think I can do it, but it will take some time, cause I have to understand the operation of the entire structure that is specified in the script mentioned. And there is a big number of things that i still don't know, of course i may be able to learn it.

beckermr:
I would start simpler. 

1. Read the graph (clone cf-graph-county-fair, use --depth=1, cd to that dir)
2. Load it (call gx = load_graph(), see the main function)
3. call the update versions function in main
4. Inside the update versions function, write the potentially new version for each package to some json in a dir

Once that works, post a PR with the code and then we can move on from there.

viniciusdc:
Ok, I will try that.

viniciusdc:
Ok, now should i work on the first code you sent , or continue with the simple tasks (3) "call the update versions function in main" ?

beckermr:
continue on 3!

CJ-Wright:
@beckermr is it worthwhile to put this into the same Circle system? I think we can have multiple cron jobs offset from each other.

beckermr:
Yep that is fine by me! Great idea!

viniciusdc:
well i got something, but i think that i missed something in the way...
`RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.
[...]

PR #1050, opened 2020-06-30
make_graph update versions from .json

viniciusdc:
Continuing the work from [separate version updates](https://github.com/regro/cf-scripts/issues/842) and [issue](https://github.com/regro/cf-scripts/pull/1049).

### DIFF
--- a/conda_forge_tick/make_graph.py
+++ b/conda_forge_tick/make_graph.py
@@ -3,6 +3,7 @@ import collections.abc
 import hashlib
 import logging
 import os
+import json
 import time
 from collections import defaultdict
 from concurrent.futures import as_completed
@@ -38,7 +39,6 @@ if typing.TYPE_CHECKING:
 logger = logging.getLogger("conda_forge_tick.make_graph")
 pin_sep_pat = re.compile(r" |>|<|=|\[")
 
-
 NUM_GITHUB_THREADS = 2
 
 github_username = env.get("USERNAME", "")
@@ -70,16 +70,16 @@ def _fetch_file(name: str, filepath: str) -> typing.Union[str, Response]:
 
 # TODO: include other files like build_sh
 def populate_feedstock_attributes(
-    name: str,
-    sub_graph: LazyJson,
-    meta_yaml: typing.Union[str, Response] = "",
-    conda_forge_yaml: typing.Union[str, Response] = "",
-    mark_not_archived=False,
-    # build_sh: typing.Union[str, Response] = "",
-    # pre_unlink: typing.Union[str, Response] = "",
-    # post_link: typing.Union[str, Response] = "",
-    # pre_link: typing.Union[str, Response] = "",
-    # activate: typing.Union[str, Response] = "",
+        name: str,
+        sub_graph: LazyJson,
+        meta_yaml: typing.Union[str, Response] = "",
+        conda_forge_yaml: typing.Union[str, Response] = "",
+        mark_not_archived=False,
+        # build_sh: typing.Union[str, Response] = "",
+        # pre_unlink: typing.Union[str, Response] = "",
+        # post_link: typing.Union[str, Response] = "",
+        # pre_link: typing.Union[str, Response] = "",
+        # activate: typing.Union[str, Response] = "",
 ) -> LazyJson:
     """Parse the various configuration information into something usable
 
@@ -105,15 +105,15 @@ def populate_feedstock_attributes(
             k: v
             for k, v in yaml.safe_load(conda_forge_yaml).items()
             if k
-            in {
-                "provider",
-                "min_r_ver",
-                "min_py_ver",
-                "max_py_ver",
-                "max_r_ver",
-                "compiler_stack",
-                "bot",
-            }
+               in {
+                   "provider",
+                   "min_r_ver",
+                   "min_py_ver",
+                   "max_py_ver",
+                   "max_r_ver",
+                   "compiler_stack",
+                   "bot",
+               }
         }
 
     yaml_dict = ChainDB(
@@ -205,7 +205,7 @@ def get_attrs(name: str, i: int, mark_not_archived=False) -> LazyJson:
 
 
 def _build_graph_process_pool(
-    gx: nx.DiGraph, names: List[str], new_names: List[str], mark_not_archived=False,
+        gx: nx.DiGraph, names: List[str], new_names: List[str], mark_not_archived=False,
 ) -> None:
     with executor("thread", max_workers=20) as pool:
         futures = {
@@ -242,7 +242,7 @@ def _build_graph_process_pool(
 
 
 def _build_graph_sequential(
-    gx: nx.DiGraph, names: List[str], new_names: List[str], mark_not_archived=False,
+        gx: nx.DiGraph, names: List[str], new_names: List[str], mark_not_archived=False,
 ) -> None:
     for i, name in enumerate(names):
         try:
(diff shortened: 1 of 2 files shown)


conda_forge_tick/make_graph.py
```diff
@@ -352,6 +352,25 @@ def update_nodes_with_bot_rerun(gx):
...
+    """Updates every node with it's new version (when available)"""
+    # check if the versions folder is available
+    if os.path.isdir("./versions"):
+        # get all the available node.json files
+        pass
+    else:
+        return
+    list_files = os.listdir("./versions/")
+    for file in list_files:
+        node = str(file).replace(".json", "")
```
CJ-Wright, line 365: why are you removing `.json`?

conda_forge_tick/make_graph.py
```diff
@@ -352,6 +352,25 @@ def update_nodes_with_bot_rerun(gx):
...
+        with open(f"./versions/{file}") as json_file:
+            version_data = json.load(json_file)
+        with gx.nodes[f"{node}"]["payload"] as attrs:
+            attrs["new_version"] = version_data.get("new_version")
```
CJ-Wright, line 369: `attrs.update(**version_data))`

CJ-Wright: If that doesn't work please add an update method to `LazyJSON`

conda_forge_tick/make_graph.py
beckermr: To compute the node name in the graph I think

viniciusdc: To get the node name to update the attrs information regarding that node

conda_forge_tick/make_graph.py
viniciusdc: I will try that, but I will have to think a way to test that though

viniciusdc: Should I add the bad feedstocks too for `update_versions` ? to solve the problems with the outdated `.json` files ?

viniciusdc: Or let it run for now and see that after ?

viniciusdc: I was looking into the `LazyJSON` config at `.utils` and there isn't any `.update` method there, so would this be enough ?
```python
    def __update__(self, data: dict) -> None:
        for key in data.items():
            self.__dict__[f'{key}'] = data[f'{key}']
```

conda_forge_tick/utils.py
```diff
@@ -182,6 +180,11 @@ def __enter__(self) -> "LazyJson":
     def __exit__(self, *args: Any) -> Any:
         self._dump()
 
+    def __update__(self, data: dict) -> None:
```
CJ-Wright, line 183: I don't think this needs to be `__update__` it should be `update`

conda_forge_tick/utils.py
```diff
@@ -182,6 +180,11 @@ def __enter__(self) -> "LazyJson":
...
+        self._load()
+        for key in data.items():
+            self.data[f"{key}"] = data[f"{key}"]
```
CJ-Wright, line 186: I would use the object's `setitem`: `self[key] = data[key]` that way you don't need to do the load and dump

conda_forge_tick/utils.py
viniciusdc: I will change that thanks :+1:

conda_forge_tick/utils.py
viniciusdc: uhm, you are right

CJ-Wright: I would just use the `[]` syntax directly. Also you don't need an f-string for the keys, they are already strings.

viniciusdc: > I would just use the `[]` syntax directly. Also you don't need an f-string for the keys, they are already strings.

Oh thanks, I was not sure about the syntax usage.

conda_forge_tick/utils.py
```diff
@@ -180,10 +180,9 @@ def __enter__(self) -> "LazyJson":
     def __exit__(self, *args: Any) -> Any:
         self._dump()
 
-    def __update__(self, data: dict) -> None:
-        self._load()
-        for key in data.items():
-            self.data[f"{key}"] = data[f"{key}"]
+    def _update(self, data: dict) -> None:
```
beckermr, line 183: why did you rename this?

viniciusdc: It was an idea from CJ: "I don't think this needs to be `__update__` it should be `update`"

beckermr: you have an extra underscore at the front.

viniciusdc: Yup, I saw with your comment thanks 👍

viniciusdc: Uhm @beckermr I am getting a problem with this statement `update(self, data: dict)` it's saying that it doesn't match the method base definition of a `MutableMapping` wich should be something as `update(self, data: dict, **kwargs)`. Do you know what can I do to solve this ?

CJ-Wright: You could structure this as:
```python
def update(self, data, **kwargs):
    self._load()
    self._data.update(data, **kwargs)
    self._dump()
```

viniciusdc: Ok, Thank you very much 👍

CJ-Wright:
It seems that the tests aren't passing

viniciusdc:
> It seems that the tests aren't passing

Yes, the `tests/test_audit.py::test_grayskull_audit_feedstock` failed

CJ-Wright:
@viniciusdc I think that is due to the new grayskull release

CJ-Wright:
This has conflicts now (although I think I fixed the grayskull issue, at least for the moment)

viniciusdc:
> This has conflicts now (although I think I fixed the grayskull issue, at least for the moment)

I will try to solve the conflicts in a pull, if it's not enough I will create a new PR with this changes.

viniciusdc:
Sorry for the delay, not enough time this weekend. I will try to solve this ASAP.

viniciusdc:
Solved! 😄 (I think)

CJ-Wright:
It seems the build is failing

viniciusdc:
> It seems the build is failing

It is the same error as before `test_grayskull_audit_feedstock` but I don't know what is wrong, there are some instances as:
```python
/tmp/grayskull-depfinder-zbc3slmg/depfinder-2.3.0/versioneer.py:969: DeprecationWarning: invalid escape sequence \s
    '''
```

CJ-Wright:
Ahh, can you merge with master and fix the grayskull test? (There's a commented out section that you'll want to uncomment and remove the currently uncommented seciton)

viniciusdc:
> Ahh, can you merge with master and fix the grayskull test? (There's a commented out section that you'll want to uncomment and remove the currently uncommented seciton)

Yes I can do it, but where is Grayskull test files ? and the section I need to uncomment and merge is easy to find ? (I am asking this not for easiness, is just to do be sure I will do the right thing)

CJ-Wright:
yes that's the right file. The comments are right at the top

viniciusdc:
> yes that's the right file. The comments are right at the top

Is this regro/cf-scripts#1056 ?

viniciusdc:
Really we are gonna test it ? Oh it's so interesting to see if it works 😃

PR #1058, opened 2020-07-06
removed conflicting '.update' method

viniciusdc:
As an attempt to solve this [error](https://app.circleci.com/pipelines/github/regro/circle_worker/11498/workflows/35c9982e-4d91-4b0d-9438-2bb7b26859f8/jobs/27967) 
```python
TypeError: update() missing 1 required positional argument: 'data'
```

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -180,11 +180,6 @@ class LazyJson(MutableMapping):
     def __exit__(self, *args: Any) -> Any:
         self._dump()
 
-    def update(self, data: dict, **kwargs) -> None:
-        self._load()
-        self.data.update(data, **kwargs)
-        self._dump()
-
 
 def render_meta_yaml(text: str, for_pinning=False, **kwargs) -> str:
     """Render the meta.yaml with Jinja2 variables.


viniciusdc:
There should be no other issues with the rest of the environment cause we are restoring the modification made to this file. Also, there will be no problem with the new `make_graph` features cause the `.update` method seen to already exists.

CJ-Wright:
Ahh ok, so this is getting update from the mutable map?

viniciusdc:
> In particular the update method for mutable map will not afaik dump to disk. If we rely on the dumping somewhere then we need to reoverride this or deal with the different cases.

yup, you are right. But I would like to test if the `attrs.update(data)` on `make_graph` will work well even without the addition of a new method. If it's not the case that I will think of another way to do so.

viniciusdc:
> Let’s make sure the new method or whatever we are replacing this with does the same thing.

Though I am not replacing just reverting for the original file. What I am trying to avoid is the conflict between some initial `.update` method and that new one.

viniciusdc:
> Ahh ok, so this is getting update from the mutable map?

Yes, at least it's what I am expecting to see .

viniciusdc:
> In particular the update method for mutable map will not afaik dump to disk. If we rely on the dumping somewhere then we need to reoverride this or deal with the different cases.

What I am expecting to see is the same behavior of this [`pr_json.update(**res)`](https://github.com/regro/cf-scripts/blob/8389ec33c2e7116aab8bc10bfbea6eb479781a60/conda_forge_tick/update_prs.py#L131) from the `MutableMap`  happening here ['attrs.update(version_data)'](https://github.com/regro/cf-scripts/blob/8389ec33c2e7116aab8bc10bfbea6eb479781a60/conda_forge_tick/make_graph.py#L355)

viniciusdc:
Well the `update_PR_JSON` phase went well this [time](https://app.circleci.com/pipelines/github/regro/circle_worker/11514/workflows/b0e97100-1844-466c-8c46-286f112a1130/jobs/27991). But we are still getting killed.

CJ-Wright:
We need tests for the lazyjson class

viniciusdc:
> We need tests for the lazyjson class

I made a test with `update_nodes_with_new_versions(gx)` and it is correctly updating the info, also there is no need to remove the not used files from `versions`, cause there will not be any conflict (the only thing that will happen is that `make_graph` will read some files which information is already been deployed before).

CJ-Wright:
Can you please put that test up in a PR so it can run on CI (and prevent future breakage)?

viniciusdc:
> Can you please put that test up in a PR so it can run on CI (and prevent future breakage)?

Well I've just checked manually (by looking into the file). I loaded the `gx=load_graph()` created a `version/auditwheel.json`  and load the code to check if the information was updated. BTW we do have an `test_update_upstream_versions` file at `tests` I also updated it to work with the new `get_last_version` function.

CJ-Wright:
Can you also put up a test for lazy json

viniciusdc:
> Can you also put up a test for lazy json

Yes I can, to test the `.update` method right ? Can I put it at the `test_update upstream version` or it should be separated ?

CJ-Wright:
It should be in the tests of lazy json, it has it's own test funciton.

viniciusdc:
> It should be in the tests of lazy json, it has it's own test funciton.

Actually, we have a [test](https://github.com/regro/cf-scripts/blob/37b9502668c80fcd8365dcb6d7ebf5552559a28e/tests/test_utils.py#L8) for the `.update`:
```python

lj.update({"hi": "globe"})
    with open(f, "r") as ff:
        assert ff.read() == dumps({"hi": "globe"})
```

CJ-Wright:
awesome! Sorry that I forgot about that

viniciusdc:
> awesome! Sorry that I forgot about that

No problem, it's good that I also can see all the features that we are currently testing.

PR #1063, opened 2020-07-11
Allow automerge for version or migrations only

CJ-Wright:
Note that this will need a PR in webservices to make the most of this and some advertising.

The hope here is that ABI migration PRs are more likely to break if there are any changes that need to be applied to the feedstock. Enabling migration automerge will greatly speed up uptake of pins, similar to the R 4.0 migration.

@beckermr @isuruf @scopatz

### DIFF
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -261,7 +261,8 @@ class MigrationYaml(GraphMigrator):
         if (
             feedstock_ctx.attrs.get("conda-forge.yml", {})
             .get("bot", {})
-            .get("automerge", False) in {'migration', True}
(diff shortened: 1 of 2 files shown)


beckermr:
IDK what the webservices has to do with this. Do you mean a PR in the automerge action code?

beckermr:
I actually think it needs a PR here to make sure the bot only sets automerge for the right kind of PR.

CJ-Wright:
Don't we have a admin command to add automerge keys to the conda-forge.yml?

beckermr:
Yes. Did you want variations of that command?

CJ-Wright:
I think that would be awesome! Unless we're interested in setting the default to automerge for migrations only

PR #1071, opened 2020-07-20
Attempt to profiling the auto-tick.py 

viniciusdc:
There is still some discussion here to be made, as also I am not sure is this actually much work. So please, I really request some review here. The main idea here is quite simple: 
1. when the cli.xsh runs the code for auto_tick main it will actually initiate the main .from `profile.py`
2. which would in sequence initiate the `cProfile` run command (with the default profile specifications);
3.The code should run in the normal way, but at the end a file containing the profiles status will be create at the `profile` folder (wich still need to be added at the `cli.xsh`, IDK where );
4. In the end, we can use (locally) the `SnakeViz` to visualize the informations.

Obs.I also want to add the `tracemalloc` package to check for memory usage (and leakages) but there is still some drawbacks with its usage [see](https://tech.buzzfeed.com/finding-and-fixing-memory-leaks-in-python-413ce4266e7d).


beckermr:
can you post the actual profile data?

CJ-Wright:
I'm hoping for the outcome of this to be the timings of the various pieces of the auto tick script. With those we should be able to answer:
1. How long does each component run for?
2. How many times is each component called?
3. Where should we try to optimize the code first?

viniciusdc:
> I'm hoping for the outcome of this to be the timings of the various pieces of the auto tick script. With those we should be able to answer:
> 
> 1. How long does each component run for?
> 2. How many times is each component called?
> 3. Where should we try to optimize the code first?

It was quite simple, but I only thought this on the last weekend. I will use the config of circle to create a job on some of my repos to run the bot (only the code of run without the deploy phase). Then use this to test the profiler and get it output.

viniciusdc:
> You will also need to remove the PR creation portion.

yes, this is why I am stuck, cause there are some part of `make_graph` that loads the PR's too. So I need to look careflu (in my perspective) with this

CJ-Wright:
The loading is fine, you just need to make certain it doesn't create PRs. Or produce the numbers off the actual bot runs.

viniciusdc:
> The loading is fine, you just need to make certain it doesn't create PRs. Or produce the numbers off the actual bot runs.

'The loading is fine, you just need to make certain it doesn't create PRs.' Ok, This is possible ? 'Or produce the numbers off the actual bot runs.' Since I will be runnig it at my profile of Circle ? (and there will be no deploy phase ?)

CJ-Wright:
I think the relevant lines are here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/auto_tick.py#L254

viniciusdc:
> I think the relevant lines are here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/auto_tick.py#L254

If this is removed for the test, is there any associated function that will bug ?

viniciusdc:
Let's try this, if it works I plan to create a `@profile` decorator to future use.

viniciusdc:
Also, I have to add the new `profile` folder to `cli.xsh`

conda_forge_tick/auto_tick.py
```diff
@@ -711,6 +715,10 @@ def _compute_time_per_migrator(mctx):
 
 
 def main(args: "CLIArgs") -> None:
+    # start profiler
+    pr = cProfile.Profile()
```
CJ-Wright, line 719: you might consider a different variable name, `pr` is a bit overloaded at the moment.

viniciusdc: sorry for that

viniciusdc: done!

viniciusdc:
Also, there should be a problem with `logging` cause this statement `profile_stats.print_stats()`. So we need to test and see if the profiler terminal output will be reasonable.

CJ-Wright:
So long as it is being faithfully represented in the file I'm not certain we need the printout.

viniciusdc:
BTW, I opted by this approach to overlook the `cprofile.run('function named(args)')`, cause I was not sure if the `args` would be recognized. If it was not the case, the function would stop, so with this new approach as soon the code runs the profile runs within it.

viniciusdc:
> So long as it is being faithfully represented in the file I'm not certain we need the printout.

I put it there to test the `stream` variable (I wanted to see if its human readable)

viniciusdc:
> Looking good! Is there a standard format for the profiler outputs? I'd be useful to have that instead of the human readable stuff. Or at least in addition to that.

Yes, the output from the `cprofile` can be rearranged locally by the specification of the user  (i.e what we need is just the output). But for the purpose of testing if the profiling works I still didn't changed that.

viniciusdc:
@CJ-Wright It should be interesting to test and give it a try ?

CJ-Wright:
Lets give this a shot

Issue #1072, opened 2020-07-21
Missed version update for pdfminer.six and FFmpeg


estshorter:
The autotick-bot missed the following update.

- pdfminer.six: v20200720 (released 1 day ago)
- FFmpeg: v4.2.3, v4.3.0 (I manually created PRs for update)

According to the [log of "run_bot" task in circle-ci](https://app.circleci.com/pipelines/github/regro/circle_worker/14989/workflows/72348713-e3ad-4980-9c90-2ff2c8c88811/jobs/32176), the bot doesn't try to check update for both projects.
I hope the following PR solves this problem.

https://github.com/regro/cf-graph-countyfair/pull/23

PR #1073, opened 2020-07-21
run upstream update if bad is from upstream update

CJ-Wright:
@beckermr @viniciusdc

I think this fixes the issues that a few maintainers have experienced with missing bot updates

closes https://github.com/regro/cf-scripts/issues/1072 (I think)


beckermr:
let's try it!

dhirschfeld:
Looks like this has worked! :rocket:

PR #1074, opened 2020-07-22
check solvability with build in addition to host and run

CJ-Wright:
This should fix the kde stack
@scopatz @beckermr


scopatz:
LGTM!

PR #1083, opened 2020-07-31
PROD catch commit errors

beckermr:
We are currently letting these errors cause the migration to not be marked as done. This is stalling python38 where ppl added the migrational yml by hand.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -199,7 +199,7 @@ def run(
         except CalledProcessError as e:
             logger.info(
                 "could not commit to feedstock - "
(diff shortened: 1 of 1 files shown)


CJ-Wright:
Shouldn't this have been handled by the diff logic?

beckermr:
The diff logic comes after this bit of code.

CJ-Wright:
Shouldn't this at least bump the version number?

beckermr:
Bump version number of what?

CJ-Wright:
Sorry the build number

beckermr:
When we add a python version we don't do a build number bump because we do not need to.

PR #1086, opened 2020-08-04
REF use the ci support files for rendering

beckermr:
OK @regro/auto-tick! Here is the PR I was chatting about that uses the CI support files and more of conda-build to render the metadata.

### DIFF
--- a/conda_forge_tick/mamba_solver.py
+++ b/conda_forge_tick/mamba_solver.py
@@ -296,7 +296,7 @@ def _is_recipe_solvable_on_platform(recipe_dir, cbc_path, platform, arch):
     # here we extract the conda build config in roughly the same way that
     # it would be used in a real build
     config = conda_build.config.get_or_merge_config(
-        None, exclusive_config_file=cbc_path, platform=platform, arch=arch,
+        None, platform=platform, arch=arch, variant_config_files=[cbc_path],
     )
     cbc, _ = conda_build.variants.get_package_combined_spec(recipe_dir, config=config)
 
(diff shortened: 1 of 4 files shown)


viniciusdc:
I saw that you solved the issue of fetching more data than necessary using the `_fetch_static_repo`

beckermr:
The zipfile module is in the python standard library. I'd read up on it there.

conda_forge_tick/mamba_solver.py
```diff
@@ -296,7 +296,7 @@ def _is_recipe_solvable_on_platform(recipe_dir, cbc_path, platform, arch):
     # here we extract the conda build config in roughly the same way that
     # it would be used in a real build
     config = conda_build.config.get_or_merge_config(
-        None, exclusive_config_file=cbc_path, platform=platform, arch=arch,
+        None, platform=platform, arch=arch, variant_config_files=[cbc_path],
```
beckermr, line 299: @wolfv IDK if you have used this code anywhere, but I found this bug in the order the config files are applied. In a real build, the .ci_support files generated by conda-smithy come last. If you have not used this bit of code anywhere else, then no need to worry about this.

beckermr:
@CJ-Wright this is ready for review

CJ-Wright:
@beckermr Have you gotten a chance to run `make_graph` locally with these changes? How was the performance?

beckermr:
I've run a subset of the nodes but not the whole thing. Happy to run it all but it won't be until later.

beckermr:
It's roughly 25% slower.

beckermr:
Ok. This code is definitely a little slower. Given that the graph is currently wrong in many cases, I think it should be merged in any case. 

I have some local changes that need to be added so please don't merge yet.

beckermr:
This PR is ready for review and merge.

PR #1096, opened 2020-08-11
Changes to auto_tick profiler.

viniciusdc:
I've opted for `prof` instead of `profile_profiler`, removed the `print_stats` statements (and correlated). I will try  to use `dump_stats`  to enable further visualization of the information (if needed), and as the last update I've put a script/function name at the profile name to facilitate the inspection  and search for the any necessary files (principally when we have multiple profiler running).

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -922,8 +922,8 @@ def main(args: "CLIArgs") -> None:
     prof.disable()
 
     # output to data
-    os.makedirs("profiler", exist_ok=True)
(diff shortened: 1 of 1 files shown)


beckermr:
Tests need to pass before this can be merged.

viniciusdc:
> Tests need to pass before this can be merged.

Right, but IDK what happened with Travis (by the logs it seems to be an error with the Migrator)

viniciusdc:
This PR will wait until #1111 works and after that we should merge #1110 to enhance this PR with the new `@profiling` decorator.

beckermr:
closing as no longer needed I think

PR #1098, opened 2020-08-11
Make ordering deterministic

CJ-Wright:


CJ-Wright:
I'm going to test on my local

CJ-Wright:
This seems to work on my local and produce the desired effect. I'm hoping this cuts down on the git thrashing we're getting

PR #1099, opened 2020-08-11
only new versions

CJ-Wright:
This updates the version number in the graph from the other bot process if and only if the new version number is greater than the existing version number. This is helpful for version numbers that tend to wander (eg. we don't always get the latest version from the server). This would also help cut down on some of the thrash associated with flaky servers giving us non existent version numbers while still letting us issue PRs against the version numbers that are ahead of master.

@regro/auto-tick


viniciusdc:
LGTM, but I do have a doubt regarding the filter `x.replace("-", ".")` . Is the standard of the versions always separated using - ?

CJ-Wright:
Most packages use `.` but some r packages have `-` in their versions see: https://github.com/conda-forge/r-matrix-feedstock/blob/master/recipe/meta.yaml#L1

viniciusdc:
> Most packages use `.` but some r packages have `-` in their versions see: https://github.com/conda-forge/r-matrix-feedstock/blob/master/recipe/meta.yaml#L1

Uhm... really, but is there any other pattern that can pass that , like
`1.8.7_3` ?

CJ-Wright:
Underscore might actually be handled properly. I took that code from some other place in the bot. It would be good to refactor that though so we have one central place for version munging.

viniciusdc:
> Underscore might actually be handled properly. I took that code from some other place in the bot. It would be good to refactor that though so we have one central place for version munging.

I am working with an update to the versions code #1142, so I could refactor the `make_graph` after that, if you prefer

PR #1100, opened 2020-08-11
This will mean that we can put status icons on the status page

CJ-Wright:


beckermr:
Can you explain this @CJ-Wright?

CJ-Wright:
One of my issues with the status page is that we don't display any icons about if a PR is mergable or not, so it is difficult to prioritize. Adding this info means that the status page can display the mergability of the PR next to the feedstock link or however we render that in JS.

conda_forge_tick/status_report.py
```diff
@@ -166,10 +167,15 @@ def graph_migrator_status(
             out["in-pr"].add(node)
             fc = "#31688e"
             fntc = "white"
+            pr_status = pr_json["PR"]['mergeable_state']
+            if pr_status == 'clean':
+                status_icon = ' ✓'
+            else:
+                status_icon = ' ❎'
```
beckermr, line 174: should we crib github here? Why is the "x" not red?

CJ-Wright: Red and Green aren't colorblind friendly, and I'm not certain how to turn individual elements of text colors in graphviz

beckermr: ahhh right - good call!

PR #1106, opened 2020-08-13
Compare grayskull audits against conda-forge

CJ-Wright:


tests/test_audit.py
```diff
@@ -15,7 +15,7 @@
   version: {{ version }}
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/depfinder-{{ version }}.tar.gz
```
beckermr, line 18: what?

conda_forge_tick/utils.py
```diff
@@ -634,3 +645,225 @@ def extract_requirements(meta_yaml):
...
+    ----------
+    name : str
+        Name of the feedstock
+    sub_graph : MutableMapping
+        The existing metadata if any
+    meta_yaml : Optional[str]
+        The string meta.yaml, overrides the file in the feedstock if provided
+    conda_forge_yaml : Optional[str]
+        The string conda-forge.yaml, overrides the file in the feedstock if provided
+    mark_not_archived
```
beckermr, line 842: This option came from my shims around pulling archived from the admin migrations. We might move a copy of that code here so both things are independent.

conda_forge_tick/audit.py
```diff
@@ -123,3 +124,48 @@ def main(args):
...
+            continue
+        # construct the expected filename
+        expected_filename = f"{node_version}.yml"
+        if expected_filename in grayskull_files:
+            # load the feedstock with the grayskull meta_yaml
+            try:
+                with open(
+                    os.path.join("audits/grayskull", expected_filename), "r",
+                ) as f:
+                    new_attrs = load_feedstock(node, {}, meta_yaml=f.read())
```
beckermr, line 148: Why reload here? This pulls from the internet. Maybe use what is in the serialized copy of the graph?

tests/test_audit.py
CJ-Wright: New grayskull release

conda_forge_tick/audit.py
CJ-Wright: This is loading up the grayskull version of the feedstock

conda_forge_tick/utils.py
CJ-Wright: ok, do you want that in this PR?

beckermr: nope just commenting :)

conda_forge_tick/audit.py
beckermr: Ah. I'd expect this block of code to take of order 20 minutes to run just itself then. It is not threaded like we did for making the graph, right?

CJ-Wright: It should be a lot shorter after the initial burn in, since we don't rerun things we've already done, but I'll put it on a thread just for the principle.

Issue #1085, opened 2020-08-04
limit to X profile files on cf-scritps repo


viniciusdc:
We could add an option to remove some of the old [profile files](https://github.com/regro/cf-graph-countyfair/tree/master/profiler) (once a while).

PR #1110, opened 2020-08-14
NEW profiler decorator

viniciusdc:
This should enable us to inspect more functions with more freedom.

### DIFF
--- a/conda_forge_tick/profiler.py
+++ b/conda_forge_tick/profiler.py
@@ -10,13 +10,10 @@ def profiling():
         @wraps(f)
         def __profiling(*rgs, **kwargs):
             with cProfile.Profile() as prof:
-                prof.enable()
-
+                # Target process
                 out_result = f(*rgs, **kwargs)
-                f(*rgs, **kwargs)
 
-                prof.disable()
-
+                # Profile information
                 # get current time
                 now = datetime.now()
                 current_time = now.strftime("%d-%m-%Y") + "_" + now.strftime("%H_%M_%S")
--- a/conda_forge_tick/profiler.py
+++ b/conda_forge_tick/profiler.py
@@ -9,25 +9,26 @@ def profiling():
     def _profiling(f):
         @wraps(f)
         def __profiling(*rgs, **kwargs):
-            prof = cProfile.Profile()
-            prof.enable()
+            with cProfile.Profile() as prof:
+                prof = cProfile.Profile()
+                prof.enable()
 
-            # out_result = f(*rgs, **kwargs) should we expect an output ?
-            f(*rgs, **kwargs)
+                out_result = f(*rgs, **kwargs)
+                f(*rgs, **kwargs)
 
-            prof.disable()
+                prof.disable()
 
-            # get current time
-            now = datetime.now()
-            current_time = now.strftime("%d-%m-%Y") + "_" + now.strftime("%H_%M_%S")
-            # function name -- aka profiler sub-folder
-            function_name = f.__name__
-            # check dir
-            os.makedirs(f"profiler/{function_name}", exist_ok=True)
-            # save stats into file
-            prof.dump_stats(f"profiler/{function_name}/{current_time}")
+                # get current time
+                now = datetime.now()
+                current_time = now.strftime("%d-%m-%Y") + "_" + now.strftime("%H_%M_%S")
+                # process name -- aka profiler sub-folder
+                process_name = os.path.basename(__file__)
+                # check dir
+                os.makedirs(f"profiler/{process_name}", exist_ok=True)
+                # save stats into file
+                prof.dump_stats(f"profiler/{process_name}/{current_time}")
 
-            return  # out_result
+                return out_result
 
         return __profiling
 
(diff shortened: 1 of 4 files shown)


CJ-Wright:
Can you implement this on the current jobs?

viniciusdc:
> Can you implement this on the current jobs?

Yes, it is possible.

viniciusdc:
I will implement this on the `update_prs` profiler stuff, this way I can test it in a better way

viniciusdc:
There is a problem with using `function_name = f.__name__` to get the process name (for the sub-folders), cause we are actuly using it on the `main` fucntions so... it is gonna be a mess. I will replace it with:
```python
# process name -- aka profiler sub-folder
process_name = os.path.basename(__file__)
# check dir
os.makedirs(f"profiler/{process_name}", exist_ok=True)
```

viniciusdc:
Local test passed

viniciusdc:
Also, after #1111 works, we could rethink this #1085, right ?

CJ-Wright:
yes

conda_forge_tick/profiler.py
```diff
@@ -9,25 +9,26 @@ def profiling():
...
-            # function name -- aka profiler sub-folder
-            function_name = f.__name__
-            # check dir
-            os.makedirs(f"profiler/{function_name}", exist_ok=True)
-            # save stats into file
-            prof.dump_stats(f"profiler/{function_name}/{current_time}")
-
-            return  # out_result
+            with cProfile.Profile() as prof:
+                prof = cProfile.Profile()
```
CJ-Wright, line 13: why is this line here?

viniciusdc: ops

viniciusdc: done with the last commit

conda_forge_tick/profiler.py
```diff
@@ -0,0 +1,34 @@
...
+from datetime import datetime
+
+
+# profiling decorator
+def profiling():
+    def _profiling(f):
+        @wraps(f)
+        def __profiling(*rgs, **kwargs):
+            with cProfile.Profile() as prof:
+                prof.enable()
```
CJ-Wright, line 13: Do you still need the enable and disable?

viniciusdc: not anymore, but I will read again the doc about `cProfile.Profile()` to check if there is any problem reggarding that (like the need of `prof.disable`).

viniciusdc: Yup, I've checked [here](https://bugs.python.org/issue29235) an it seems to be unnecessary (redundant and wrong) the usage of the enable and disable in this case. Thanks for the remind CJ.

viniciusdc: tested it locally, seams to be working fine. I will update the PR

viniciusdc:
As [commented](https://github.com/regro/cf-scripts/pull/1111#issuecomment-674887491) by @CJ-Wright there is no need for two PR's with the same goal, then I will be closing #1111 as this PR now comprehends those changes too (in this case as a test for the new `@profiling()`)

conda_forge_tick/profiler.py
viniciusdc: Done

viniciusdc:
> Can you implement this on the current jobs?

Once we test it with the `update_prs` I can insert the on the other process

viniciusdc:
Once tested and working well, we should go to #1085

CJ-Wright:
Is this decorator being applied to the auto tick runs as well?

viniciusdc:
> Is this decorator being applied to the auto tick runs as well?

not yet

viniciusdc:
I want to confirm that it will work as expected, here it goes well but there is always an unexpected thing that could happen

viniciusdc:
If it works, I will update the autotick with `@profiling()` and work on the limit stuff (there is so much information there, only consuming space)

CJ-Wright:
Do you need this PR merged first?

viniciusdc:
> Do you need this PR merged first?

Yes, if possible. The order should be something like:
1. This PR with the `profile.py` and the `update_prs` (as a test);
2. Update `auto_tick` profiler using the `@profiling()`;
3. Remove old instances of profiled files

CJ-Wright:
It would be good if all the changes related to the profiler decorator could be done in one PR.

viniciusdc:
> It would be good if all the changes related to the profiler decorator could be done in one PR.

For that, I must figure out a way to remove the oldest files from `profiler/*`, do you have any idea ?

CJ-Wright:
I would leave that portion out for the moment, since formally it isn't about the decorator, also that requires manual changes to the database.

viniciusdc:
> I would leave that portion out for the moment, since formally it isn't about the decorator, also that requires manual changes to the database.

yup, Y 'r right. So for the profiling that's all here. Unless there is other part of the bot I should add the decorator.

CJ-Wright:
You could put it on make graph, that might be interesting.

PR #1111, opened 2020-08-14
Decorator test for profiler (usage example)

viniciusdc:
This is an attempt to use the new `@profiling` decorator (see #1110 ).

### DIFF
--- a/conda_forge_tick/update_prs.py
+++ b/conda_forge_tick/update_prs.py
@@ -182,25 +182,25 @@ def profiling():
     def _profiling(f):
         @wraps(f)
         def __profiling(*rgs, **kwargs):
-            prof = cProfile.Profile()
-            prof.enable()
+            with cProfile.Profile() as prof:
+                prof = cProfile.Profile()
+                prof.enable()
 
-            # out_result = f(*rgs, **kwargs) should we expect an output ?
-            f(*rgs, **kwargs)
+                out_result = f(*rgs, **kwargs)
 
-            prof.disable()
+                prof.disable()
 
-            # get current time
-            now = datetime.now()
-            current_time = now.strftime("%d-%m-%Y") + "_" + now.strftime("%H_%M_%S")
-            # process name -- aka profiler sub-folder
-            process_name = os.path.basename(__file__)
-            # check dir
-            os.makedirs(f"profiler/{process_name}", exist_ok=True)
-            # save stats into file
-            prof.dump_stats(f"profiler/{process_name}/{current_time}")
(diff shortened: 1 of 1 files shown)


conda_forge_tick/update_prs.py
```diff
@@ -176,14 +177,38 @@ def close_dirty_prs(gx: nx.DiGraph, dry_run: bool = False) -> nx.DiGraph:
...
-    current_time = now.strftime("%d-%m-%Y") + "_" + now.strftime("%H_%M_%S")
+# profiling decorator
+def profiling():
+    def _profiling(f):
+        @wraps(f)
+        def __profiling(*rgs, **kwargs):
+            prof = cProfile.Profile()
+            prof.enable()
+
+            # out_result = f(*rgs, **kwargs) should we expect an output ?
```
CJ-Wright, line 188: I would allow for that possibility

conda_forge_tick/update_prs.py
```diff
@@ -176,14 +177,38 @@ def close_dirty_prs(gx: nx.DiGraph, dry_run: bool = False) -> nx.DiGraph:
...
-def main(args: "CLIArgs") -> None:
-    # get current time
-    now = datetime.now()
-    current_time = now.strftime("%d-%m-%Y") + "_" + now.strftime("%H_%M_%S")
+# profiling decorator
+def profiling():
+    def _profiling(f):
+        @wraps(f)
+        def __profiling(*rgs, **kwargs):
+            prof = cProfile.Profile()
```
CJ-Wright, line 185: You might consider structuring this as a context manager, that way we make certain to clean up even if the underlying function fails

viniciusdc: I will do that, thanks for the comment

viniciusdc: just reading more about, I don't have much experience with

viniciusdc: `cProfile` have a supported Profile class that can be used as an Contex Manager, like this:
```python
with cProfile.Profile() as pr:
    # ... do something ...
```

conda_forge_tick/update_prs.py
viniciusdc: done

CJ-Wright:
Should #1111 and #1110 be merged into the same PR?

viniciusdc:
> Should #1111 and #1110 be merged into the same PR?

I didn't thought about that when I created, but you are right. I will set the changes from this one^ to that #1110

PR #1114, opened 2020-08-14
Add new class of action that the bot can use to update mappings

mariusvniekerk:
Starting with pypi -> conda-forge name

See https://github.com/regro/circle_worker/pull/166

### DIFF
--- a/conda_forge_tick/pypi_name_mapping.py
+++ b/conda_forge_tick/pypi_name_mapping.py
@@ -1,0 +1,110 @@
+"""
+Builds and maintains mapping of pypi-names to conda-forge names
+
+1: Pacakges should be build from a `https://pypi.io/packages/` source
+2: Packages MUST have a test: imports section importing it
+"""
+
+import glob
+import json
+import sys
+import yaml
+import pathlib
+
+from typing import Dict, List
+
+
+def load_node_meta_yaml(filename: str) -> List[Dict[str, str]]:
+    node_attr = json.load(open(filename))
+    meta_yaml = node_attr.get("meta_yaml")
+    return meta_yaml
+
+
+def extract_pypi_information(cf_graph: str) -> List[Dict[str, str]]:
+    package_mappings = []
+    for f in list(glob.glob(f"{cf_graph}/node_attrs/*.json")):
+        meta_yaml = load_node_meta_yaml(f)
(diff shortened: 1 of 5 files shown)


conda_forge_tick/pypi_name_mapping.py
```diff
@@ -0,0 +1,110 @@
...
+import json
+import sys
+import yaml
+import pathlib
+
+from typing import Dict, List
+
+
+def load_node_meta_yaml(filename: str) -> List[Dict[str, str]]:
+    node_attr = json.load(open(filename))
```
CJ-Wright, line 18: you need to use the in-house load here, we have special types (namely set)

conda_forge_tick/pypi_name_mapping.py
```diff
@@ -0,0 +1,110 @@
...
+def extract_pypi_information(cf_graph: str) -> List[Dict[str, str]]:
+    package_mappings = []
+    for f in list(glob.glob(f"{cf_graph}/node_attrs/*.json")):
+        meta_yaml = load_node_meta_yaml(f)
+        if not meta_yaml:
+            continue
+        if "source" in meta_yaml:
+            if "url" in meta_yaml["source"]:
+                src_urls = meta_yaml["source"]["url"]
+                if isinstance(src_urls, str):
```
CJ-Wright, line 32: you could use `as_iterable`

conda_forge_tick/pypi_name_mapping.py
mariusvniekerk: we sorta do, but they're not used for these :D

CJ-Wright: Perhaps, but if we change the underlying json bindings it might be safer to use the in house loaders.

conda_forge_tick/pypi_name_mapping.py
```diff
@@ -0,0 +1,113 @@
+"""
+Builds and maintains mapping of pypi-names to conda-forge names
+
+1: Packages should be build from a `https://pypi.io/packages/` source
+2: Packages MUST have a test: imports section importing it
```
CJ-Wright, line 5: Out of curiosity how many feedstocks fail on this condition?

PR #1122, opened 2020-08-18
Use set outputs to fix issues where more than one recipe outputs the same package

CJ-Wright:
Put up here as a draft/catalog of my thoughts

@beckermr

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -47,7 +47,7 @@ from .utils import (
     LazyJson,
     CB_CONFIG,
     parse_meta_yaml,
-    eval_cmd,
+    eval_cmd, get_deps_from_outputs_lut,
 )
 from .xonsh_utils import env
 from typing import (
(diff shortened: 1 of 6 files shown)


beckermr:
So I've been thinking carefully about how we should be doing things wrt feedstocks and packages. I think at a practical level, this PR has the right approach implementation wise. The original algorithm I had in mind for turning a package graph into a feedstock graph reduces down to the approach here. 

There are a few things I'd like to change though.

1. Not every feedstock outputs its name as a package and we should not assume that.
2. Per 1, we should store an explicit list of the actual outputs of each feedstock computed from the improved rendering in my other PR.
3. We should then use this mapping as you have done to build up the connections between the feedstocks.
4. Everywhere in the code we should always be putting the packages in the requirements sections through this mapping to get feedstocks. In other words, we should never assume the graph is of packages.

CJ-Wright:
Yeah, I think we need to build the data model for the graph before taking this on, if we could explicitly state that something is a feedstock or a package I think we'd gain a lot of ground

PR #1123, opened 2020-08-18
Revert "NEW profiler decorator"

CJ-Wright:
Reverts regro/cf-scripts#1110


CJ-Wright:
This seems to be causing errors on most of the jobs so I'm going to back this out for the moment

viniciusdc:
> This seems to be causing errors on most of the jobs so I'm going to back this out for the moment

No problem. At least I know that there is some kind of error using built in `__enter__` method for `cProfile.Profile` when using at `make_graph.py`

viniciusdc:
I will look for some alterantive way to save it, if not possible I saw some other ways to use the contex manager for the profiler

viniciusdc:
For an update, I am trying to use `@contextmanager` from `contextlib` together with the `@profiling()` decorator we had before, something like this
```python
import cProfile
from contextlib import contextmanager
import os

from functools import wraps
from datetime import datetime


@contextmanager
def ___profiled(dump_filename: str):
     prof = cProfile.Profile()
    prof.enable()
    yield
    prof.disable()
    # dump data
    prof.dump_stats(dump_filename)


# profiling decorator
def profiling():
    def _profiling(f):
        @wraps(f)
        def __profiling(*rgs, **kwargs):
            with ___profiled(profile_filename):
                # Target process
                out_result = f(*rgs, **kwargs)
            return out_result

        return __profiling

    return _profiling
```
It's working, but as the context manager is actualy in another script the profile information currently don't show the information for the target function .

PR #1131, opened 2020-08-20
Profiler decorator

viniciusdc:
Implemented using a context manager custom class to enable and disable the profiler. I've tested it locally with a test script not related to the bot, and it worked (checked with python 3.7)


viniciusdc:
isn't ` ['git', 'add', 'profiler/*']` sufficient to deploy every file and folder within `profiler/` ?

CJ-Wright:
I think so

PR #1141, opened 2020-08-31
Improve name detection for packages like zope.

mariusvniekerk:
We now rely on a commonprefix based method.  

This successfully determines the name for all of the zope, sphinxcontrib, etc packages


mariusvniekerk:
https://gist.github.com/mariusvniekerk/54fc531873f84fab82057025b18fd28c

PR #1153, opened 2020-09-09
Import name priority mapping

mariusvniekerk:
This writes the ordered list of conda-packages determined for an import name to some file


CJ-Wright:
Travis seems unhappy

Issue #1142, opened 2020-09-01
Multi release versions update


viniciusdc:
Some while ago, we started getting some drawbacks with feedstocks that updates to oftem during the week requesting too much of the process. So to surpass the problem, the idea is to insert a regex key to check the update availity.

For that, there is two things that could be done:
-  first split the nodes filtering to `update_version` alone;
- Then we should perform `re.match(version_pattern, new_version_string)`  if either of these matches, then the version bump is performed.

viniciusdc:
So the idea for the first item, would be removing the filters of the `update_upstream_version_sequential` and `update_upstream_version_pool`

```python
_all_nodes = [t for t in gx.nodes.items()]
    random.shuffle(_all_nodes)

    # Inspection the graph object and node update:
    # print(f"Number of nodes: {len(gx.nodes)}")
    node_count = 0
    to_update = []
    for node, node_attrs in _all_nodes:
        attrs = node_attrs["payload"]
        if "Upstream" not in attrs.get("bad", []) or attrs.get("archived"):
            continue
        to_update.append((node, attrs))
```
and insert these statements into the `update_versions` alone, leaving just the computation for the two cited functions.

viniciusdc:
The second item, needs that to new keys in the `conda_forge_config.yml` under `bot`:
 - one like `version_update_regex`;
- and another `version_update_string`
the regex one will be a python regular expression, the update stright will look like x.x when it should come from `get_update_versions`. If either of these match, then the version bump is performed if they don't match, then it won't.

The regular expression pattern, as @scopatz suggested would be something like `\d+\.\d+\.[02468]`.

beckermr:
Let's start with an option to release only every N versions.

viniciusdc:
> Let's start with an option to release only every N versions.

OK, I was wondering about this last week. So, using the regex to verify the number of versions we have released like `new_version` - `current_versions` >= N? or something like a counter for each week release?

beckermr:
probably a counter? or we could just do odd versus even versions?

viniciusdc:
> probably a counter? or we could just do odd versus even versions?

The latter in my opinion is the best, but I can't find a useful way to do that

beckermr:
sure yeah - lots of version parsers are around, like those used in conda, that we can use to extract the parts.

viniciusdc:
I will link a PR (within a hour) with an idea for that then

viniciusdc:
pin #1159

PR #1159, opened 2020-09-15
Verify version order

viniciusdc:
inspects the `new_version` from `get_latest_version` and the current `version` strings, performing a regular expression match between them.

The goal here is to perform a string parse/ordering action inside the `new_version` string (when necessary e.g. the multiple releases of Vim) as a matter of solution to decrease the opened PR's of each new update. Also there is an open issue 1152 and this could be a solution.

### DIFF
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -28,7 +28,7 @@ def verify(version: str, next_version: str):
     """This function takes two versions strings and compare them.
     If either of these match, then the version bump is performed if they don't match, then it won't."""
 
-    pattern = r'\d+\.\d+\.[02468]'
+    pattern = r"\d+\.\d+\.[02468]"
     split_version = re.match(pattern, next_version)
     if split_version.group() > version:
         return True
@@ -96,8 +96,8 @@ def _update_upstream_versions_sequential(
             # check for latest version
             new_version = get_latest_version(node, attrs, sources)["new_version"]
             # if new_version is inferior or not totally released we set new_version value as version
-            if not verify(attrs.get('version'), new_version):
-                new_version = attrs.get('version')
+            if not verify(attrs.get("version"), new_version):
+                new_version = attrs.get("version")
             version_data.update(new_version)
         except Exception as e:
             try:
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -4,6 +4,7 @@ import random
 import json
 import time
 import os
+import re
 import tqdm
 from concurrent.futures import as_completed
 
@@ -21,6 +22,17 @@ from typing import Any, Iterable
 
 # conda_forge_tick :: cft
 logger = logging.getLogger("conda-forge-tick._update_versions")
+
+
+def verify(version: str, next_version: str):
+    """This function takes two versions strings and compare them.
+    If either of these match, then the version bump is performed if they don't match, then it won't."""
+
+    pattern = r'\d+\.\d+\.[02468]'
+    split_version = re.match(pattern, next_version)
+    if split_version.group() > version:
+        return True
+    return False
 
 
 def get_latest_version(
@@ -82,7 +94,11 @@ def _update_upstream_versions_sequential(
         # New version request
         try:
             # check for latest version
-            version_data.update(get_latest_version(node, attrs, sources))
+            new_version = get_latest_version(node, attrs, sources)["new_version"]
+            # if new_version is inferior or not totally released we set new_version value as version
+            if not verify(attrs.get('version'), new_version):
+                new_version = attrs.get('version')
+            version_data.update(new_version)
         except Exception as e:
             try:
                 se = repr(e)
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -29,10 +29,10 @@ def version_match(version, new_version: str, payload):
         If either of these match, then the version bump is performed if they don't match, then it won't."""
     # Example pattern \d+\.\d+\.[02468] TODO need to add an standard version parser when there isn't any given pattern
 
-    yml = payload.__getitem__('conda-forge.yml')
-    with yml['bot'] as bot:
-        if bot['version_pattern']:
-            split_version = re.match(bot.get('version_pattern'), new_version)
+    yml = payload.__getitem__("conda-forge.yml")
+    with yml["bot"] as bot:
+        if bot["version_pattern"]:
+            split_version = re.match(bot.get("version_pattern"), new_version)
             if split_version.group() > version:
                 return True
         else:
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -24,14 +24,20 @@ from typing import Any, Iterable
 logger = logging.getLogger("conda-forge-tick._update_versions")
 
 
-def verify(version: str, next_version: str):
+def version_match(version, new_version: str, payload):
     """This function takes two versions strings and compare them.
-    If either of these match, then the version bump is performed if they don't match, then it won't."""
+        If either of these match, then the version bump is performed if they don't match, then it won't."""
+    # Example pattern \d+\.\d+\.[02468] TODO need to add an standard version parser when there isn't any given pattern
 
-    pattern = r"\d+\.\d+\.[02468]"
-    split_version = re.match(pattern, next_version)
-    if split_version.group() > version:
-        return True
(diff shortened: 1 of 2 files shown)


conda_forge_tick/update_upstream_versions.py
```diff
@@ -23,6 +24,17 @@
 logger = logging.getLogger("conda-forge-tick._update_versions")
 
 
+def verify(version: str, next_version: str):
```
beckermr, line 27: This function needs some unit testing.

viniciusdc: inside it? I was planning to create a test script for the profiler and use the opportunity to create one for this one too.

beckermr: no i mean in the tests dir - it is supposed to order versions - we should make sure it does so - you might also consider using the Version class from conda itself

viniciusdc: I will have a look at the Version class, and think os a test for this too (maybe using the test information of `update_upstream_versions`)

viniciusdc: I saw one of the discussions about the Verison ordering class of Conda, and I learned a lot with that hshshs, I will try to use theirs. As a test code, I will use the current `test_update_upstream_versions` with some modifications...

viniciusdc:
@scopatz What do you think ?

conda_forge_tick/update_upstream_versions.py
```diff
@@ -24,14 +24,20 @@
...
-    If either of these match, then the version bump is performed if they don't match, then it won't."""
-
-    pattern = r"\d+\.\d+\.[02468]"
-    split_version = re.match(pattern, next_version)
-    if split_version.group() > version:
-        return True
+        If either of these match, then the version bump is performed if they don't match, then it won't."""
+    # Example pattern \d+\.\d+\.[02468] TODO need to add an standard version parser when there isn't any given pattern
+
+    yml = payload.__getitem__('conda-forge.yml')
```
CJ-Wright: Why are you using a dunder method directly?

conda_forge_tick/update_upstream_versions.py
```diff
@@ -24,14 +24,20 @@
...
+    with yml['bot'] as bot:
```
CJ-Wright: I doubt that this works, since the dict inside the payload is just a dict not lazyjson

CJ-Wright (changes requested):

conda_forge_tick/update_upstream_versions.py
viniciusdc: Yeah I've checked it now, sorry for the mistake.

viniciusdc: I'm also going to insert a 'N version' checker too, what do you think ? (i.e Bump the version every N versions)

CJ-Wright:
I'm confused what this is doing. We check if the version is a development version and exclude it here https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/update_sources.py#L129 we then get the latest version we could find here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/update_sources.py#L135

conda_forge_tick/update_upstream_versions.py
```diff
@@ -23,6 +24,23 @@
 logger = logging.getLogger("conda-forge-tick._update_versions")
 
 
+def version_match(version, new_version: str, payload):
```
CJ-Wright, line 27: This only seems to need the version pattern from the bot, so I would pass that in here.

CJ-Wright:
It would also be good to add some tests

viniciusdc:
> I'm confused what this is doing. We check if the version is a development version and exclude it here https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/update_sources.py#L129 we then get the latest version we could find here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/update_sources.py#L135

Hmm interesting by some reason I thought I have read it before (but can't remember if this code is new or not -- prob. the latter)

viniciusdc:
> It would also be good to add some tests

Yup, I am working on that... also the reason for this should be this #1142

conda_forge_tick/update_upstream_versions.py
viniciusdc: Yup you are right, the reason I oppened the manager here was to perform a count on the number of version performed (like `version_1 - version_2 >= N` or something)

CJ-Wright, line 27: You might choose a more descriptive name. Also why does this need the existing version?

viniciusdc: to check if the current version (after the match) corresponds to something greater the the current one

viniciusdc: Isn't `version_match` good ?

CJ-Wright: I'm not certain if that is needed, since we already have the maximal version, and we check the new version against the current version in the migrator itself.

To some extent I'd rather have this logic on the migrator itself in the filter method. That way we always have the latest new version in case third parties want it.

CJ-Wright: I would have something more verbose like `check_version_matches_cadence`.
With `version_match` what are you matching the version to?
With `check_pattern` what are you checking the pattern against, which pattern?

viniciusdc: Looking througly at the code you sent before (the `get_version` function) I think the it's easier just to add the `re.match(patter, version)` there. -- If almost the same operation occuring there (this is just an extension)

CJ-Wright: Would it be possible to have it in the version migrator code?

viniciusdc: For what I read until now (line 400~) It can be inserted ther (inside the `VersionOrder` of filter)

viniciusdc: There is a possibility to insert after the 
```python
# if no new version do nothing
if "new_version" not in attrs or not attrs["new_version"]:
    return True
```

CJ-Wright: I would put it lower, right before the final return, so that we check that we can update the version and that the version is later than the current version.

viniciusdc: > I would have something more verbose like `check_version_matches_cadence`.
> With `version_match` what are you matching the version to?
> With `check_pattern` what are you checking the pattern against, which pattern?

Thanks, this realy helps me a lot (not only with this but further names choices too)

viniciusdc: > I'm not certain if that is needed, since we already have the maximal version, and we check the new version against the current version in the migrator itself.
> 
> To some extent I'd rather have this logic on the migrator itself in the filter method. That way we always have the latest new version in case third parties want it.

This is an excellent point (storing the actual versions strings), I will do that (can I set this PR as closed then ?)

viniciusdc: > I would put it lower, right before the final return, so that we check that we can update the version and that the version is later than the current version.

Uhmmm, I do agree if this (it's the best logical way) but I can't understand the ` version_filter` object (more especificaly how it's been used)

CJ-Wright: `version_filter` is a boolean that states if the feedstock should be filtered out. It has two main pieces (potentially 3 if you add your code to it) the first [clause](https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/version.py#L411) checks if the current version is greater than the new version we have. The second [clause](https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/version.py#L415) determines if we've issued a PR for a later version. If either of these clauses are true we don't issue a PR.

viniciusdc: BTW, why `or not attrs.get("new_version"),  # if no new version` is performed as an option? Don't we exclude it when 
```python
if "new_version" not in attrs or not attrs["new_version"]:
            return True
```

viniciusdc: > `version_filter` is a boolean that states if the feedstock should be filtered out. It has two main pieces (potentially 3 if you add your code to it) the first [clause](https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/version.py#L411) checks if the current version is greater than the new version we have. The second [clause](https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/version.py#L415) determines if we've issued a PR for a later version. If either of these clauses are true we don't issue a PR.

It was very explicative, thank you very much 😄

PR #1169, opened 2020-09-22
Make lazyjson lazier to deal with memory usage.

mariusvniekerk:


conda_forge_tick/utils.py
```diff
@@ -130,63 +131,66 @@ def __getitem__(self, name: Any) -> str:
...
 
     def __setitem__(self, key: Any, value: Any) -> None:
-        self._load()
-        assert self.data is not None
         self.data[key] = value
         self._dump()
 
     def __getstate__(self) -> dict:
         state = self.__dict__.copy()
-        state["data"] = None
```
beckermr, line 189: This is for pickling IIRC. Do we need to clear the class level cache in the state?

mariusvniekerk: the class state isn't the instance state, so should be fine

beckermr: great

CJ-Wright:
@mariusvniekerk the tests are unhappy

PR #1171, opened 2020-09-23
Version migrator update with new filter extension

viniciusdc:
The goal here is to perform a regular expression match between the new_version from get_latest_version and the current version strings using (when available) a given pattern. As an attempt/solution to decrease the opened PR's of each new update.

### DIFF
--- a/conda_forge_tick/migrators/version.py
+++ b/conda_forge_tick/migrators/version.py
@@ -432,9 +432,11 @@ class Version(Migrator):
                     for h in attrs.get("PRed", set())
                 )
                 # ...
-                or check_version_matches_cadence(str(attrs.get("version")),
-                                                 str(attrs["new_version"]),
-                                                 attrs["conda-forge.yml"].get("bot"))
+                or check_version_matches_cadence(
+                    str(attrs.get("version")),
+                    str(attrs["new_version"]),
+                    attrs["conda-forge.yml"].get("bot"),
+                )
             )
         except conda.exceptions.InvalidVersionSpec as e:
             warnings.warn(
--- a/conda_forge_tick/migrators/version.py
+++ b/conda_forge_tick/migrators/version.py
@@ -361,6 +361,19 @@ def _fmt_error_message(errors, version):
     return msg
 
 
+def check_version_matches_cadence(new_version: str, version: str, bot):
+    """This function takes two versions strings and compare them by using a given pattern."""
+    if "version_pattern" not in bot or not bot["version_pattern"]:
+        return False
+    pattern = bot.get("version_pattern")
+    if pattern is not None:
+        split_version = re.match(pattern, new_version)
+        # if version is greater then the new_version (~parsed) then we should not perform the PR
+        if version >= split_version.group():
+            return True
+    return False
+
+
 class Version(Migrator):
     """Migrator for version bumping of packages"""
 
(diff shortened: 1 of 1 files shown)


conda_forge_tick/migrators/version.py
```diff
@@ -361,6 +361,19 @@ def _fmt_error_message(errors, version):
...
 
+def check_version_matches_cadence(new_version: str, version: str, bot):
+    """This function takes two versions strings and compare them by using a given pattern."""
+    if "version_pattern" not in bot or not bot["version_pattern"]:
+        return False
+    pattern = bot.get("version_pattern")
+    if pattern is not None:
+        split_version = re.match(pattern, new_version)
+        # if version is greater then the new_version (~parsed) then we should not perform the PR
+        if version >= split_version.group():
```
CJ-Wright, line 372: I think this is handled elsewhere

conda_forge_tick/migrators/version.py
```diff
@@ -418,6 +431,12 @@ def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
                     >= VersionOrder(str(attrs["new_version"]))
                     for h in attrs.get("PRed", set())
                 )
+                # ...
+                or check_version_matches_cadence(
```
CJ-Wright, line 435: I think you need a `not` here, my understanding is that we want to issue a PR if the pattern matches.

viniciusdc: do we need I True statment for the `filter_version` or a False for it to issue a PR (I thought that to perform the PR that should return a True value)

conda_forge_tick/migrators/version.py
viniciusdc: yeah... but not using the pattern (We can set just the `re.match(pattern, new_version).group() >= version` inside the `version_filter` directly if you want -- but there is the problem with getting the key then)

viniciusdc: Here if `version >= the_new_one` then (it means that the current version is the last stable) we don't need to merge it.

conda_forge_tick/migrators/version.py
CJ-Wright: False to issue PR, maybe I have it backwards

viniciusdc: I need to insert a `else` after the comparasion cause if `new_version >= version` it have to return a False (currently there is no response for that situation, and we only return false when `pattern == None`)

conda_forge_tick/migrators/version.py
viniciusdc: ahhh now I get what you've said, sorry... yeah there is no need for the comparasion here just an `if re.match(pattern, new_version).group() == new_version` should work well right ?

conda_forge_tick/migrators/version.py
```diff
@@ -361,6 +361,22 @@ def _fmt_error_message(errors, version):
     return msg
 
 
+def check_version_matches_cadence(new_version: str, version: str, bot):
```
beckermr, line 364: We should write some tests for this function. We put in an example regex someone might use with various version strings and make sure it does what we think it should do.

beckermr: These tests could then be turned into documentation examples for the bot.

viniciusdc: > We should write some tests for this function. We put in an example regex someone might use with various version strings and make sure it does what we think it should do.

Right, I will think of some tests for it.

viniciusdc: > These tests could then be turned into documentation examples for the bot.

As an example for the regex patterns that can be added to the feedstocks configs ?

beckermr: Yup.

beckermr:
closing as we merged versions for a random fraction of releases

PR #1178, opened 2020-09-30
extract canonical imports

CJ-Wright:


CJ-Wright:
(There might not be such a thing as canonical imports)

PR #1182, opened 2020-10-01
handle discrepency between name of feedstock and package (looking at you proj and proj.4)

CJ-Wright:


CJ-Wright:
@regro/auto-tick I think this is ready to go! (I checked some of the migration graphs and they seemed ok)

beckermr (approved):

PR #1183, opened 2020-10-05
ENH enable mirgations of build keys w/ compilers

beckermr:
This PR enables migrations of compilers via two new features in the migration yaml.

1. You can now add an `include_by_build_only` key which cases the bot to only look at build sections when figuring out if it should include a node in the graph. The bot still uses host+run+test for making edges.

2. You can now add an `override_cbc_keys` key which is a list of things that should appear in host/build for a package to be included in the migration. This will allow folks to list things like `fortran_compiler_stub` when migrating against the `fortran_compiler_version` key, for example.

This will also help the CUDA 11 migration if I am not mistaken. 

xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/794
xref: https://github.com/conda-forge/conda-forge.github.io/issues/1160

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -92,6 +92,7 @@ if typing.TYPE_CHECKING:
     from .migrators_types import (
         MetaYamlTypedDict,
         PackageName,
+        MigrationUidTypedDict,
     )
 
 logger = logging.getLogger("conda_forge_tick.auto_tick")
@@ -259,7 +260,8 @@ def run(
     else:
         # push up
         try:
-            # TODO: remove this hack, but for now this is the only way to get the feedstock dir into pr_body
+            # TODO: remove this hack, but for now this is the only way to get
+            # the feedstock dir into pr_body
             feedstock_ctx.feedstock_dir = feedstock_dir
             pr_json = push_repo(
                 session_ctx=migrator.ctx.session,
@@ -449,6 +451,7 @@ def add_rebuild_migration_yaml(
         package_names,
         excluded_feedstocks,
         include_noarch=config.get("include_noarch", False),
+        include_by_build_only=config.get("include_by_build_only", False),
     )
 
     # Note at this point the graph is made of all packages that have a
@@ -456,15 +459,21 @@ def add_rebuild_migration_yaml(
     # Some packages don't have a host section so we use their
     # build section in its place.
 
-    package_names = {
-        p if p in gx.nodes else output_to_feedstock[p] for p in package_names
-    } - excluded_feedstocks
+    if "override_cbc_keys" not in config:
+        # this will fail if we've not used keys in the cbc that are actual
+        # packages in the graph - this we skip it in that case
+        package_names = {
+            p if p in gx.nodes else output_to_feedstock[p] for p in package_names
+        } - excluded_feedstocks
 
-    top_level = {
-        node
-        for node in {gx.successors(package_name) for package_name in package_names}
-        if (node in total_graph) and len(list(total_graph.predecessors(node))) == 0
-    }
+        top_level = {
+            node
+            for node in {gx.successors(package_name) for package_name in package_names}
+            if (node in total_graph) and len(list(total_graph.predecessors(node))) == 0
+        }
+    else:
+        top_level = None
+
     cycles = list(nx.simple_cycles(total_graph))
     migrator = MigrationYaml(
         migration_yaml,
@@ -524,10 +533,14 @@ def migration_factory(
         excluded_feedstocks = set(migrator_config.get("exclude", []))
         pr_limit = min(migrator_config.pop("pr_limit", pr_limit), MAX_PR_LIMIT)
 
-        package_names = (
-            (set(loaded_yaml) | {l.replace("_", "-") for l in loaded_yaml})
-            & all_package_names
-        ) - excluded_feedstocks
+        if "override_cbc_keys" in migrator_config:
+            package_names = set(migrator_config.get("package_names"))
+        else:
+            package_names = (
+                (set(loaded_yaml) | {ly.replace("_", "-") for ly in loaded_yaml})
+                & all_package_names
+            )
+        package_names = package_names - excluded_feedstocks
 
         if not paused:
             add_rebuild_migration_yaml(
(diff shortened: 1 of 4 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -524,10 +533,14 @@ def migration_factory(
         excluded_feedstocks = set(migrator_config.get("exclude", []))
         pr_limit = min(migrator_config.pop("pr_limit", pr_limit), MAX_PR_LIMIT)
 
-        package_names = (
-            (set(loaded_yaml) | {l.replace("_", "-") for l in loaded_yaml})
-            & all_package_names
-        ) - excluded_feedstocks
+        if "override_cbc_keys" in migrator_config:
+            package_names = set(migrator_config.get("package_names"))
```
beckermr, line 537:
```suggestion
            package_names = set(migrator_config.get("override_cbc_keys"))
```

conda_forge_tick/auto_tick.py
```diff
@@ -449,22 +451,29 @@ def add_rebuild_migration_yaml(
...
     # dependency on the pinned package via Host, run, or test.
     # Some packages don't have a host section so we use their
     # build section in its place.
 
-    package_names = {
-        p if p in gx.nodes else output_to_feedstock[p] for p in package_names
-    } - excluded_feedstocks
+    if "override_cbc_keys" not in config:
+        # this will fail if we've not used keys in the cbc that are actual
+        # packages in the graph - this we skip it in that case
```
beckermr, line 464:
```suggestion
        # packages in the graph - thus we skip it in that case
```

beckermr:
Any initial comments on the approach or implementation would be very much appreciated @CJ-Wright!

jakirkham:
Thanks for working on this Matt! 😄

conda_forge_tick/auto_tick.py
```diff
@@ -449,22 +451,29 @@ def add_rebuild_migration_yaml(
...
     # build section in its place.
 
-    package_names = {
-        p if p in gx.nodes else output_to_feedstock[p] for p in package_names
-    } - excluded_feedstocks
+    if "override_cbc_keys" not in config:
+        # this will fail if we've not used keys in the cbc that are actual
+        # packages in the graph - thus we skip it in that case
+        package_names = {
+            p if p in gx.nodes else output_to_feedstock[p] for p in package_names
```
CJ-Wright, line 466:
```suggestion
            output_to_feedstock.get(p, p) for p in package_names
```
This should handle things like proj/proj.4

conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -428,7 +430,10 @@ def create_rebuild_graph(
         requirements = attrs.get("requirements", {})
         host = requirements.get("host", set())
         build = requirements.get("build", set())
-        bh = host or build
+        if include_by_build_only:
```
CJ-Wright, line 433: Is this needed, shouldn't we have exported strong exports into the host and run requirements in addition to build?

beckermr: It is missing from host. See this node: https://github.com/regro/cf-graph-countyfair/blob/master/node_attrs/camb.json and this feedstock: https://github.com/conda-forge/camb-feedstock/blob/master/recipe/meta.yaml

I can see if I can fix that.

conda_forge_tick/auto_tick.py
beckermr: I am also worried that putting a compiler stub through this block of logic will error, but I can test.

CJ-Wright: if the compiler stubs error it will error either way, I don't think they are in the outputs table

conda_forge_tick/auto_tick.py
```diff
@@ -453,13 +453,15 @@ def add_rebuild_migration_yaml(
     # Some packages don't have a host section so we use their
     # build section in its place.
 
-    package_names = {
-        p if p in gx.nodes else output_to_feedstock.get(p, p) for p in package_names
-    } - excluded_feedstocks
+    feedstock_names = {output_to_feedstock.get(p, p) for p in package_names}
+    feedstock_names = {p for p in package_names if p in gx.nodes} - excluded_feedstocks
```
CJ-Wright, line 457: duplicated names?

conda_forge_tick/auto_tick.py
```diff
@@ -456,15 +453,17 @@ def add_rebuild_migration_yaml(
     # Some packages don't have a host section so we use their
     # build section in its place.
 
-    package_names = {
-        p if p in gx.nodes else output_to_feedstock[p] for p in package_names
-    } - excluded_feedstocks
+    feedstock_names = {output_to_feedstock.get(p, p) for p in package_names}
+    feedstock_names = {p for p in package_names if p in gx.nodes} - excluded_feedstocks
```
beckermr, line 457:
```suggestion
    feedstock_names = {p for p in feedstock_names if p in gx.nodes} - excluded_feedstocks
```

jakirkham:
For this migration, we added CUDA 11.0 to the `nvcc` wrapper package manually ( https://github.com/conda-forge/nvcc-feedstock/pull/43 ). I'm wondering if the migration work here would handle migrating the `nvcc` wrapper package automatically for us in the future or if we would still need to do that manually in the future. Thoughts on this?

CJ-Wright:
It doesn't look like it is in the graph @beckermr posted so I think not. I'm not certain why `nvcc` is not in the graph though

jakirkham:
FWIW we did manually migrate `nvcc` before this PR was created. Idk if that would explain it or not.

CJ-Wright:
I think not, since the bot's graphs only care about the bot's PRs. We know not to issue a PR if the changes have already been made, but the bot notes this by making a fake PR for it to keep internally

beckermr:
So nvcc is the compiler for cuda. Thus it does not appear in the graph because when we render the recipes, it is converted to the cuda_compiler_stub. 

This feature is generic for all compilers.

jakirkham:
Yeah that makes sense.

We did some work early on so that `nvcc` could reuse content like `cuda_compiler_version` from `conda-forge-pinning` ( https://github.com/conda-forge/nvcc-feedstock/pull/19 ). So things like adding a migrator do just work ( https://github.com/conda-forge/nvcc-feedstock/pull/47 ).

Though can understand if that is not something we want to do generally.

beckermr:
Right. We'd need support in the bot for linking nvcc to the cuda compiler stub. That is more nontrivial.

jakirkham:
Yep makes sense. I don't think this is critical.

PR #1190, opened 2020-10-08
Check for solvable more frequently

mariusvniekerk:
Since we want to always check for solvability when doing migrations, actually do so.


mariusvniekerk:
An extension of this we can do later is to make the solvability constraints for a migrator a ternary.

1: always
2: if automerge (the current state)
3. nope (what we do for version)

mariusvniekerk:
Short term we need this to avoid making lots of bad prs for py39

PR #1195, opened 2020-10-09
BUG do not use pip as python dep

beckermr:
I think this does the trick. I took a bit of hacking to figure out.

cc @mariusvniekerk


mariusvniekerk:
Neat I would probably just have set

$CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY=False

mariusvniekerk (approved):

beckermr:
> Neat I would probably just have set

> $CONDA_ADD_PIP_AS_PYTHON_DEPENDENCY=False

So I set this in my condarc and mamba ignored it. The bot is going through its internal APIs, not the command line, so I think we have to set the flag in the code.

PR #1199, opened 2020-10-09
ENH expose mamba info

beckermr:
This PR attempts to expose mamba solving info if it is there.

cc @xhochy

### DIFF
--- a/conda_forge_tick/status_report.py
+++ b/conda_forge_tick/status_report.py
@@ -76,6 +76,12 @@ def graph_migrator_status(
 ) -> Tuple[dict, list, nx.DiGraph]:
     """Gets the migrator progress for a given migrator"""
 
+    if hasattr(migrator, "name"):
+        assert isinstance(migrator.name, str)
+        migrator_name = migrator.name.lower().replace(" ", "")
+    else:
+        migrator_name = migrator.__class__.__name__.lower()
+
     num_viz = 0
 
     out: Dict[str, Set[str]] = {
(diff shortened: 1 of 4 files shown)


conda_forge_tick/status_report.py
```diff
@@ -199,6 +205,15 @@ def graph_migrator_status(
             for k in sorted(gx2.successors(node))
             if not gx2[k].get("payload", {}).get("archived", False)
         ]
+        if migrator_name in attrs.get("pre_pr_migrator_status", {}) and node in (
+            out["awaiting-pr"] | out["awaiting-parents"]
```
CJ-Wright, line 209: do you need awaiting parents?

beckermr: likely not - I am also wondering if we should just list solvable as another category. Thoughts?

CJ-Wright: I would put not solvable in an independent category. Although the messaging may need to state that not solvable doesn't mean there is a problem, just that the CDN doesn't have the pacakge (or there is an actual problem).

CJ-Wright: We might consider putting the feedstocks that actually error under `bot-error` so we know to look at them more closely (right now I think they just sit as awaiting parents)

CJ-Wright: We could even set the URL to be the run that they most recently errored on

beckermr: not in this pr :)

beckermr:
Any other comments here @CJ-Wright?

CJ-Wright (approved):
LGTM, thank you!

xhochy:
Thank you, this is really helpful!

PR #1213, opened 2020-10-14
Add pinning propogation to the mamba checks.

mariusvniekerk:
This will ensure that we don't issue a PR when not all python versions
have a valid run environment

### DIFF
--- a/conda_forge_tick/mamba_solver.py
+++ b/conda_forge_tick/mamba_solver.py
@@ -238,6 +238,8 @@ def is_recipe_solvable(feedstock_dir) -> Tuple[bool, List[str], Dict[str, bool]]
         in the CI scripts.
     errors : list of str
         A list of errors from mamba. Empty if recipe is solvable.
+    solvable_by_variant : dict
+        A lookup by variant config that shows if a particular config is solvable
     """
 
     errors = []
@@ -249,7 +251,7 @@ def is_recipe_solvable(feedstock_dir) -> Tuple[bool, List[str], Dict[str, bool]]
             "This attempted migration is being reported as not solvable.",
         )
         logger.warning(errors[-1])
-        return False, errors
+        return False, errors, {}
 
     if not os.path.exists(os.path.join(feedstock_dir, "recipe", "meta.yaml")):
         errors.append(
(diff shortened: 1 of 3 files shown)


conda_forge_tick/mamba_solver.py
```diff
@@ -280,7 +280,7 @@ def is_recipe_solvable(feedstock_dir):
             arch,
         )
         solvable = solvable and _solvable
-        errors.extend(_errors)
+        errors.extend(["f{cbc_fname}: {e}" for e in _errors])
```
beckermr, line 283: I already put a fix in for this. Can you merge master in and figure out which one we should keep. I think yours might be better.

beckermr: Also you need to extract the basename here. This is a full path.

mariusvniekerk: made a hybrid

beckermr (approved):

beckermr:
What?

> conda.exceptions.InvalidVersionSpec: Invalid version '>=4.0': invalid character(s)

beckermr:
Once the tests are fixed we can merge @mariusvniekerk

beckermr:
Does the latest version still handle python properly? Maybe we should add a test?

conda_forge_tick/mamba_solver.py
```diff
@@ -217,7 +218,7 @@ def _mamba_factory(channels, platform):
     return MambaSolver(list(channels), platform)
 
 
-def is_recipe_solvable(feedstock_dir):
+def is_recipe_solvable(feedstock_dir) -> Tuple[bool, List[str], Dict[str, bool]]:
```
beckermr, line 221: you need to update the doc string.

conda_forge_tick/mamba_solver.py
```diff
@@ -282,8 +284,9 @@ def is_recipe_solvable(feedstock_dir):
         solvable = solvable and _solvable
         cbc_name = os.path.basename(cbc_fname).rsplit(".", maxsplit=1)[0]
         errors.extend([f"{cbc_name}: {e}" for e in _errors])
+        solvable_by_cbc[cbc_name] = _solvable
 
-    return solvable, errors
+    return solvable, errors, solvable_by_cbc
```
beckermr, line 289: There are two branches above that return early. They need a third return argument too.

beckermr (changes requested):

PR #1215, opened 2020-10-15
Improve mamba solver pins

mariusvniekerk:
This adds in support for injecting virtual packages into the solve space allowing us to solve for things like `__glibc`, `__cuda` etc.

Additionally this opens up a future prospect to instead of ignoring self outputs , we can just make a fake channel that contains them and have the solve work as normal.

### DIFF
--- a/conda_forge_tick/mamba_solver.py
+++ b/conda_forge_tick/mamba_solver.py
@@ -255,7 +255,7 @@ class MambaSolver:
             repo.set_priority(priority, subpriority)
             self.repos.append(repo)
 
-    def solve(self, specs):
+    def solve(self, specs) -> Tuple[bool, List[str]]:
         """Solve given a set of specs.
 
         Parameters
--- a/conda_forge_tick/mamba_solver.py
+++ b/conda_forge_tick/mamba_solver.py
@@ -314,7 +314,7 @@ def virtual_package_repodata():
     fake_packages = [
         FakePackage("__glibc", "2.12"),
         FakePackage("__glibc", "2.17"),
-        FakePackage("__cuda", "9.2"),        
+        FakePackage("__cuda", "9.2"),
         FakePackage("__cuda", "10.0"),
         FakePackage("__cuda", "10.1"),
         FakePackage("__cuda", "10.2"),
--- a/conda_forge_tick/mamba_solver.py
+++ b/conda_forge_tick/mamba_solver.py
@@ -9,13 +9,18 @@ The basic workflow is for yaml file in .ci_support
 Most of the code here is due to @wolfv in this gist,
 https://gist.github.com/wolfv/cd12bd4a448c77ff02368e97ffdf495a.
 """
+import json
 import os
 import logging
 import glob
 import functools
+import pathlib
 import pprint
 import re
-from typing import Dict, Tuple, List
+import time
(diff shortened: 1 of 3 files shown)


conda_forge_tick/mamba_solver.py
```diff
@@ -373,11 +394,20 @@ def _is_recipe_solvable_on_platform(recipe_dir, cbc_path, platform, arch):
                 errors.append(_err)
 
         def apply_pins(reqs):
-            from conda_build.render import get_pin_from_build
+            return reqs
```
beckermr, line 397: You have an early return here.

conda_forge_tick/mamba_solver.py
```diff
@@ -302,6 +302,30 @@ def _clean_reqs(reqs, names):
     return reqs
 
 
+def filter_problematic_reqs(reqs):
```
beckermr, line 305: This will sure break things that actually depend on this package. There has to be a better way and I skeptical we need this.

beckermr (changes requested):

conda_forge_tick/mamba_solver.py
mariusvniekerk: nothing should depend on this anymore in this day and age, these are generally reserved for compatibility hacks with legacy states

beckermr (changes requested):
The issues we are having are when the repos are rerendered with the migrations. So in the test suite, we should use an explicit CI support file and recipe file that has those rerenderings. This way our tests don't depend on the live state of the repos.

mariusvniekerk:
though some of these cases failed for all variants (like the grpc one).  I think generating fake repodata is the only long term way to solve this

beckermr:
I mean yeah in principle we want fake repodata too, but using explicit recipe outputs in the tests is an easy fix. The cupy test is passing only because it does not include the cuda11 migration file in its smithy render, for example.

mariusvniekerk:
The nvcc test now passes

mariusvniekerk:
@beckermr ready to go?

beckermr:
I am looking at this this morning.

conda_forge_tick/mamba_solver.py
```diff
@@ -218,7 +299,52 @@ def _mamba_factory(channels, platform):
...
+
+    tmp_path = pathlib.Path(tmp_dir)
+    repodata = FakeRepoData(tmp_path)
+    fake_packages = [
+        FakePackage("__glibc", "2.12"),
+        FakePackage("__glibc", "2.17"),
+        FakePackage("__cuda", "10.0"),
+        FakePackage("__cuda", "10.1"),
+        FakePackage("__cuda", "11.0"),
+        FakePackage("__cuda", "11.2"),
```
beckermr, line 320: cuda versions should be 9.2 10.0 10.1 10.2 11.0 11.1

conda_forge_tick/mamba_solver.py
beckermr: i'll fix this in another PR can we remove this code for now?

beckermr: We've chatted offline and are keeping this code for now.

beckermr (approved):

PR #1218, opened 2020-10-17
start work on type things

CJ-Wright:
I'm not certain if this can go in. It certainly doesn't pass all of mypy at this point.
My hope is that it does help to clarify some of the types, especially when something is a package name or not.

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -35,7 +35,7 @@ from xonsh.lib.collections import _convert_to_dict, ChainDB
 
 if typing.TYPE_CHECKING:
     from mypy_extensions import TypedDict, TestTypedDict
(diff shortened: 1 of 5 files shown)


conda_forge_tick/utils.py
```diff
@@ -679,12 +679,12 @@ def _fetch_static_repo(name, dest):
 
 def populate_feedstock_attributes(
     name: str,
-    sub_graph: typing.MutableMapping,
+    sub_graph: AttrsTypedDict_,
```
mariusvniekerk, line 682:
```suggestion
    sub_graph: "AttrsTypedDict_",
```

Either quote this, or import AttrsTypedDict outside of a `typing.TYPE_CHECKING` blockj

PR #1225, opened 2020-10-24
ENH support PRs into branches

### LINKED ISSUE #130: Make PRs for back tags
It may be good to create PRs for updates to older tags (eg PostgresSQL long term support security updates).

beckermr:
We can support this via the `abi_migration_branches` key in the `bot` section of the conda-forge.yml.

closes #130

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -906,10 +906,11 @@ def main(args: "CLIArgs") -> None:
                     base_branches = (
                         attrs.get("conda-forge.yml", {})
                         .get("bot", {})
-                        .get("abi_migration_branches", ["master"])
+                        .get("abi_migration_branches", [])
                     )
                 else:
-                    base_branches = ["master"]
+                    base_branches = []
+                base_branches += ["master"]
 
                 # Don't let CI timeout, break ahead of the timeout so we make certain
                 # to write to the repo
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -253,7 +253,7 @@ def run(
             if pre_key not in feedstock_ctx.attrs:
                 feedstock_ctx.attrs[pre_key] = {}
             feedstock_ctx.attrs[pre_key][migrator_name] = sanitize_string(
-                "not solvable: %s: %s" % (base_branch, sorted(set(errors)))
+                "not solvable: {}: {}".format(base_branch, sorted(set(errors))),
             )
             eval_cmd(f"rm -rf {feedstock_dir}")
             return False, False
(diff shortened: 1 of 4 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -133,9 +134,8 @@ def run(
         Whether to rerender
     fork : bool
         If true create a fork, defaults to true
-    gh : github3.GitHub instance, optional
-        Object for communicating with GitHub, if None build from $USERNAME
-        and $PASSWORD, defaults to None
+    base_branch : str, optional
```
CJ-Wright, line 137: You might mention that this defaults to `master`.

conda_forge_tick/auto_tick.py
```diff
@@ -901,6 +902,16 @@ def main(args: "CLIArgs") -> None:
...
             with mctx.graph.nodes[node_name]["payload"] as attrs:
+                if not isinstance(migrator, Version):
+                    base_branches = (
+                        attrs.get("conda-forge.yml", {})
+                        .get("bot", {})
+                        .get("abi_migration_branches", [])
+                    )
+                else:
+                    base_branches = []
+                base_branches += ["master"]
```
CJ-Wright, line 913: You might consider putting master first, in case the other branches balk

conda_forge_tick/auto_tick.py
```diff
@@ -921,92 +932,105 @@ def main(args: "CLIArgs") -> None:
...
-                    if pre_key not in attrs:
-                        attrs[pre_key] = {}
-                    attrs[pre_key][migrator_name] = sanitize_string(
-                        "bot error: %s" % str(traceback.format_exc()),
-                    )
-                else:
-                    if migrator_uid:
-                        # On successful PR add to our counter
-                        good_prs += 1
+                        try:
```
CJ-Wright, line 945: This seems like a double try block?

beckermr: Right, it is a double try block. I didn't know if we wanted to clean out the tmp dirs after each branch or not. Thoughts?

CJ-Wright: hmm, I don't know. On the one hand it seems safer to clear them out, but I'm not certain what that safety is guarding against.

beckermr: The code is going to refetch the repo from github anyways, so I think maybe we should clean them out so that everything is consistent with the old way of doing things. We can come back later to optimize both the fetch and cleaning if we care. That ok?

CJ-Wright:
Do we have any feedstocks that support this?

beckermr:
I am going to start testing today. Not yet.

PR #1227, opened 2020-10-25
DEP remove explicit pinning stuff since not needed

beckermr:
This should fix the jpeg errors and avoid us having to maintain a list of openssl-like version schemes.


beckermr:
Let's try this. If the openssl errors come back, we can revert or try something else.

PR #1231, opened 2020-10-26
ENH get actual run exports

beckermr:
This is WIP code that gets the actual run exports from the packages. I am avoiding conda-build for doing this to keep things simple. We might include extra due to ignored exports. Also we are not handling channels correctly. We really are at the limit of what the tooling supports. :/

to do
- [x] figure out what to download with channel data
- [ ] make sure to use proper channels given in the `ci_support` file
- [x] use libcfgraph for run exports

cc @wolfv @ocefpaf @CJ-Wright @mariusvniekerk  for viz


conda_forge_tick/mamba_solver.py
```diff
@@ -228,6 +236,45 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                    shell=True,
+                )
+
+            rxpth = f"{tmpdir}/info/run_exports.json"
+
+            if os.path.exists(rxpth):
+                with open(rxpth) as fp:
+                    rxdata = json.load(fp)
+                for key in ["weak", "strong"]:
+                    if key not in rxdata:
```
CJ-Wright, line 266: you might consider `setdefault`

conda_forge_tick/mamba_solver.py
```diff
@@ -297,8 +357,52 @@ def solve(self, specs) -> Tuple[bool, List[str]]:
...
+            for _, _, jdata in to_link:
+                data = json.loads(jdata)
+                solution.append(
+                    " ".join([data["name"], data["version"], data["build"]]),
+                )
+
+            if get_run_exports:
+                run_exports = self._get_run_exports(to_link)
+
+        if get_run_exports:
```
CJ-Wright, line 379: I'm not over the moon about having a variable length return. It may be the only way forward, but it does produce some extra cognitive load

beckermr: Yeah me neither but this is the most efficient thing.

conda_forge_tick/mamba_solver.py
```diff
@@ -297,8 +364,73 @@ def solve(self, specs) -> Tuple[bool, List[str]]:
                 solver.problems_to_str(),
             )
             err = solver.problems_to_str()
+            solution = None
+            run_exports = {"weak": set(), "strong": set()}
```
beckermr, line 368:
```suggestion
            run_exports = copy.deepcopy(DEFAULT_RUN_EXPORTS)
```

beckermr:
So once we finish updating libcfgrqph this pr will be as fast as the code is now. This let's wait a week or so and then merge.

beckermr:
We can punt the channels to another issue.

conda_forge_tick/mamba_solver.py
```diff
@@ -9,14 +9,19 @@
 Most of the code here is due to @wolfv in this gist,
 https://gist.github.com/wolfv/cd12bd4a448c77ff02368e97ffdf495a.
 """
-import json
+import rapidjson as json
```
CJ-Wright, line 12: thought for another PR: should we just shim this globally?

conda_forge_tick/mamba_solver.py
```diff
@@ -228,6 +247,45 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.write()
 
 
+def _get_run_exports(link_tuple):
+    c, pkg, jdata = link_tuple
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        try:
```
CJ-Wright, line 254: would it be possible to make this try except smaller and stuff some of it into an else? If `json.load` fails we might not want to catch that.

conda_forge_tick/mamba_solver.py
```diff
@@ -228,6 +247,45 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            subprocess.run(
+                f"cd {tmpdir} && curl -s -L {c}/{pkg} --output {pkg}",
+                shell=True,
+            )
+
+            # unpack and read if it exists
+            if os.path.exists(f"{tmpdir}/{pkg}"):
+                conda_package_handling.api.extract(f"{tmpdir}/{pkg}")
+
+            if pkg.endswith(".tar.bz2"):
```
CJ-Wright, line 265: Is this an actual file path? If so you might consider using path tools to split the extension off.

conda_forge_tick/mamba_solver.py
```diff
@@ -297,8 +364,114 @@ def solve(self, specs) -> Tuple[bool, List[str]]:
...
+        else:
+            return success, err, solution
+
+    def _get_run_exports(self, link_tuples):
+        """Given tuples of (channel, file, json repodata shard) produce a
+        dict with the weak and strong run exports for the packages.
+        """
 
-        return success, err
+        with ProcessPoolExecutor(max_workers=None) as exe:
```
CJ-Wright, line 392: Is this something we're going to want to control the max_workers on?

conda_forge_tick/mamba_solver.py
```diff
@@ -297,8 +364,114 @@ def solve(self, specs) -> Tuple[bool, List[str]]:
...
+                                .get("build", {})
+                                .get("run_exports", {})
+                            )
+                            if rx:
+                                self.run_exports[link_tuple] = copy.deepcopy(
+                                    DEFAULT_RUN_EXPORTS,
+                                )
+                                for k in rx:
+                                    if k in DEFAULT_RUN_EXPORTS:
+                                        for _k in rx[k]:
```
CJ-Wright, line 432: I think `set.update` lets you do multiple additions at once

conda_forge_tick/mamba_solver.py
```diff
@@ -297,8 +364,114 @@ def solve(self, specs) -> Tuple[bool, List[str]]:
...
+                                            self.run_exports[link_tuple][k].add(_k)
+                                    else:
+                                        self.run_exports[link_tuple]["weak"].add(k)
```
CJ-Wright, line 435: Isn't `weak` in default run exports?

conda_forge_tick/mamba_solver.py
CJ-Wright: Also this seems to not be doing too much heavy lifting in terms of code. Should this be on a thread? Or are we running into "conda isn't threadsafe" issues?

conda_forge_tick/mamba_solver.py
beckermr: I don't have a strong opinion here.

conda_forge_tick/mamba_solver.py
beckermr: This else captures syntax like

```yaml
build:
  run_exports:
    - blah
```

which makes `blah` a weak run export by default.

conda_forge_tick/mamba_solver.py
```diff
@@ -297,8 +364,114 @@ def solve(self, specs) -> Tuple[bool, List[str]]:
...
+                                .get("run_exports", {})
+                            )
+                            if rx:
+                                self.run_exports[link_tuple] = copy.deepcopy(
+                                    DEFAULT_RUN_EXPORTS,
+                                )
+                                for k in rx:
+                                    if k in DEFAULT_RUN_EXPORTS:
+                                        for _k in rx[k]:
+                                            self.run_exports[link_tuple][k].add(_k)
```
beckermr, line 433:
```suggestion
                                        self.run_exports[link_tuple][k].update(rx[k])
```

conda_forge_tick/mamba_solver.py
beckermr: Yeah. This is annoying to code around

```
In [3]: import os

In [4]: os.path.splitext("blah.tar.gz")
Out[4]: ('blah.tar', '.gz')
```

I'm happy to try but the if statement above seems clearer, rather than splitting and it it is not `.conda`, splitting again.

conda_forge_tick/mamba_solver.py
beckermr: We have to decompress the downloaded package which is CPU limited, thus processes seem more appropriate. 

Also, conda on threads scares me too. 

I can add a max_workers constructor argument to the class, but I'm ok letting the system choose.

conda_forge_tick/mamba_solver.py
CJ-Wright: oof, ok

conda_forge_tick/mamba_solver.py
CJ-Wright: we support that? oof

CJ-Wright (approved):
Thank you for answering my questions

conda_forge_tick/mamba_solver.py
beckermr: yup it is: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements

conda-build might always put out an explicit weak section, but I didn't know so I added this just in case.

conda_forge_tick/mamba_solver.py
beckermr: Right. I agree this could be smaller. What error do we want to throw here? Should we throw an error for anything?

beckermr: Actually, thinking through this further, I think we should keep the code as-is. Basically, if we cannot find a run export, I'd rather us issue an erroring PR than not issue a PR at all. I'm going to merge. If we find this is swallowing a lot of errors we'd rather not, we can address it.

PR #1234, opened 2020-11-01
BUG fix string bug in mamba solver

beckermr:
Sometimes libcfgraph has the run export as a single string. This catches that.


tests/test_mamba_solvable.py
```diff
@@ -256,6 +256,18 @@ def test_arrow_solvable(tmp_path):
     assert solvable
 
 
+def test_guiqwt_solvable(tmp_path):
+    """this has been throwing errors so test it"""
```
beckermr, line 260:
```suggestion
    """test for run exports as a single string in pyqt"""
```

PR #1241, opened 2020-11-03
BUG archived nodes are always migrated

beckermr:
I am seeing some archived nodes in the graph and they appear to be messing some stuff up.

### DIFF
--- a/conda_forge_tick/make_graph.py
+++ b/conda_forge_tick/make_graph.py
@@ -25,7 +25,6 @@ from .utils import (
     load_graph,
     dump_graph,
     LazyJson,
-    pluck,
 )
 from .xonsh_utils import env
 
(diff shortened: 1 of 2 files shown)


conda_forge_tick/make_graph.py
```diff
@@ -173,6 +174,11 @@ def make_graph(
 
     logger.info("inferring nodes and edges")
 
+    # pluck anything marked as archived
+    for name in list(gx.nodes):
+        if gx.nodes[name].get("payload", {}).get("archived", False):
+            pluck(gx, name)
```
CJ-Wright, line 180: Are we sure we want to do that here? If it is a true dep we should have it in the graph. For instance if a migration needs to run through an archived feedstock then that is something we'll need to handle, either by moving the dep or by reviving the feedstock.

conda_forge_tick/migrators/core.py
```diff
@@ -471,7 +472,9 @@ def predecessors_not_yet_built(self, attrs: "AttrsTypedDict") -> bool:
             if (
                 m_pred_json
                 and m_pred_json.get("PR", {"state": "open"}).get("state", "") == "open"
+                and not payload.get("archived", False)
```
beckermr, line 475: @CJ-Wright Here is the bug. We need to always say archived nodes in the graph are migrated. We did that in the if statement above, but not in this one.

CJ-Wright: ok, you might refactor this to use a common var.

beckermr: done.

CJ-Wright (approved):

PR #1242, opened 2020-11-03
ENH close migrations after a time

beckermr:
I am starting from @CJ-Wright's branch and then adding the time thing.

### DIFF
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -478,7 +478,7 @@ class GraphMigrator(Migrator):
                 ts = (
                     payload.get("PRed", [])[pr_index]
                     .get("PR", {})
-                    .get("updated_at", None)
+                    .get("created_at", None)
                 )
                 state = payload.get("PR", {"state": "open"}).get("state", "")
                 if ts is not None and state == "open":
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -1,6 +1,8 @@
 """Classes for migrating repos"""
 import os
 import re
+import dateutil.parser
+import datetime
 from itertools import chain
 import typing
 import logging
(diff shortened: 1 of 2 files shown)


conda_forge_tick/migrators/core.py
```diff
@@ -443,6 +445,51 @@ def __init__(
...
+            try:
+                payload = self.graph.nodes[node]["payload"]
+            except KeyError as e:
+                print(node)
+                raise e
+
+            # we don't need to look at archived feedstocks
+            # they are always "migrated"
+            if payload.get("archived", False):
+                continue
```
CJ-Wright, line 460: Is it worthwhile pulling this out as a generator that we walk through, to reduce code copy paste?

conda_forge_tick/migrators/core.py
```diff
@@ -443,6 +445,51 @@ def __init__(
...
+                ts = (
+                    payload.get("PRed", [])[pr_index]
+                    .get("PR", {})
+                    .get("updated_at", None)
+                )
+                state = payload.get("PR", {"state": "open"}).get("state", "")
+                if ts is not None and state == "open":
+                    now = datetime.datetime.now(datetime.timezone.utc)
+                    ts = dateutil.parser.parse(ts)
+                    if now - ts < datetime.timedelta(days=30):
```
CJ-Wright, line 479: Should we get the timedelta from the migration yaml itself? (And default to some default value?)

beckermr: The time delta is for the PRs from the migration, not when the migrator was started. We want to close when all PRs have been issued and every PR is at least 30 days old.

conda_forge_tick/migrators/core.py
beckermr: yup we can do that!

beckermr:
ok @CJ-Wright ready for another look!

conda_forge_tick/migrators/core.py
CJ-Wright: Sorry. I meant more should this be a configurable field in the migration yaml. For instance some migrations may want to be longer/shorter on that time. It isn't a desperately needed feature though so we can punt if you'd like.

beckermr: Let's punt I think. 30 days is a fail-safe really and this only applies to migrations which have issued all PRs.

CJ-Wright (approved):
LGTM thank you!

PR #1254, opened 2020-11-07
PERF make commits faster and use less memory

beckermr:
Trying to get things moving again.


conda_forge_tick/cli.py
```diff
@@ -43,98 +43,120 @@ def deploy(args):
...
+        # modified and staged but not deleted
+        + subprocess.run(
+            "git diff --name-only --cached --diff-filter=d",
+            shell=True,
+            capture_output=True,
+        ).stdout.decode("utf-8").splitlines()
+    )
+    n_added = 0
+    for file in files_to_add:
+        if file is not None and len(file) > 0 and os.path.exists(file):
```
CJ-Wright, line 75: would `bool(file)` suffice to replace the first two clauses

conda_forge_tick/cli.py
```diff
@@ -43,98 +43,120 @@ def deploy(args):
...
-
-    status = 1
-    num_try = 0
-    while status != 0 and num_try < 10 and graph_ok:
+        # make sure the graph can load, if not we will error
         try:
-            if num_try == 9:
+            gx = load_graph()
+            # TODO: be more selective about which json to check
+            for node, attrs in gx.nodes.items():
```
CJ-Wright, line 93: Do you want to use your file list here to check only the nodes that have been updated? (If any)

CJ-Wright (commented):
some light suggestions

conda_forge_tick/cli.py
beckermr: It might but I'd be a lot less clear with a marginal performance gain.

conda_forge_tick/cli.py
beckermr: Not worth the effort.

conda_forge_tick/cli.py
beckermr: I simplified a bit here.

CJ-Wright (approved):

PR #1266, opened 2020-11-11
PERF cache mamba solver solve method

beckermr:
This might speed it up a bit.


beckermr:
Doesn't help much at all. Closing.

PR #1273, opened 2020-11-12
ENH no longer set bad in version updates

beckermr:
Let's try this.

### DIFF
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -20,7 +20,7 @@ from .update_sources import (
 from typing import Any, Iterable
 
 # conda_forge_tick :: cft
-logger = logging.getLogger("conda-forge-tick._update_versions")
+logger = logging.getLogger("conda-forge-tick.update_upstream_versions")
 
 
 def get_latest_version(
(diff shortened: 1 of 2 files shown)


viniciusdc:
cool, LGTM.

CJ-Wright (approved):
LGTM, this may make it more difficult to debug issues for particular sources but I'm not certain that we're doing much of that now and this code would run properly locally if we truly needed to debug things.

viniciusdc (approved):

beckermr:
I went ahead and did the update of the graph nodes with more specific code. That way we can add debugging info to the version json at some later point.

viniciusdc:
Great! Just a question, now we don't have the `set False` inside the `for source in sources` loop... so do we have any case wehre we are currently opting for a `new_version = False`?

conda_forge_tick/update_upstream_versions.py
```diff
@@ -20,18 +20,18 @@
...
 
 
 def get_latest_version(
     name: str,
     meta_yaml: Any,
     sources: Iterable[AbstractSource],
 ) -> dict:
-    version_data = {}
-    # avoid
+    version_data = {"new_version": False}
```
beckermr, line 31: @viniciusdc I set things to False at the top

viniciusdc: Oh, Great!

viniciusdc: isn't it better to set `version_data = {}` from the beginning, then modify the else to elif and then set an else for `False` condition?... (this way we don't have the problem CJ pointed)

beckermr: It doesn't matter. As long as we only override the version in the graph nodes if it is newer or the graph nodes don't have a valid version, then it will work.

beckermr:
If this breaks things I will back it out.

PR #1278, opened 2020-11-14
BUG fix requirement aggregation for Nones

beckermr:
It turns out when we use ChainDB for requirement aggregation, it makes a mess when aggregating `None` with lists. We get a `None` for a requirements section in some cases like this

```
requirements:
  build:
    - blah   # [not win]
```

When the selectors are evaluated, it leaves

```
requirements:
  build:
```

which results in a `None` for `build`. 

This PR fixes this issue by setting `None`s to empty lists if we find them.

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -765,8 +765,9 @@ def populate_feedstock_attributes(
             # aggregated meta_yaml
             if "requirements" in varient_yamls[-1]:
                 for section in ["build", "host", "run"]:
-                    val = varient_yamls[-1]["requirements"].get(section, [])
-                    varient_yamls[-1]["requirements"][section] = val or []
+                    if section in varient_yamls[-1]["requirements"]:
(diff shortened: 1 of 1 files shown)


beckermr:
cc @isuruf This PR fixes `libwebp`.

conda_forge_tick/utils.py
```diff
@@ -761,6 +761,14 @@ def populate_feedstock_attributes(
                 ),
             )
 
+            # sometimes the requirements come out to None and this ruins the
+            # aggregated meta_yaml
+            if "requirements" in varient_yamls[-1]:
+                for section in ["build", "host", "run"]:
+                    if section in varient_yamls[-1]["requirements"]:
```
beckermr, line 768: We make sure to set a section only if it is actually in the recipe. Adding a section when it is not there might confuse migrators trying to move CB2 recipes to CB3.

beckermr:
Going to merge. If it breaks things, I will back it out.

PR #1282, opened 2020-11-14
BUG add branch to muid

beckermr:
This should let the bot issue PRs.

### DIFF
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -347,6 +347,11 @@ class Migrator:
         # Carveout for old migrators w/o obj_versions
         if self.obj_version:
             d["migrator_object_version"] = self.obj_version
+
+        branch = attrs.get("branch", "master")
+        if attrs.get("branch", "master") != "master":
+            d["branch"] = branch
+
         return d
 
     def order(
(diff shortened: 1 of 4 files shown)


beckermr:
If we merge this PR, the bot should be able to issue PRs to branches. However, once the PR on the master branch is merged, the bot will think the whole feedstock is done. This actually seems correct in that we don't want to hold up PRs on stale version branches at some level.

conda_forge_tick/migrators/core.py
```diff
@@ -347,6 +347,11 @@ def migrator_uid(self, attrs: "AttrsTypedDict") -> "MigrationUidTypedDict":
         # Carveout for old migrators w/o obj_versions
         if self.obj_version:
             d["migrator_object_version"] = self.obj_version
+
+        branch = attrs.get("branch", "master")
+        if attrs.get("branch", "master") != "master":
```
beckermr, line 352:
```suggestion
        if branch != "master":
```

beckermr:
I think until we can do the outputs things with graph nodes per branch, this is the best solution.

beckermr:
When I say only master matters if it is closed, that is in reference only for checking whether it as a parent is done. The bot will still issue all of the prs even if the one on master is merged afaik.

CJ-Wright:
I think this looks ok. Will the branch pr always be made first?

beckermr:
They all get made in the same job one after the other. It should be close enough.

beckermr:
Any other comments @CJ-Wright or shall we merge?

CJ-Wright:
Nope, please feel free to go ahead!

PR #1285, opened 2020-11-15
ENH set outputs again

beckermr:
This is a follow on to #1122 by @CJ-Wright. I have copied bits of code from there and made some slight changes to how it works. We now keep all outputs in the lut and (at least attempt) to always put requirements through it to find feedstocks in the graph.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -663,7 +663,9 @@ def create_migration_yaml_creator(migrators: MutableSequence[Migrator], gx: nx.D
         # TODO - we are grabbing one element almost at random here
         # the sorted call makes it stable at least?
         fs_name = next(
-            iter(sorted(gx.graph["outputs_lut"].get(package_name, {package_name}))),
+            iter(
+                sorted(gx.graph["outputs_lut"].get(package_name, {package_name})),
+            ),
         )
 
         if (
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -9,6 +9,25 @@ import logging
 import os
 import typing
 from subprocess import CalledProcessError
+from typing import (
+    Optional,
+    MutableSequence,
+    MutableSet,
+    Sequence,
+    Tuple,
+    Set,
+    Mapping,
+    MutableMapping,
+    Union,
+)
+
+if typing.TYPE_CHECKING:
+    from .cli import CLIArgs
+    from .migrators_types import (
+        MetaYamlTypedDict,
+        PackageName,
+        MigrationUidTypedDict,
+    )
 
 # from conda_forge_tick.profiler import profiling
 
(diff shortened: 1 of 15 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -627,30 +628,42 @@ def create_migration_yaml_creator(migrators: MutableSequence[Migrator], gx: nx.D
             config=Config(**CB_CONFIG),
         )
     feedstocks_to_be_repinned = []
-    for k, package_pin_list in pinnings.items():
+    for pinning_name, package_pin_list in pinnings.items():
+        # there are three things:
+        # pinning_name - entry in pinning file
+        # package_name - the actual package, could differ via `-` -> `_`
+        #                from pinning_name
+        # feedstock_name - the feedstock the outputs the package
```
CJ-Wright, line 636:
```suggestion
        # feedstock_name - the feedstock that outputs the package
```

beckermr:
The bot is now generating a migration pr for libv8 and openturns whereas it was not before. I'll need to track that down.

conda_forge_tick/feedstock_parser.py
```diff
@@ -0,0 +1,346 @@
...
+        return sub_graph
+    sub_graph["raw_meta_yaml"] = meta_yaml
+
+    # Get the conda-forge.yml
+    if isinstance(conda_forge_yaml, str):
+        sub_graph["conda-forge.yml"] = {
+            k: v
+            for k, v in yaml.safe_load(conda_forge_yaml).items()
+            if k
+            in {
```
CJ-Wright, line 165: This should most likely be pulled up into a constant, and then eventually out into smithy?

conda_forge_tick/make_graph.py
```diff
@@ -57,13 +59,36 @@
...
+    return deps
+
+
+def make_outputs_lut_from_graph(gx):
+    outputs_lut = defaultdict(set)
+    for node_name, node in gx.nodes.items():
+        for k in node.get("payload", {}).get("outputs_names", []):
+            if node_name != "pypy-meta":
+                outputs_lut[k].add(node_name)
+            elif k == "pypy":
```
CJ-Wright, line 79: Is this correct? It seems that this will never be triggered.

CJ-Wright:
I have some questions but otherwise this looks great!

conda_forge_tick/feedstock_parser.py
beckermr: This the list of keys we keep from the conda-forge.yml. Happy to put it as a constant but it should not be in smithy. We don't keep all of the keys, likely due to it having had encrypted secrets in it.

conda_forge_tick/make_graph.py
beckermr: Yes. It is correct. The pypy-meta node (which is a feedtsock), has python, cffi and pypy as outputs:

https://github.com/conda-forge/pypy-meta-feedstock/blob/master/recipe/meta.yaml#L48

We only want to map it to pypy.

conda_forge_tick/feedstock_parser.py
beckermr: Moved to a global.

CJ-Wright (approved):
This looks awesome!

PR #1286, opened 2020-11-15
BUG consider branches when filtering

beckermr:
This should be the last bug for branch PRs. 🤞


beckermr:
Works locally. I'm gonna try it.

PR #1288, opened 2020-11-16
BUG checkout properly and do not check solvable for branches

beckermr:
I think this should be the last of the bugs.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -237,17 +237,26 @@ def run(
                 )
             ]
 
-    if (
-        migrator.check_solvable
-        # we always let stuff in cycles go
-        and feedstock_ctx.attrs["name"] not in getattr(migrator, "cycles", set())
-        # we always let stuff at the top go
-        and feedstock_ctx.attrs["name"] not in getattr(migrator, "top_level", set())
-        # for solveability always assume automerge is on.
-        and feedstock_ctx.attrs["conda-forge.yml"].get("bot", {}).get("automerge", True)
-    ) or feedstock_ctx.attrs["conda-forge.yml"].get("bot", {}).get(
-        "check_solvable",
-        False,
+    if base_branch == "master" and (
+        (
+            migrator.check_solvable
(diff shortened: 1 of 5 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -826,7 +835,7 @@ def _compute_time_per_migrator(mctx, migrators):
         else:
             num_nodes.append(len(mmctx.effective_graph.nodes))
 
-    num_nodes_tot = sum(num_nodes)
+    num_nodes_tot = max(sum(num_nodes), 1)
```
CJ-Wright, line 838: Is this correct? Some migrations may have no nodes, like the migration migrator.

beckermr: This is to prevent a divide by zero in the next line. I will move it there.

beckermr: fixed

conda_forge_tick/auto_tick.py
```diff
@@ -849,7 +859,7 @@ def _compute_time_per_migrator(mctx, migrators):
         time_per_migrator[i] = time_per_migrator[i] * time_fac
 
     # recompute the total here
-    tot_time_per_migrator = sum(time_per_migrator)
+    tot_time_per_migrator = max(sum(time_per_migrator), 1)
```
CJ-Wright, line 862: Can this also be zero?

CJ-Wright:
Also why no solve check for branches?

conda_forge_tick/auto_tick.py
beckermr: Yes. We divide by it later. It is only used for generating output IIRC. I'll move this call too.

beckermr:
> Also why no solve check for branches?

We cannot report the errors properly on the status page, so it makes more sense to let the PRs through. Otherwise, people will not get branch PRs and will have to dig into the bot to figure out why.

This goes in the general direction of keeping branches as second class citizens until we can do a more proper job.

beckermr:
Any more comments @CJ-Wright?

PR #1297, opened 2020-11-17
exclude the pkgs being migrated themselves

CJ-Wright:
@beckermr I think this fixes the migrate arrow-cpp pr into arrow-cpp branches.

### DIFF
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -471,6 +471,8 @@ def create_rebuild_graph(
 ) -> nx.DiGraph:
     total_graph = copy.deepcopy(gx)
     excluded_feedstocks = set() if excluded_feedstocks is None else excluded_feedstocks
+    # Always exclude the packages themselves from the migration
+    excluded_feedstocks.update({gx.graph['outputs_lut'].get(node, node) for node in package_names})
 
     included_nodes = set()
 
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -473,7 +473,7 @@ def create_rebuild_graph(
     excluded_feedstocks = set() if excluded_feedstocks is None else excluded_feedstocks
     # Always exclude the packages themselves from the migration
     for node in package_names:
-        excluded_feedstocks.update(gx.graph["outputs_lut"].get(node, node))
+        excluded_feedstocks.update(gx.graph["outputs_lut"].get(node, {node}))
 
     included_nodes = set()
 
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -472,7 +472,9 @@ def create_rebuild_graph(
     total_graph = copy.deepcopy(gx)
     excluded_feedstocks = set() if excluded_feedstocks is None else excluded_feedstocks
     # Always exclude the packages themselves from the migration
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
This PR appears wrong. The output of outputs_lut is itself a python set.

beckermr:
Do we know why these feedstocks were being excluded before but aren't now?

CJ-Wright:
They might have a self edge, since one of their outputs depends on itself (pyarrow and arrow-cpp). I think that if the bot went to issue a PR into arrow-cpp it would come back empty handed, since there would be no changes and pass over it.

beckermr:
> They might have a self edge, since one of their outputs depends on itself (pyarrow and arrow-cpp). I think that if the bot went to issue a PR into arrow-cpp it would come back empty handed, since there would be no changes and pass over it.

I don't follow.

conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -471,6 +471,9 @@ def create_rebuild_graph(
 ) -> nx.DiGraph:
     total_graph = copy.deepcopy(gx)
     excluded_feedstocks = set() if excluded_feedstocks is None else excluded_feedstocks
+    # Always exclude the packages themselves from the migration
+    for node in package_names:
+        excluded_feedstocks.update(gx.graph["outputs_lut"].get(node, node))
```
beckermr, line 476:
```suggestion
        excluded_feedstocks.update(gx.graph["outputs_lut"].get(node, {node}))
```

beckermr: The default needs to be a set too.

CJ-Wright:
There is nothing that explicitly removes the feedstock that causes the migration from the migration itself. If the feedstock doesn't depend on itself then it won't be added, but if it depends on itself, in this case it had an output with a self dependency, then it will remain in the graph.

The reason that we haven't issued PRs like this before is because the branch we've always issued PRs into hasn't needed a rerender (since it would have recently been rerendered in the version bump to cause the migration). In this case the arrow branches needed a rerender. The bot checks to see if the migration made any substantive changes in the codebase, if none it writes the migration for that feedstock off as being finished already. Since it made changes it wasn't written off and so a PR was opened [here](https://github.com/conda-forge/arrow-cpp-feedstock/pull/216).

This PR will cause the migration to always exclude the instigating feedstock since it could be problematic/weird down the road.

beckermr:
Thank you for the explanation! 

So a question. Are there any pinned packages provided by more than one feedstock? If so, are we going to exclude more feedstocks than we should here? 

Maybe it is better to have the bot rely on the rerendering. I think this is a one time case because the migration has been ongoing. I don't think it will happen every time given the explanation above.

CJ-Wright:
I think we're doing the needed exclusions here. I think the closest we get is blas or mpi, but I'm not certain if we move those all at once. Do they share a common API that gets moved in sync?

I guess the nasty answer to that is yes, python but the cpython and pypy apis are a bit disconnected at the moment so it is a moot point.

beckermr:
Right. Also, should we add code to the create_edges function to remove self-references at the start? I suspect we are producing more of them now.

CJ-Wright:
Those may be real references. I'd leave them in for the moment, and pull them out as needed.

beckermr:
> Those may be real references.

How so? 

If they show up as cycles later then the bot will issue gobs of PRs without considering the tree.

beckermr:
To be clear, I am talking about an edge of a feedstock to itself.

CJ-Wright:
Right we currently remove those at the end when creating the sub graph.

PR #1312, opened 2020-11-24
extract missing deps via set overlaps with providing pkgs

CJ-Wright:


conda_forge_tick/audit.py
```diff
@@ -54,6 +56,8 @@
     "versioneer",
     # not a real dep
     "cross-python",
+    "c_compiler_stub",
+    "cxx_compiler_stub",
```
beckermr, line 60: you might as well put the rest of them in here, especially fortran_compiler_stub and cuda_compiler_stub

CJ-Wright: Do we have a list of all the stubs?

CJ-Wright: nevermind I got it

conda_forge_tick/audit.py
```diff
@@ -50,14 +50,18 @@
...
+    {
+        "python",
+        "setuptools",
+        "pip",
+        "versioneer",
+        # not a real dep
+        "cross-python",
+    }
+    | BUILTINS
+    | set(COMPILER_STUBS_WITH_STRONG_EXPORTS)
```
beckermr, line 63: This is fine, but do be aware that this is not the full list of possible stubs. Any compiler that only statically links is not in this list. I think this code is fine. This comment is an FYI.

CJ-Wright: Sure, and we should document all the stubs somewhere, but something tells me that the actual answer is hidden somewhere in conda-build or pinnings, and I didn't have my spelunking hat on :construction_worker_man:

beckermr: Yep. The answer is in pinnings and changes as that file changes. So much fun!

beckermr:
huzzah!

PR #1316, opened 2020-12-01
Don't bother generating pypi mappings for archived feedstocks

mariusvniekerk:
This should prevent the situation where we might prefer archived feedstocks

### DIFF
--- a/conda_forge_tick/pypi_name_mapping.py
+++ b/conda_forge_tick/pypi_name_mapping.py
@@ -20,8 +20,10 @@ from os.path import commonprefix
 from .utils import load, as_iterable, load_graph, dump, loads
 
 
-def load_node_meta_yaml(filename: str) -> Dict[str, str]:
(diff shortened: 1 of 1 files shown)


conda_forge_tick/pypi_name_mapping.py
```diff
@@ -154,6 +156,8 @@ def extract_pypi_information(cf_graph: str) -> List[Dict[str, str]]:
     # TODO: exclude archived node_attrs
     for f in list(glob.glob(f"{cf_graph}/node_attrs/*.json")):
         meta_yaml = load_node_meta_yaml(f)
+        if meta_yaml is None:
```
CJ-Wright, line 159: I think the next clause covers this

mariusvniekerk: it does, but this keeps mypy completely happy

PR #1350, opened 2021-02-13
REF use an allow list for curl

beckermr:
Here is a version of #1348 that we can merge right away. I have put in a specific list of things that use the new `curl`-based test. everything else uses the old `wget`-based one. This should satisfy all parties.

@CJ-Wright want to do the honors?

cc @isuruf 

closes #1348

### DIFF
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -1,6 +1,5 @@
 import abc
 import collections.abc
-import subprocess
 import re
 import logging
 import typing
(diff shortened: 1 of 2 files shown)


conda_forge_tick/update_sources.py
```diff
@@ -32,6 +32,11 @@
 
 logger = logging.getLogger("conda_forge_tick._update_version.update_sources")
 
+CURL_ONLY_URL_SLUGS = [
+    "https://eups.lsst.codes/",
+    "ftp://ftp.info-zip.org",
```
beckermr, line 37:
```suggestion
    "ftp://ftp.info-zip.org/",
```

conda_forge_tick/update_sources.py
```diff
@@ -343,21 +348,37 @@ def get_sha256(url: str) -> Optional[str]:
...
-        return False
-    # For FTP servers an exception is not thrown
-    if "No such file" in output.decode("utf-8"):
-        return False
-    if "not retrieving" in output.decode("utf-8"):
-        return False
-
-    return True
+    """
+    We use curl here, as opposed requests.head, because github urls redirect
```
beckermr, line 352:
```suggestion
    We use curl/wget here, as opposed requests.head, because github urls redirect
```

conda_forge_tick/update_sources.py
```diff
@@ -343,21 +348,37 @@ def get_sha256(url: str) -> Optional[str]:
...
-    # For FTP servers an exception is not thrown
-    if "No such file" in output.decode("utf-8"):
-        return False
-    if "not retrieving" in output.decode("utf-8"):
-        return False
-
-    return True
+    """
+    We use curl/wget here, as opposed requests.head, because github urls redirect
+    with a 3XX code even if the file doesn't exist.
```
beckermr, line 353:
```suggestion
    We use curl/wget here, as opposed requests.head, because 
     - github urls redirect with a 3XX code even if the file doesn't exist
     - requests cannot handle ftp
```

PR #1351, opened 2021-02-13
use the full graph to run the impact analysis of a pin change

CJ-Wright:
CC: @beckermr


beckermr (commented):
So what I cannot tell from this is if `self.graph` is used in more than one way here. Can we pass in a `full_graph` parameter instead and set a new attribute to make sure nothing else breaks?

CJ-Wright:
Sure, I did try to track down if it is used elsewhere and came up with nothing.

beckermr:
Ahh ok. If you are confident in that then merge away!

CJ-Wright:
I've pushed the requested changes

beckermr (approved):

PR #1360, opened 2021-02-18
REF replace pull with hard reset to upstream

beckermr:
cc @chrisburr


beckermr:
I'm gonna roll with this. We can simplify later.

PR #1363, opened 2021-02-22
check all noarch for arch migrations

### LINKED ISSUE #1362: Handle recipes with all outputs noarch
Currently we don't consider pkgs with all outputs noarch to be noarch, see: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/migrators/arch.py#L98

This causes things like `black` to get hit with recompile migrators where they are not needed

CJ-Wright:
Closes #1362

CC: @beckermr


conda_forge_tick/migrators/arch.py
```diff
@@ -10,7 +10,7 @@
 from conda_forge_tick.utils import frozen_to_json_friendly, pluck, as_iterable
 from conda_forge_tick.xonsh_utils import indir
 from conda_forge_tick.make_graph import get_deps_from_outputs_lut
-
+from .migration_yaml import _all_noarch
```
beckermr, line 13: We may want to declare this a public API by removing the underscore.

CJ-Wright: done

beckermr (approved):

PR #1364, opened 2021-02-23
Only check output noarch if needed

CJ-Wright:


CJ-Wright:
I didn't see large movement among any of of our major migrations, mostly removing black

PR #1371, opened 2021-02-28
ENH ship rerender errors for version PRs anyways

### LINKED ISSUE #1370: don't block version bump PRs on rerender errors
If version migrations don't rerender properly, maybe we should issue them, make a comment in the PR to trigger feedstock rerendering, and then call it good.

beckermr:
This PR has the bot ship PRs for versions that hash properly but cannot be rerendered locally. Hopefully, this will mean we let more maintainers know about possible version bumps and they can fix the rerendering errors.

closes #1370


CJ-Wright (approved):
LGTM! thank you!

PR #1417, opened 2021-07-16
BUG skip path sources if other url sources are there

beckermr:
This PR skips path sources if other URL sources are there. It needs a test using the conda-standalone feedstock.


beckermr:
Going to push this without a specific test for now since it doesn't appear to break anything.

PR #1420, opened 2021-07-30
Support pre-release versions from PyPI

mariusvniekerk:
In order to properly support packages that exist ONLY as prerelease (eg portions of opentelemetry) we need to allow pypi to return that version


ocefpaf:
I have to confess that when we need this kind of workaround I truly believe that the packaging world is just broken! Oficial packages, recommended packages, only exist as pre-releases, and `pip` installs them as non-pre-releases... why!?

CJ-Wright (approved):
LGTM, thank you!

PR #1451, opened 2021-09-19
ENH add PRs for broken packages

beckermr:
@isuruf here is the migrator for the broken travis stuff.


beckermr:
It doesn't handle versions and branches properly, which is next to impossible anyways. So we'll have to handle the python ones by hand for example.

isuruf (approved):

PR #1476, opened 2021-11-03
ENH set memory limit

beckermr:
trying this to help debug


conda_forge_tick/auto_tick.py
```diff
@@ -1241,8 +1241,19 @@ def _run_migrator(migrator, mctx, temp, time_per, dry_run):
     return good_prs
 
 
+def _setup_limits():
+    import resource
+    limit_gb = 7.0
```
mariusvniekerk, line 1246: I'd STRONGLY recommend reading this from an environment variable as it makes changing it at the bot level much simpler than a code redeploy

PR #1509, opened 2022-02-17
Expand grayskull pypi mapping to include common matches

mariusvniekerk:
Instead of constraining grayskull mappings to only cases where there are mismatches, returrn the whole set. 

This should allow some degree of detection of the case where there is a pypi package `A` and a conda package `A` and the two have nothing in common.


beckermr:
thank you!

Issue #1513, opened 2022-03-10
How to update name mapping?


saulshanabrook:
I am writing to ask how to update the mappings from the [`grayskull_pypi_mapping.yaml` file](https://github.com/regro/cf-graph-countyfair/blob/master/mappings/pypi/grayskull_pypi_mapping.yaml).

We had an issue opened in beni (https://github.com/Quansight-Labs/beni/issues/17) which wanted to update the `seaborn` pypi package to map to the `seaborn` conda package instead of `seaborn-split`, which does not exist.

Thanks! I appreciate the existence of this mapping!

mdekstrand:
Some further details:

The [Seaborn recipe](https://github.com/conda-forge/seaborn-feedstock/blob/main/recipe/meta.yaml) names itself `seaborn-split`, but does not actually emit such a package - it emits `seaborn` and `seaborn-base`. However, it looks like the autotick bot is not able to pick up that correctly, and believes the `seaborn` PyPI package maps to `seaborn-base`.

Regro devs - this means, among other things, that `conda-lock` will not work for any project that uses a PyPI-named dependency source and depends on `seaborn`.

beckermr:
@mariusvniekerk can you help with this seaborn thing?

mariusvniekerk:
Sure just make an entry to https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/pypi_name_mapping_static.yaml

mdekstrand:
Just submitted a PR (#1533) to do that.

mdekstrand:
The change in #1533 seems to no longer be active - the PyPI mapping is mapping `seaborn` to `seaborn-split` again, although the static mapping is in the static name mapping file.

PR #1533, opened 2022-06-08
Add Seaborn static mapping

mdekstrand:
The Seaborn package has some oddities going on with its package names (#1513), which this static override fixes.


beckermr:
Usually when there is a base and non-base package, conda-forge prefers the base one. Do we know why the seaborn feedstock was split this way?

beckermr:
Looks like they are avoiding pulling in statsmodels.

mdekstrand:
The Seaborn PyPI package only depends on `statsmodels` through an extra, so I think routing this to `seaborn-base` would be appropriate. Will update.

PR #1538, opened 2022-07-21
Hatchet-level fix to disable depfinder usage in the bot

mariusvniekerk:


beckermr:
Thanks for this. I don't think we'll need it with the latest depfinder but let's keep it open just in case.

beckermr:
closing as no longer needed!

PR #1539, opened 2022-07-26
Add static mapping for ipopt/cyipopt

zmbc:
The cyipopt package on conda is published under two names on PyPI: ipopt and cyipopt. It also exports
both 'cyipopt' (preferred) and 'ipopt' (now deprecated) packages. I am not sure whether all these combinations
need to be listed or if some are redundant. I'm also not sure if having duplicates like this works.

### DIFF
--- a/conda_forge_tick/pypi_name_mapping_static.yaml
+++ b/conda_forge_tick/pypi_name_mapping_static.yaml
@@ -40,10 +40,6 @@
   import_name: xxhash
   conda_name: python-xxhash
 
-- pypi_name: docker
-  import_name: docker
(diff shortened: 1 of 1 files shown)


conda_forge_tick/pypi_name_mapping_static.yaml
```diff
@@ -40,10 +40,6 @@
   import_name: xxhash
   conda_name: python-xxhash
 
-- pypi_name: docker
```
zmbc, line 43: This is duplicated above.

zmbc:
Is anyone able to review this change?

PR #1543, opened 2022-08-01
TST added test for dbt

beckermr:
This PR adds a test for a missed version PR.


maresb:
Thanks so much for looking into this!

I think there could be some bizarre issue with the URL:

https://pypi.io/packages/source/d/dbt-core/dbt-core-1.2.0.tar.gz

It wasn't working when the package was first released on PyPI (404 if I remember correctly). Then a day or two later it started working. I wasn't sure how to debug this since it's not the standard PyPI link, and it seems to be some sort of server-side issue.

Have you encountered anything similar with other packages?

beckermr:
There was a temporary bug in pypi that I think was fixed.

PR #1544, opened 2022-08-04
Add json map to mapping output

ericdill:
i'd like to get this info in json too. it's programmatically used in depfinder in its yaml format which is substantially slower to load than json (~30x). i _think_ this should just transparently work. this will change the output format of `mappings/pypi/import_name_priority_mapping.yaml` because json.dump doesn't know what to do with `list_reverseiterator`
```
TypeError: Object of type list_reverseiterator is not JSON serializable
```

Is updating the output of import_name_priority_mapping.yaml to look like the following snippet instead going to be a problem for anything else?

```diff
-  ranked_conda_names: !!python/object/apply:builtins.reversed
-    args:
-    - - pycryptodome
-      - pycrypto
-    state: 1
+  ranked_conda_names:
+  - pycrypto
+  - pycryptodome
```

### DIFF
--- a/conda_forge_tick/pypi_name_mapping.py
+++ b/conda_forge_tick/pypi_name_mapping.py
@@ -306,7 +306,7 @@ def main(args: "CLIArgs") -> None:
 
     yaml_dump = functools.partial(yaml.dump, default_flow_style=False, sort_keys=True)
     # import pdb; pdb.set_trace()
-    for dumper, suffix in ((yaml_dump, 'yaml'), (json.dump, 'json')):
+    for dumper, suffix in ((yaml_dump, "yaml"), (json.dump, "json")):
         with (dirname / f"grayskull_pypi_mapping.{suffix}").open("w") as fp:
             dumper(grayskull_style, fp)
 
--- a/conda_forge_tick/pypi_name_mapping.py
+++ b/conda_forge_tick/pypi_name_mapping.py
@@ -6,6 +6,7 @@ Builds and maintains mapping of pypi-names to conda-forge names
 """
 
 import glob
+import json
 
 import requests
 import yaml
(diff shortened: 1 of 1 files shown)


ericdill:
cc @mariusvniekerk @beckermr

mariusvniekerk:
Presumably just using `safe_dump` fixes all of it?

conda_forge_tick/pypi_name_mapping.py
```diff
@@ -304,24 +305,25 @@ def main(args: "CLIArgs") -> None:
     dirname.mkdir(parents=True, exist_ok=True)
 
     yaml_dump = functools.partial(yaml.dump, default_flow_style=False, sort_keys=True)
```
mariusvniekerk, line 307:
```suggestion
    yaml_dump = functools.partial(yaml.safe_dump, default_flow_style=False, sort_keys=True)
```

mariusvniekerk (approved):
Looks good, If we use safe_dump we can avoid some of the bad behavior automatically

Eladrian05:
Desa

Eladrian05:
> i'd like to get this info in json too. it's programmatically used in depfinder in its yaml format which is substantially slower to load than json (~30x). i _think_ this should just transparently work. this will change the output format of `mappings/pypi/import_name_priority_mapping.yaml` because json.dump doesn't know what to do with `list_reverseiterator`
> 
> ```
> 
> TypeError: Object of type list_reverseiterator is not JSON serializable
> 
> ```
> 
> 
> 
> Is updating the output of import_name_priority_mapping.yaml to look like the following snippet instead going to be a problem for anything else?
> 
> 
> 
> ```diff
> 
> -  ranked_conda_names: !!python/object/apply:builtins.reversed
> 
> -    args:
> 
> -    - - pycryptodome
> 
> -      - pycrypto
> 
> -    state: 1
> 
> +  ranked_conda_names:
> 
> +  - pycrypto
> 
> +  - pycryptodome
> 
> ```
> 
>

PR #1561, opened 2022-11-02
Add __win to faked virtual packages

mbargull:
This fixes the
```
Error: not solvable [...] Encountered problems while solving:\n - nothing provides __win needed by [...]
```
seen in the Py 3.11 migration.


beckermr:
Thank you!

PR #1598, opened 2023-02-03
rewrite CLI using typer

ericdill:
The current CLI is confusing and arcane, requiring the caller to know what different integers mean. This rewrites it entirely and provides flexibility for each entrypoint to have different args that can be passed to it specifically. I'm interested in feedback. 

Companion PR to the bot https://github.com/regro/autotick-bot/pull/579

for now this is going in as "cli2.py" so it's easier to review the changes


conda_forge_tick/pypi_name_mapping.py
```diff
@@ -264,7 +264,7 @@ def score(pkg_name):
 
     pkgs = list(gx.graph["outputs_lut"])
     ranked_list = list(sorted(pkgs, key=score))
-    with open(pathlib.Path(cf_graph) / "ranked_hubs_authorities.json", "w") as f:
```
beckermr, line 267: Are we concerned that removing this effective type cast will break things if we have been feeding strings to this function by accident?

conda_forge_tick/auto_tick.py
```diff
@@ -1303,26 +1302,32 @@ def _setup_limits():
...
 
 # @profiling
-def main(args: "CLIArgs") -> None:
+def main(
+    debug: bool = False,
+    dry_run: bool = False,
+    github_username: str = "",
+    github_password: str = "",
+    github_token: str = "",
+) -> None:
```
beckermr, line 1311: Don't we need to change the old CLI functions to accommodate the change in function signature here?

beckermr (changes requested):

beckermr:
Thanks for the work here. Happy to merge as long as we don't break anything.

conda_forge_tick/auto_tick.py
ericdill: my intent with this PR is to have this replace the existing CLI with its integer-based invoking. It seems like an unusual pattern that isn't understandable without reading the code. This initial review has them separate for readability.

conda_forge_tick/pypi_name_mapping.py
ericdill: the only place this function is called is from the `main` function a few dozen lines down. That main function gets called from `mappings.py` but that module seems entirely superfluous and so can just be removed and replaced with a direct call to the function it's wrapping

conda_forge_tick/auto_tick.py
ericdill: if we want to keep the old CLI around then I can update the way they're invoked. Or I can make another layer of indirection and have `cli2` call into a corresponding wrapper and leave `main` alone. but the whole intent here is to update the CLI of this project with modern CLI infrastructure to simplify the code base not add another level of abstraction. Let me know your thoughts

beckermr: Right, my concern is only that if this PR breaks the old CLI, then we will break the bot since it runs live right off master. So we either need to do the work to start versioning things, or we need to make sure everything is backward compatible for at least a bit.

ericdill: ohhhh got it okay. so the new plan will be:

1. in this PR, add cli2.py and wrappers around each of the main() functions so that cli.py doesn't break
2. merge the bot PR that switches stuff to cli2.py
3. make another PR here that cleans up the old cli and gets rid of the shims. this pr should be transparent to the bot code

beckermr: That's perfect. We gotta do work on the truck while we're hauling ass at 70 MPH down the highway.

beckermr:
closing due to the conflicts and subsequent changes to the bot infrastructure.

PR #1606, opened 2023-02-19
PROD remove xonsh env and add version

beckermr:
This PR removes the last bit of xonsh env and adds the version to the package.


conda_forge_tick/utils.py
```diff
@@ -84,6 +84,19 @@ def pushd(new_dir):
...
+def env_swap(var, val):
+    has_var = var in os.environ
+    old_val = os.environ.get(var, None)
+
+    try:
+        os.environ[var] = val
+        yield
+    finally:
+        if has_var:
+            os.environ[var] = old_val
```
beckermr, line 97:
```suggestion
            os.environ[var] = old_val
        else:
            del os.environ[var]
```

PR #1640, opened 2023-03-25
Refactor conda_forge_tick.pypi_name_mapping

maresb:
This is in preparation for various fixes I have incoming. These changes should not alter functionality.

See commit comments for more details on some individual changes.

PR #1641, opened 2023-03-25
Prefer statically defined matches in PyPI import mapping

maresb:
This PR consists of a single commit 2406b43 on top of #1640.

Due to concatenating the graph mapping and the static mapping, there might be two candidates with the same `conda_name` but different PyPI names. Currently we pick an arbitrary one (whichever one appears last in the list). This PR changes this behavior to favor matches defined in the static mapping, and to otherwise emit a warning.

### DIFF
--- a/conda_forge_tick/pypi_name_mapping.py
+++ b/conda_forge_tick/pypi_name_mapping.py
@@ -335,7 +335,11 @@ def determine_best_matches_for_pypi_import(
                 f"needs {import_name} <- provided_by: {ranked_conda_names} : "
                 f"chosen {winning_name}",
             )
-        final_map[import_name] = map_by_conda_name[winning_name]
+        # Due to concatenating the graph mapping and the static mapping, there might
+        # be multiple candidates with the same conda_name but different PyPI names.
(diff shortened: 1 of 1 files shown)


maresb:
This has the effect:

```diff
diff --git a/mappings/pypi/grayskull_pypi_mapping.yaml b/mappings/pypi/grayskull_pypi_mapping.yaml
index b8d8a4f93915..f92ff78979a9 100644
--- a/mappings/pypi/grayskull_pypi_mapping.yaml
+++ b/mappings/pypi/grayskull_pypi_mapping.yaml
@@ -32963,6 +32963,13 @@ pyqstrat:
   import_name: pyqstrat
   mapping_source: regro-bot
   pypi_name: pyqstrat
+pyqt4:
+  conda_name: pyqt
+  delimiter_max: 5.0.0
+  delimiter_min: 4.0.0
+  import_name: PyQt4
+  mapping_source: static
+  pypi_name: pyqt4
 pyqt5:
   conda_name: pyqt
   delimiter_max: 6.0.0
```

PR #1642, opened 2023-03-25
Prefer statically defined matches in PyPI Grayskull mapping

maresb:
This PR consists of a single commit 98e51a2 on top of #1640.

Since the input mappings are indexed by `import_name`, it sometimes happens that there are multiple `Mapping` objects with the same `pypi_name`. Currently we pick an arbitrary one, but this PR makes it explicit that we should favor one from the static mapping if it's defined. Otherwise emit a warning.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,7 @@ repos:
         args:
           - --ignore-words-list=rever,pring,pullrequest,pullrequests
           - --exclude-file=.codespellignorelines
+          - -L=statics
   # - repo: https://gitlab.com/pycqa/flake8
   #   rev: 3.8.4
   #   hooks:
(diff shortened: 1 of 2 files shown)


maresb:
This has the effect:

```diff
diff --git a/mappings/pypi/grayskull_pypi_mapping.yaml b/mappings/pypi/grayskull_pypi_mapping.yaml
index b8d8a4f93915..5e4a2ce764ea 100644
--- a/mappings/pypi/grayskull_pypi_mapping.yaml
+++ b/mappings/pypi/grayskull_pypi_mapping.yaml
@@ -37311,9 +37311,9 @@ rtxpy:
   mapping_source: regro-bot
   pypi_name: rtxpy
 ruamel.yaml:
-  conda_name: ruamel_yaml
-  import_name: ruamel_yaml
-  mapping_source: regro-bot
+  conda_name: ruamel.yaml
+  import_name: ruamel.yaml
+  mapping_source: static
   pypi_name: ruamel.yaml
 ruamel.yaml.jinja2:
   conda_name: ruamel.yaml.jinja2
```

PR #1644, opened 2023-03-26
Clip tiny hub/authority scores to zero

maresb:
This is somewhat of a follow-up to #1638. Thanks so much @beckermr for the quick merge on that!

Tiny values are evidently due to rounding error. (There is a large gap in orders of magnitude between positive values and these tiny values.)

This should stabilize `ranked_hubs_authorities.json`.

### DIFF
--- a/conda_forge_tick/pypi_name_mapping.py
+++ b/conda_forge_tick/pypi_name_mapping.py
@@ -232,6 +232,11 @@ def determine_best_matches_for_pypi_import(
     # whilst authorities are packages with many edges to them.
     hubs, authorities = networkx.hits(gx)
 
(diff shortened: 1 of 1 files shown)


beckermr (commented):
Did you mean to handle only positive values?

maresb:
Yes, up to rounding error, there are only positive values:

```python
hvals = np.array(list(hubs.values()))
hvals.min()  # -2.0032189385215855e-22
```

Similarly with authorities.

maresb:
Thanks a lot @beckermr, this helps a lot!

maresb:
It's working!!! :tada: 

> [no files to commit](https://github.com/regro/cf-scripts/actions/runs/4525396815/jobs/7969857766#step:7:20)

Issue #1595, opened 2023-01-27
wait_for_migrators not working anymore


h-vetinari:
Hey all

I just noticed that `wait_for_migrators:` doesn't seem to work as intended anymore.

For background: the aws-c-* ecosystem has a lot of small migrations, and letting them run at the same time -- e.g. when someone comes and merges the bot PRs in the pinning repo -- means that some migrations end up getting blocked by a diamond-shaped upgrade race condition between the bots. For this reason, it's very convenient to chain them together using `wait_for_migrators`, and then they used to run through on auto-pilot (here's an [example](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3725) from November 2022 where this worked).

I now did the same in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3991, but even though the first migration is fully done, the status bot thinks all feedstocks of the following migrator are "waiting for parents" (see [status page](https://conda-forge.org/status/#other_migrations)).

I had a look at the most recent changes here (#1517, a while ago), and also looked at the most recent migrator [run](https://github.com/regro/autotick-bot/actions/runs/4027429681/jobs/6923150135), where I cannot even find the phrase "need to wait for" that should be added by the logger (or maybe it's not shown because debug-level logging is not shown by default?).

Would be nice if we could keep using this feature.

CC @isuruf @becker @jakirkham

beckermr:
This will require some debugging of the migrator by hand I bet. I am not sure when I'll have time to get to it.

h-vetinari:
OK, thanks.

Could we please merge https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4003 in the meantime?

beckermr:
Do you have a current example where this is broken?

h-vetinari:
> Do you have a current example where this is broken?

See https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4003 (the respective migrators did not get triggered even though all the ancestors had been built)

beckermr:
Right sorry. I need a current ongoing migration where things look busted and the wait for migrators is set. To debug the code I need the live state of the bot.

PR #1655, opened 2023-04-18
add platform_allowlist for migrator

h-vetinari:
Very specifically for https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1359, though hopefully useful in general.

I thought about this a bit, but I'm completely new to this codebase, and it's a bit hard to penetrate TBH. I mainly tried to follow Matthew's [idea](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1359#issuecomment-1029945907).

The migrator for fortran-only-on-windows would then look like:
```
__migrator:
  kind:
    version
  migration_number: 1
  build_number: 1
  longterm: True
  platform_allowlist:  # new
    - win_64           # new
  override_cbc_keys:
    - fortran_compiler_stub
```

The underscore in `win_64` comes from the same pattern [elsewhere](https://github.com/regro/cf-scripts/blob/36cdec4560688d99aeced594f9d7e5da10e093b2/conda_forge_tick/feedstock_parser.py#L321) in `feedstock_parser.py`, but unsurprisingly I don't care whether it's `-` or `_` (or whatever).

The tricker bit is writing tests for this. I tried to look for tests for the similar `wait_for_migrators`, but this has no tests either (which is probably why it's [broken](https://github.com/regro/cf-scripts/issues/1595))... Advice & help welcome

### DIFF
--- a/conda_forge_tick/feedstock_parser.py
+++ b/conda_forge_tick/feedstock_parser.py
@@ -221,7 +221,7 @@ def populate_feedstock_attributes(
             ci_support_files = glob.glob(
                 os.path.join(feedstock_dir, ".ci_support", "*.yaml"),
             )
-            varient_yamls = []
+            variant_yamls = []
             plat_arch = []
             for cbc_path in ci_support_files:
                 LOGGER.debug("parsing conda-build config: %s", cbc_path)
@@ -242,7 +242,7 @@ def populate_feedstock_attributes(
                         break
                 plat_arch.append((plat, arch))
 
-                varient_yamls.append(
+                variant_yamls.append(
                     parse_meta_yaml(
                         meta_yaml,
                         platform=plat,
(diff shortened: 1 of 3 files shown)


beckermr:
Thank you so much!

From the name "platform_restriction" I cannot tell if the list is platforms to skip or platforms to keep. Can we think of another name?

beckermr:
In terms of tests, writing them for this part of the bot is next to impossible.

h-vetinari:
> From the name "platform_restriction" I cannot tell if the list is platforms to skip or platforms to keep. Can we think of another name?

Happy to bikeshed, not attached to any particular name. For me, "filter" can be both a positive (e.g. air filter) or a negative (e.g. dust filter), so I tried to avoid it. I guess I pictured "restrict to" and not "restricted", so didn't notice it's also ambivalent.

Maybe `__migrator.platform_constraint`?

beckermr:
Yes sorry to nitpick. Platform allow list is perfect.

h-vetinari:
> In terms of tests, writing them for this part of the bot is next to impossible.

Could we not have a micro-version of conda-forge (say 10-20 packages), for which we construct a graph, parse the feedstocks, execute migrations, etc?

beckermr:
Yep. That's the way to do them. We've been thinking of doing that for a while but it's a lot of work and covering all cases the bot encounters is not trivial.

h-vetinari:
> We should support both `_` and `-` in both names in the feedstock data and the migrator data.

Migrator side is fine, but not sure what you mean by supporting it in feedstock data. The parser doesn't collect the arches from the meta.yaml directly and uses just one style - AFAICT it needs to use underscores because it constructs keys from it like `sub_graph[f"{plat_arch_name}_meta_yaml"]`

beckermr:
I mean the migrator side only. Sorry for being confusing.

PR #1715, opened 2023-06-15
Slight formatting tweaks to YAML migrator's PR body text

jakirkham:
As `super().pr_body(...)` adds its own text after the `additional_body` constructed by this migrator, make sure to have an `<hr>` right before that text so that the sections are not blended together.

Also quotes the `commit_body` text to make sure it stands out as separate from other text.


jakirkham:
Thanks Matt! 🙏

PR #1722, opened 2023-06-29
BUG skip archived nodes and nodes without pr_info in refresh

beckermr:
cc @jakirkham


jakirkham:
Ah ok. This makes sense. Thanks Matt! 🙏

Sorry had to step AFK for a bit

Issue #1784, opened 2023-08-05
Missing version updates for `mpich`


jakirkham:
It looks like bot version updates are being missed by `mpich`. Raising here to track

xref: https://github.com/conda-forge/mpich-feedstock/pull/81#issuecomment-1666461929

cc @dalcinl

beckermr:
Ouch ok.

jakirkham:
Looking at the version logic here, there seems to be some work needed to go between version and URL. For example [here's the logic]( https://github.com/conda-forge/mpich-feedstock/blob/3c9a0ff0d9800459a2e164ea1504b4bdbf901237/recipe/meta.yaml#L4-L8 ) to convert the version to something that can be used in the URL

```yaml
{% if version.rpartition(".")[2] == "0" %}
{% set version_url = version.rpartition(".")[0] %}
{% else %}
{% set version_url = version %}
{% endif %}
```

Then the [URL looks like this]( https://github.com/conda-forge/mpich-feedstock/blob/3c9a0ff0d9800459a2e164ea1504b4bdbf901237/recipe/meta.yaml#L22 ):

```yaml
  url: https://www.mpich.org/static/downloads/{{ version_url }}/mpich-{{ version_url }}.tar.gz
```

Maybe there is a better way to construct this from the bot's perspective? For example what if the `version_url` was set and `version` was constructed from that?

dalcinl:
@jakirkham That logic in the recipe was suggested to me as a possible way to workaround the issue. MPICH does not really follow SemVer version number scheme, so when they change the minor version `Y`, the new version number is `X.Y`, not `X.Y.0`, therefore the URL has to use `X.Y`, not `X.Y.0`. 

I first experienced the bot update issues when MPICH released version `4.1`. Then I added the logic you mentioned. I manually updated to `4.1.1`, but now the bot did not trigger the update to `4.1.2`. Do you believe this is because we have `{{version_url}}` instead of `{{version}}` in the template?

@hzhou Why are you guys not adding the  `.0`  patch prefix to minor releases to get `X.Y.0`? That simple change would simplify a lot of special-handling in build recipes, deploy scripts, etc. Most other important projects I'm aware of follow this rule, which by the way is simply what [SemVer](https://semver.org/) mandates, and to the best of my knowledge MPICH adheres, except for this minor detail about version number scheme.

beckermr:
Hmmmm. Let's write some tests and see what we can do to fix this. Sorry for the trouble.

PR #1785, opened 2023-08-07
TST add mpich test case

beckermr:
closes https://github.com/regro/cf-scripts/issues/1784

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,19 +26,6 @@ jobs:
           environment-file: environment.yml
           create-args: >-
             python==${{ vars.PYVER }}
-            pip
-            pytest
-            codecov
-            requests-mock
-            doctr
-            sphinx
-            pre-commit
-            pip
-            flaky
-            pytest-xdist
-            pytest-cov
-            setuptools_scm>=7
-            python-build
           init-shell: bash
           cache-environment: true
           environment-name: cf-scripts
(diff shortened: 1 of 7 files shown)


beckermr:
@jakirkham @dalcinl IDK if I gave bad advice before or not. If I did, my apologies. The syntax

```
{% set version_url = version if version[-2:] != ".0" else version[:-2] %}
```

appears to work. I think the fix is to use an inline if statement.

I added tests to the bot to make sure it continues to work.

If the logic appears sound, then let's make the change in the mpich feedstock and merge this PR.

If this is what mpich had before the last change, then we'll need ot dig more.

dalcinl:
> If I did, my apologies.

Oh, no, it was not you, and it is not a big deal anyway, corner cases happen all the time.

> If the logic appears sound,

It does. I'll submit a fix PR in the mpich feedstock and ping you.

beckermr:
On the logic, you may need to handle the case of "5.0" vs "5.0.0". I think the current snippet would strip "5.0" down to "5".

dalcinl:
@beckermr Maybe we are not in the same page. Let me try to explain myself better. If nothing change, at some point MPICH will release the next minor release `4.2`, and the version number in the MPICH side will be `4.2` WITHOUT a trailing `.0` as SemVer would require. However, in the recipe, I want the version to follow SemVer, that is, I want the recipe version to say `version = "4.2.0"`.  Therefore, I must to use the `version_url=4.2` hack to be able to download the release tarball. But will the bot actually figure out that there is a new release? I'm not sure how all this works.

beckermr:
ohhhh nvm you are correct!

PR #1787, opened 2023-08-08
TST add test for random fraction to keep

beckermr:
xref: https://github.com/conda-forge/aws-sdk-cpp-feedstock/issues/662

This test confirms the current code works if it finds the version. It should have shipped 1.11.132 last week but did not. Thus I think this means that the bot did not find the version. I will check that next


beckermr:
cc @h-vetinari

PR #1886, opened 2023-10-10
format not solvable error messages

jaimergp:
Comes from https://matrix.to/#/!mPxkHLLyiXutAGtuaf:gitter.im/$0AVaD6APc2TPw1OsZoEPH76C86zjexoe1tV0gTTssxQ?via=gitter.im&via=matrix.org

The current output is a bit unreadable:

image

This hopefully improves it a bit.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -371,9 +371,11 @@ def run(
                 <summary>
                     not solvable (<a href="{ci_job_url}">bot CI job</a>) @ {base_branch}
                 </summary>
+                <div align="left">
                 <pre>
                 {'</pre><pre>'.join(sorted(set(errors)))}
                 </pre>
+                </div>
                 </details>
                 """
             ).strip()
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -367,10 +367,8 @@ def run(
             ci_job_url = os.getenv("CIRCLE_BUILD_URL", "#")
             _solver_err_str = dedent(
                 f"""
+                not solvable (<a href="{ci_job_url}">bot CI job</a>) @ {base_branch}
                 <details>
-                <summary>
-                    not solvable (<a href="{ci_job_url}">bot CI job</a>) @ {base_branch}
-                </summary>
                 <div align="left">
                 <pre>
                 {'</pre><pre>'.join(sorted(set(errors)))}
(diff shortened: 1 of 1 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -363,11 +364,18 @@ def run(
...
         if not solvable:
-            _solver_err_str = "not solvable ({}): {}: {}".format(
-                ('<a href="' + os.getenv("CIRCLE_BUILD_URL", "") + '">bot CI job</a>'),
-                base_branch,
-                sorted(set(errors)),
-            )
+            ci_job_url = os.getenv("CIRCLE_BUILD_URL", "#")
+            _solver_err_str = dedent(
+                f"""
+                <details>
```
beckermr, line 370: Missing the closing details tag?

jaimergp: Good catch

jaimergp:
Currently trying to reproduce by hand on the HTML canvas. Wait before merge.

jaimergp:
This is more or less what it should look like, to the best of my manual HTML edition abilities:

Collapsed:

image


Revealed:

image


I can't use `<summary>` because for some reason the triangle toggle is missing 🤷 I guess it's been overridden by some CSS rule in the site, but current approach works well enough.

beckermr:
Where does the newline before the details tag come from? Can we remove it?

beckermr:
Ohh nvm this might not be in the final thing. LGTM!

beckermr:
Ideally we should do this formatting in the front-end and not here. That is my bug and I will try to fix it at some point.

jaimergp:
> Ideally we should do this formatting in the front-end and not here

Agreed, something more structured would be better (e.g. a platform-to-error dict instead of the list of str) and would allow for a nicer collection of collapsible blocks. I'm not happy with the align=left stuff either here, we can move that to the `status` site if you want.

jaimergp:
Ideally we would have something like:

```diff
- vpri["new_version_errors"][_new_ver] = _solver_err_str
+ vpri["new_version_errors"][_new_ver] = {
+     "error_type": "not solvable",
+     "job_url": os.getenv("CIRCLE_BUILD_URL"),
+     "base_branch": base_branch,
+     "messages": {
+          "platform_identifier": "string of problems",
+          ...
+     },
+ }
```

jaimergp:
> Where does the newline before the details tag come from?

I think that's the default `details` block formatting. We can't customize the title with a `summary` because the style has been overridden in the frontend somehow.

beckermr:
Yolo.

PR #1978, opened 2023-12-21
REF back to conda

beckermr:
Going back to conda w/ libmamba solver.

### DIFF
--- a/.github/workflows/bot-bot.yml
+++ b/.github/workflows/bot-bot.yml
@@ -28,28 +28,18 @@ jobs:
         run: |
           free
 
-      - uses: mamba-org/setup-micromamba@v1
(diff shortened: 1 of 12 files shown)


xhochy:
What's the reason for this? I currently prefer `setup-micromamba` everywhere due to its better caching mechanism.

beckermr:
Commands like `mamba update --all` and `micromamba update --all` are currently broken. See this comment: https://github.com/mamba-org/mamba/issues/3033#issuecomment-1842508799. They use the local env index information ahead of the downloaded one and so repo data patches to fix incompatibility issues are ignored. This caused us a lot of pain last night and I don't want to deal with it again.

PR #1991, opened 2024-01-10
update_sources: ignore preview candidate (pc) versions

igorsafo:
This is a fix for #1550
#1441 would be really great to fix this issue for everybody in long term.

Here is branch naming convention in oneDNN: https://github.com/oneapi-src/oneDNN/issues/1779

### DIFF
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -159,6 +159,7 @@ class VersionFromFeed(AbstractSource):
         "test",
         "pre",
         "git",
+        "pc",
     ]
 
     def get_version(self, url) -> Optional[str]:


igorsafo:
@beckermr Thanks for promoting the change! Is there a way to add the missing/latest releases? oneDNN already released `rls-v3.3.4`, but due to `rls-v3.3-pc` there are no updates on conda after `rls-v3.3`

beckermr:
You'll need to adjust the latest version in the bot metadata.

igorsafo:
> You'll need to adjust the latest version in the bot metadata.

Do I understand correctly that I should change this file https://github.com/regro/cf-graph-countyfair/blob/master/versions/8/f/1/9/b/onednn.json to something like  `"new_version": "3.3"` and let the bot to pickup the latest version?

PR #2132, opened 2024-02-02
Use Consistent Lowercase Logger

ytausch:


xhochy (approved):

ytausch:
I re-ran the tests without changing anything because there was an upstream issue with one test

PR #2229, opened 2024-03-05
REF use local import to pkg mappings

beckermr:
This PR changes the code to use the import to package mappings generated by the bot. With this change, we can turn off libcfgraph and deprecate it.

### DIFF
--- a/conda_forge_tick/pypi_name_mapping.py
+++ b/conda_forge_tick/pypi_name_mapping.py
@@ -8,6 +8,7 @@ Builds and maintains mapping of pypi-names to conda-forge names
 import functools
 import json
 import math
+import os
 import pathlib
 import traceback
 from collections import Counter, defaultdict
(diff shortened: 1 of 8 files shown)


conda_forge_tick/pypi_name_mapping.py
```diff
@@ -302,12 +309,17 @@ def determine_best_matches_for_pypi_import(
     # TODO: filter out archived feedstocks?
 
     try:
-        clobberers = loads(
-            requests.get(
-                "https://raw.githubusercontent.com/regro/libcfgraph/master/"
-                "clobbering_pkgs.json",
-            ).text,
-        )
+        clobberers_pth = "import_to_pkg_maps/import_to_pkg_maps_clobbering_pkgs.json"
```
beckermr, line 312: Need to use var from module.

beckermr:
@jaimergp @mariusvniekerk Once this PR is merged, we can turn off libcfgraph in favor of streaming from anaconda.org + the OCI stuff. libcfgraph is dead now anyways and has been for months, so I think we should go that route. The next item is to change the conda-forge-metadata package to deprecate libcfgraph and pull import to package maps from this repo.

PR #2239, opened 2024-03-06
Add Pydantic Data Model to Document Node Attributes Data Structure

ytausch:
Currently, there is no documentation about the data model of the cf-graph that this bot implicitly uses.

This is the first step of improving this:
- add a Pydantic model for the node attributes
- add a CI check that validates the model against the current cf-graph state
- the above step lead to the discovery of invalid feedstocks and node attributes files, which are listed in `test_validate.py`.
- README updates
- add PR template encouraging contributors to update the Pydantic model

Documenting the data model also included discovering some undocumented features of the `meta.yaml` file. I submitted corresponding upstream docs updates (https://github.com/conda/conda-build/pull/5214, https://github.com/conda/conda-build/pull/5215, https://github.com/conda/conda-build/pull/5217, https://github.com/conda/conda-build/pull/5218)

PR #2265, opened 2024-03-14
Add Pydantic Model for `pr_info`, `pr_json`, and `versions`

ytausch:
Extending #2239, this PR adds documenting Pydantic models for the `pr_info`, `pr_json`, and `versions` JSON collections.

#2239 should be merged first.


conda_forge_tick/auto_tick.py
```diff
@@ -121,15 +121,6 @@
 MAX_SOLVER_ATTEMPTS = 50
 CHECK_SOLVABLE_TIMEOUT = 90  # 90 days
 
-BOT_RERUN_LABEL = {
```
beckermr, line 124: We cannot remove this. It is used by the bot code below.

beckermr: Oh woops wrong file. Sorry!

conda_forge_tick/auto_tick.py
```diff
@@ -609,8 +599,8 @@ def add_rebuild_migration_yaml(
     excluded_feedstocks: MutableSet[str],
     exclude_pinned_pkgs: bool,
     migration_yaml: str,
-    config: dict = {},
-    migration_name: str = "",
+    config: dict,
+    migration_name: str,
```
beckermr, line 603: Please don't change defaults on functions. The bot's code is not well tested and these functional changes are risking a breakage. Let's stick to only documenting types for these PRs.

conda_forge_tick/git_utils.py
```diff
@@ -405,10 +401,7 @@ def _munge_dict(dest, src, keys):
 
 
 def lazy_update_pr_json(
-    pr_json: Union[Dict, LazyJson],
-    ctx: GithubContext,
-    force: bool = False,
-    trim: bool = True,
```
beckermr, line 411: Please back out the changes to the keyword here.

conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -146,8 +146,7 @@ def __init__(
             ignored_deps_per_node=ignored_deps_per_node,
         )
         self.yaml_contents = yaml_contents
-        assert isinstance(name, str)
```
beckermr, line 149: Why was this assert removed?

beckermr (changes requested):
Please remove all functional changes from this PR. We are more than happy to accept them in other PRs, but it makes it hard to read the diff. Let's only do typing and pydantic models here.

conda_forge_tick/auto_tick.py
ytausch: Using an empty dictionary as default value is an obvious error here. See https://stackoverflow.com/a/26320938

I don't see how this change "risks a breakage". `add_rebuild_migration_yaml` is used exactly once, that is in `migration_factory` below. There, `config` and `migration_name` is set correctly.

I can create a separate PR for this little fix if you want but I usually had good experiences with a "you can always include little fixes to save time" policy.

conda_forge_tick/auto_tick.py
```diff
@@ -689,9 +679,9 @@ def add_rebuild_migration_yaml(
     cycles = list(nx.simple_cycles(total_graph))
     migrator = MigrationYaml(
         migration_yaml,
+        name=migration_name,
         graph=total_graph,
         pr_limit=pr_limit,
-        name=migration_name,
         top_level=top_level,
```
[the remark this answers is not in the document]
ytausch: The argument ordering of the `MigrationYaml` constructor changed. This is why I changed the ordering here as well to ensure consistency.

conda_forge_tick/git_utils.py
ytausch: Why? `trim` is never set to anything other than the default value `True`. If trim is set to `False`, the code below crashes.

conda_forge_tick/migrators/migration_yaml.py
ytausch: The type annotation of `name` is `Optional[str]`, so it doesn't make a lot of sense to assert a string type IMO.
On top of that, I'd suggest sticking to type annotations for hinting types to developers.
But thanks for pointing out, I think actually the correct solution is to change the annotation to `str` - the name ultimately comes from `os.path.splittext` which returns a string. I will change that - but I don't see why we need the assertion then.

ytausch: fixed in 4aecb95e

conda_forge_tick/git_utils.py
beckermr: This pr mixes typing changes with functional ones. I can't tell if things will break. I'm all for the cleanup but I need it in smaller chunks so that it can be reviewed properly.

conda_forge_tick/auto_tick.py
ytausch: Moved to #2273

conda_forge_tick/migrators/migration_yaml.py
beckermr: I don't recall when or why the assert was added, but I assume it was for a reason. Unless we can figure that out and ensure it is safe to remove, we should keep it.

conda_forge_tick/auto_tick.py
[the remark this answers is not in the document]
ytausch: Moved to #2273

conda_forge_tick/git_utils.py
ytausch: Moved to #2273.

conda_forge_tick/migrators/migration_yaml.py
ytausch: Moved to #2273.

ytausch:
@beckermr
It's been a while since I last worked on this - I don't see any functional changes anymore and I also remember opening a separate PR for them, feel free to tell me if there are any left

Also, any clue why pydantic is not found in the CI? It's part of the environment

beckermr:
you need to regenerate the lock file I think.

conda_forge_tick/models/common.py
```diff
@@ -122,14 +146,38 @@ def none_to_empty_dict(value: T | None) -> T | dict[Never, Never]:
...
 
 
-class LazyJsonReference(StrictBaseModel):
+VersionString = Annotated[str, Field(..., pattern=r"^\d+(\.\d+){0,2}$")]
+"""
+A string that matches version numbers in one of the following formats:
+- `X`
+- `X.Y`
+- `X.Y.Z`
+"""
```
beckermr, line 155: This regex is too restrictive in terms of allowed version strings.

conda_forge_tick/models/pr_info.py
```diff
@@ -0,0 +1,342 @@
...
+    branch: str | None = None
+    """
+    The branch that was migrated. Only set if not equal to "master" or "main", which seems a bit questionable.
+    """
+
+    # noinspection PyNestedDecorators
+    @field_validator("branch")
+    @classmethod
+    def check_branch(cls, value: str) -> str:
+        if value in ("master", "main"):
```
beckermr, line 180: We apply ABI migrations to branches other than main or master. Why do we restrict those names here?

beckermr (commented):
Some questions.

conda_forge_tick/models/common.py
ytausch: Currently, this field is only used for the field `smithy_version`. Is there any other schema that conda-smithy follows for its version numbers?

conda_forge_tick/models/pr_info.py
ytausch: I think you misunderstood. The `pr_info` field cannot be `master` or `main`, not the other way around.

If I'm not mistaken, this refers to this code snippet:
https://github.com/regro/cf-scripts/blob/f0796ef8db6379164454cf766df74f608f37f5c9/conda_forge_tick/migrators/core.py#L377-L379

ytausch:
Why does regenerating the lock file with `conda-lock` change everything about the lock file (including key ordering)? This seems a bit weird. Maybe the current lock file was generated with an outdated version of coda-lock? I used 2.5.6.

beckermr:
Yeah it is annoying. It turns out conda-lock itself doesn't produce lock files with nice diffs. I ended up writing some custom code in 

https://github.com/regro/cf-scripts/blob/master/autotick-bot/relock_me.py

that produces stable, sorted lock files that have much smaller diffs. 

I'm ok merging the changes here, but we could run the latest lock file through the function in that script to make them cleaner. 

I should upstream that bit of code to conda lock, but have not had the time. 

If you want to do that, great, but no pressure from me. I know you are busy.

beckermr:
Closing and reopening to cycle the ci

conda_forge_tick/models/pr_info.py
beckermr: Ah yes. That is for backwards compat from before we allowed migrations for other branches. Resolving this one and thank you!

conda_forge_tick/models/common.py
beckermr: Not right now yes but we could change smithy in the future. Restricting to just smithy's current semver scheme adds extra coupling between this code and smithy. 

If we are going to go to the trouble of validating that the version string from conda smithy is correct, then we should definitely use a regex that covers all of the possible allowed versions /. Schemes allowed by conda.  

On the other hand, validating that smithy has a correctly formatted version is adding several lines of code to catch an error that is unlikely to ever happen. From that point of view, I'd be just as happy to remove this bit of schema validation.

ytausch:
I think the CI is not working because of the lock file conflicts.

beckermr:
pre-commit.ci autofix

beckermr:
@ytausch one more issue to be resolved on smithy version validation and then this one is good to go.

conda_forge_tick/models/common.py
beckermr: Ah I see. Thanks for finding this!

I'd check the conda source code for a version regex or maybe google around a bit.

ytausch: @beckermr I decided to use `VersionOrder` because there seems to be a lot of custom logic involved. Think it's ne cleanest solution.

Should be fixed with this.

beckermr (approved):

PR #2267, opened 2024-03-17
ENH use conda-lock

beckermr:
This PR works just on conda-lock.


.github/workflows/tests.yml
```diff
@@ -9,32 +9,33 @@ on:
...
-          miniforge-version: latest
-          environment-file: environment.yml
-          python-version: "${{ vars.PYVER }}"
-          activate-environment: cf-scripts
+          # environment-file: environment.yml
+          environment-name: cf-scripts
           condarc-file: autotick-bot/condarc
+          create-args: >-
+            conda
+            conda-lock
```
xhochy, line 35: With locking, I always have all dependencies in the environment.yml as otherwise further install steps lead to spurious updates again.

PR #2273, opened 2024-03-18
Small Fixes

ytausch:
This fixes some minor issues that were discovered while working on #2239. There are also discussions about some of the things.

### DIFF
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -122,8 +122,8 @@ class MigrationYaml(GraphMigrator):
     def __init__(
         self,
         yaml_contents: str,
+        name: str,
         graph: nx.DiGraph = None,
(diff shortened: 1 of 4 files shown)


conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -146,8 +146,7 @@ def __init__(
             ignored_deps_per_node=ignored_deps_per_node,
         )
         self.yaml_contents = yaml_contents
-        assert isinstance(name, str)
```
beckermr, line 149: Let's put this back for now.

beckermr (changes requested):
One request and then we're good to go! Thank you!

conda_forge_tick/migrators/migration_yaml.py
ytausch: Done.

beckermr (approved):

Issue #2272, opened 2024-03-18
Bot Crashes with Meaningless Error Message for Unsupported `bot.inspection` Value, wrong type annotation


ytausch:
`update_deps.get_updates_and_hints` has the following issues:

1. ~~The type annotation for `update_deps` is wrong, it can also be `Literal[False]`~~ (I propose it not to be "disabled" instead of False in the referenced PR below, which is better IMO)
2. If `update_deps` has an unsupported value, the bot crashes with a meaningless error message. Note that `dep_comparison, hint` is referenced before assignment in this case.
3. The function should [return early](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement).

https://github.com/regro/cf-scripts/blob/bdc3ec20693cadeb3c60a401981ae3d50a22a078/conda_forge_tick/update_deps.py#L171-L238

Reference: https://github.com/conda-forge/conda-smithy/pull/1865

beckermr:
The bot doesn't crash here. 

We catch the exception 

https://github.com/regro/cf-scripts/blob/bdc3ec20693cadeb3c60a401981ae3d50a22a078/conda_forge_tick/migrators/dep_updates.py#L6

or we have a branch before that call

https://github.com/regro/cf-scripts/blob/bdc3ec20693cadeb3c60a401981ae3d50a22a078/conda_forge_tick/migrators/version.py#L16

I am not in favor of changing the current use of `False` because I don't personally want to go around fixing the 11 feedstocks that use it, changing the documentation, etc.

beckermr:
Great! I would not characterize the current setup as purposeful LOL, though for sure we explicitly test for `False` in the Version migrator. :)

Here is a list of steps (in the order to do them) to make this change:

- [ ] Submit a PR to the bot code base that will accept both `False` and `disabled` as values to turn off the dependency checking and change the message the bot leaves with PRs telling folks how to turn it off to say `disabled` instead of `False`. Make sure this PR keeps the `try...except` handling because the dependency update codes fail for reasons outside the bot's control and the bot needs to continue even if those codes fail.
- [ ] Submit PRs to the 11 feedstocks changing the value from `False` to `disabled`. Per my comments on the other PR you made, it'd be best to not rerender the feedstock but instead just submit the PR with `[ci skip]` in the commit message.
- [ ] Submit a PR updating the docs in conda-smithy (You may have already done part of this, but it needs to be separated out so it can be merged.)
[...]

PR #2274, opened 2024-03-18
Support `disabled` bot inspection

ytausch:
see #2272

### DIFF
--- a/conda_forge_tick/migrators/dep_updates.py
+++ b/conda_forge_tick/migrators/dep_updates.py
@@ -48,9 +48,14 @@ class DependencyUpdateMigrator(MiniMigrator):
             )
         except (BaseException, Exception):
             logger.warning("Dep update failed!", exc_info=True)
-        else:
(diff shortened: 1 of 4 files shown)


ytausch:
@beckermr You mentioned something about changing user instructions that say something about `bot.inspection = False` but I could not find them.

beckermr:
Yep here: https://github.com/ytausch/cf-scripts/blob/disabled-inspection/conda_forge_tick/update_deps.py#L433

ytausch:
I just found out that the relevant code is (very unfortunately) duplicated here:
https://github.com/regro/cf-scripts/blob/ac2187a6a4166ddc23380686812dc43ed2e01d5d/conda_forge_tick/migrators/dep_updates.py#L33-L56

This code duplication should be removed - and also here, `get_dep_updates_and_hints` can be called with `False` or `disabled` which is incorrect.

beckermr (approved):
Thank you!

PR #2288, opened 2024-03-22
ENH relock on the regs

beckermr:
cc @jaimergp

### DIFF
--- a/.github/workflows/relock.yml
+++ b/.github/workflows/relock.yml
@@ -34,4 +34,3 @@ jobs:
           git add conda-lock.yml
           git commit --allow-empty -m "[ci skip] relock deps w/ conda-lock"
           git push
-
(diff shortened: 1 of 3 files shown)


.github/workflows/relock.yml
```diff
@@ -0,0 +1,44 @@
+name: relock
+
+on:
+  workflow_dispatch: null
+  schedule:
+    - cron:  '15 * * * *'
```
jaimergp, line 6: Isn't this too often? Every 15 mins? 🤔 
I'd say once or twice a week is more than enough.

beckermr: I have a ton of changes coming through to make this a lot smarter. Sorry to bump you early.

jaimergp: Ah ok, no probs!

beckermr: OK I pushed to main for testing. See relock_me.py in the autotick bot directory.

PR #2317, opened 2024-03-26
Fix corner case in stdlib piggyback

h-vetinari:
First follow-up to #2135 after looking at a failure in the rdkit feedstock, which uses a conda-build corner case that I hadn't thought of (and really dislike...)


conda_forge_tick/migrators/cstdlib.py
```diff
@@ -200,6 +201,11 @@ def migrate(self, recipe_dir, attrs, **kwargs):
 
             sections = _slice_into_output_sections(lines, attrs)
             for name, section in sections.items():
+                if name == feedstock_name:
```
h-vetinari, line 204:
```suggestion
                if (name == feedstock_name) and len(section) == 1:
```

PR #2325, opened 2024-03-26
Label relock PRs

jaimergp:
Auto-label the relock PRs so they can be easily omitted in PR searches (same label as Dependabot ones).


beckermr:
Thank you!

beckermr:
I added a bit more code to reduce some of the noise here.

.github/workflows/relock.yml
```diff
@@ -43,13 +43,15 @@ jobs:
           title: relock w/ conda-lock
           body: >
```
beckermr, line 44:
```suggestion
          body: |
```

PR #2372, opened 2024-04-07
switch on stdlib-piggyback for all migrations

h-vetinari:
Now that we've fixed the initial set of [issues](https://github.com/regro/cf-scripts/issues/2328) uncovered by running the stdlib migrator (#2135) for the boost 1.84 migration, it should be time to roll it out to all migrations.

CC @beckermr @isuruf

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -689,10 +689,9 @@ def add_rebuild_migration_yaml(
         piggy_back_migrations.append(JpegTurboMigrator())
     if migration_name == "boost_cpp_to_libboost":
         piggy_back_migrations.append(LibboostMigrator())
-    if migration_name == "boost1840":
-        # testing phase: only a single migration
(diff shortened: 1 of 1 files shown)


beckermr:
Should we add to the version migrator too?

h-vetinari:
> Should we add to the version migrator too?

That's a great idea, though I'm kinda on the fence about it. One current failure mode for the piggyback is the output-names are templated (e.g. `name: {{ name }}`), as we then cannot reasonably slice the recipe into sections. This will cause some small amount of failures for the migrators (kinda unavoidable, though I tend to get rid of such fake variables once I touch a feedstock), but I think we should not encumber the version bot with that? 🤔 

Or maybe we can loosen the safety net of the piggyback there...?

h-vetinari:
In any case, I would defer the question about adding it to the version migrator for now; we can reconsider later, if/once we've found a way to avoid spurious failures.

beckermr:
SGMT and thank you!

PR #2374, opened 2024-04-08
STY enable ruff with fixes

beckermr:
This PR enables ruff formatting and puts in the associated fixes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,3 +38,13 @@ line-length = 88
 
 [tool.isort]
 profile = "black"
+
+[tool.ruff]
+extend-exclude = ["conda_forge_tick/migrators/disabled/legacy.py"]
+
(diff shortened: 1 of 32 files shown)


pyproject.toml
```diff
@@ -38,3 +38,13 @@ line-length = 88
 
 [tool.isort]
 profile = "black"
+
+[tool.ruff]
+extend-exclude = ["conda_forge_tick/migrators/disabled/legacy.py"]
+
+[tool.ruff.lint]
+select = ["E", "F", "W"]
```
jaimergp, line 46: Cool! [This is what `conda/conda` uses now](https://github.com/conda/conda/blob/298ca49ff062e93622c951bcba04390d48a0e64b/pyproject.toml#L111-L136) and I've found that config super balanced in what it fixes vs how noisy it gets.

PR #2375, opened 2024-04-08
ENH add option to turn off local cache and document lazy JSON

beckermr:
This PR does as it says. It is a small one so I added in all of the items needed at once.


jaimergp:
Awesome, thanks for the docs!

PR #2377, opened 2024-04-08
REF use custom entrypoint and separate build workflow

beckermr:
Some more docker changes.


beckermr:
more docker

cc @jaimergp

PR #2387, opened 2024-04-09
ENH update docker image and settings

beckermr:
This PR

 - updates the docker image build to be simpler
 - adds extra settings for running the images
 - centralizes test infrastructure related to selecting the right image


tests/test_container_tasks.py
```diff
@@ -63,8 +62,6 @@ def test_container_tasks_version_json():
 
 def test_get_latest_version_containerized(monkeypatch):
```
beckermr, line 63:
```suggestion
def test_get_latest_version_containerized():
```

PR #2389, opened 2024-04-10
ENH reentrant container use in version update, better error handling

beckermr:
This PR has

 - better error handling for getting latest versions
 - logging for getting latest versions
 - reentrant container use
 - documentation for env vars added and used

### DIFF
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -383,7 +383,8 @@ def url_exists(url: str, timeout=2) -> bool:
                 stderr=subprocess.STDOUT,
                 timeout=timeout,
             )
-        except Exception:
+        except Exception as e:
+            logger.debug("url_exists wget exception: %s", repr(e))
             return False
         # For FTP servers an exception is not thrown
         if "No such file" in output.decode("utf-8"):
@@ -399,7 +400,8 @@ def url_exists(url: str, timeout=2) -> bool:
                 capture_output=True,
                 check=True,
             )
-        except subprocess.CalledProcessError:
+        except subprocess.CalledProcessError as e:
+            logger.debug("url_exists curl exception: %s", repr(e))
             return False
 
         return True
--- a/conda_forge_tick/update_upstream_versions.py
+++ b/conda_forge_tick/update_upstream_versions.py
@@ -1,11 +1,8 @@
 import functools
 import hashlib
-import json
 import logging
 import os
-import pprint
 import random
-import subprocess
 import time
 from concurrent.futures import as_completed
 from typing import (
@@ -40,12 +37,7 @@ from .update_sources import (
     RawURL,
     ROSDistro,
 )
-from .utils import (
-    get_default_container_name,
-    get_default_container_run_args,
-    get_keys_default,
-    load_existing_graph,
-)
+from .utils import get_keys_default, load_existing_graph, run_container_task
 
 T = TypeVar("T")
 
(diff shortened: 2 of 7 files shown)


beckermr:
cc @jaimergp

beckermr:
I am going to merge but feel free to add comments later. The same goes for any of my recent PRs from yesterday:

 - #2387 
 - #2386 
 - #2385 
 - #2384 
 - #2377 
 - #2375

conda_forge_tick/update_sources.py
```diff
@@ -383,7 +383,8 @@ def url_exists(url: str, timeout=2) -> bool:
                 stderr=subprocess.STDOUT,
                 timeout=timeout,
             )
-        except Exception:
+        except Exception as e:
+            logger.debug("url_exists wget exception: %s", repr(e))
```
jaimergp, line 387: I think you can do `%r` and save the `repr(e)`. IOW, this is equivalent:

```
logger.debug("url_exists wget exception: %r", e)
```

conda_forge_tick/update_upstream_versions.py
```diff
@@ -71,16 +63,27 @@ def ignore_version(attrs: Mapping[str, Any], version: str) -> bool:
...
-    :param sources: the version sources to use (sources can be excluded by the package but not added)
+
+    Parameters
+    ----------
+    name : str
+        The name of the feedstock.
+    attrs : Mapping[str, Any]
+        The node attributes of the feedstock.
+    sources : Iterable[AbstractSource]
+        The version sources to use.
```
jaimergp, line 81: Personally, if the signature is already typed, I'm not adding that info in the docstring.

beckermr: hahahaha - this was all github copilot! I let it add things at ~random and double check they don't look crazy. :p

conda_forge_tick/update_upstream_versions.py
```diff
@@ -170,34 +173,80 @@ def get_latest_version(
...
+    use_container : bool, optional
+        Whether to use a container to run the version parsing.
+        If None, the function will use a container if the environment
+        variable `CF_TICK_IN_CONTAINER` is 'false'. This feature can be
```
jaimergp, line 234: Is this accurate? My first impression that use_container would be True if CF_TICK_IN_CONTAINER is truthy. But I guess this env var means that "we are already in a container, do not start another one", right?

conda_forge_tick/utils.py
```diff
@@ -128,6 +131,77 @@ def get_default_container_run_args(tmpfs_size_mb: int = 10):
...
 
 
+def run_container_task(name, args, json_loads=json.loads):
+    """Run a bot task in a container.
+
+    Parameters
+    ----------
+    name : str
+        The name of the task.
+    args : List[str]
```
jaimergp, line 141: Let's add the types to the signature instead. As a general principle, the rule of thumb is "be abstract in the inputs and concrete in the outputs". In this case, that means the signature could have `Iterable[str]` instead of `list[str]`.

By the way, we can use `from __future__ import annotations` and use the builtins as types (list, tuple, dict) directly with square brackets. I think that's available with Py38+

conda_forge_tick/update_upstream_versions.py
beckermr: Yes. This is to prevent docker in docker.

tests/test_container_tasks.py
```diff
@@ -9,64 +8,45 @@
...
-        with LazyJson("node_attrs/conda-smithy.json") as lzj:
-            attrs = copy.deepcopy(lzj.data)
-
-    data = get_latest_version_containerized(
-        "conda-smithy", attrs, all_version_sources()
-    )
-    assert data["new_version"] == conda_smithy.__version__
+    with tempfile.TemporaryDirectory() as tmpdir, pushd(tmpdir):
+        with lazy_json_override_backends(["github"], use_file_cache=False):
+            with LazyJson("node_attrs/conda-smithy.json") as lzj:
```
jaimergp, line 46: If we require Python 3.10 for the bot code, then we can use [parenthesised context managers](https://docs.python.org/3/whatsnew/3.10.html#parenthesized-context-managers) for cleaner blocks:

```
with (
    tempfile.TemporaryDirectory() as tmpdir,
    pushd(tmpdir),
    lazy_json_override_backends(["github"], use_file_cache=False),
    LazyJson("node_attrs/conda-smithy.json") as lzj,
):
```

PR #2396, opened 2024-04-11
ENH address CR for docker

beckermr:
This PR addresses CR comments from @jaimergp for docker.


jaimergp:
Nice! Looks like we are already [requiring python=3.11](https://github.com/regro/cf-scripts/blob/bbe8ed11595aa6c21f92579f96e41d98788a8064/environment.yml#L45), so this should work. I'd still add the [`python-requires` metadata to pyproject.toml](https://packaging.python.org/en/latest/specifications/pyproject-toml/#requires-python) for completeness.

PR #2401, opened 2024-04-11
ENH dedupe meta.yaml aggregated over platforms

beckermr:
This PR deduplicates the meta_yaml dicts aggregated over various platforms. It will cause a lot of turnover in the bot metadata for a while, but should reduce the overall size on disk eventually.

### DIFF
--- a/conda_forge_tick/feedstock_parser.py
+++ b/conda_forge_tick/feedstock_parser.py
@@ -36,6 +36,35 @@ PIN_SEP_PAT = re.compile(r" |>|<|=|\[")
 BOOTSTRAP_MAPPINGS = {}
 
 
+def _dedupe_list_ordered(list_with_dupes):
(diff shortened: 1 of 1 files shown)


conda_forge_tick/feedstock_parser.py
```diff
@@ -36,6 +36,35 @@
...
+    if not all(isinstance(x, str) for x in list_with_dupes):
+        return list_with_dupes
+
+    seen = set()
+    list_without_dupes = []
+    for item in list_with_dupes:
+        if item not in seen:
+            seen.add(item)
+            list_without_dupes.append(item)
+    return list_without_dupes
```
jaimergp, line 52: For order-keeping deduplication, this idiom has been super useful for me:

```
list_without_dupes = list(dict.fromkeys(list_with_dupes))
```

PR #2462, opened 2024-04-18
BUG make sure to skip container tests if docker not there

### LINKED ISSUE #2460: Local tests now need docker
This is probably a by-product of @beckermr's recent work on hardening security (e.g. in [`conda_forge_tick/utils.py`](https://github.com/regro/cf-scripts/blame/master/conda_forge_tick/utils.py)), but is it really necessary that even the testing of a migrator now needs a docker daemon to run?

I was trying to add a new test for a corner case of the stdlib migration, but cannot run the tests locally on my windows machine anymore, which makes this much harder to do.

beckermr:
This PR ensures the test suite runs without docker. 

closes #2460

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,3 +36,25 @@ def set_cf_tick_pytest_envvar():
         del os.environ["CF_TICK_PYTEST"]
     else:
         os.environ["CF_TICK_PYTEST"] = old_ci
+
+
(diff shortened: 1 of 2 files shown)


beckermr:
cc @h-vetinari

h-vetinari:
Cool, thanks a lot for the quick fix! 🙏

PR #2470, opened 2024-04-20
add mini-piggyback migrator for removing `pin_compatible("numpy"...)`

### LINKED ISSUE #2469: Piggyback migrator to drop `{{ pin_compatible("numpy") }}`
As `numpy` already [contains a `run_exports`]( https://github.com/conda-forge/numpy-feedstock/blob/b60684fab0eae8c1d9f51ba30adcd03c8691bf87/recipe/meta.yaml#L29-L30 ), which has been present since `numpy` version `1.22.3` number `1` ( https://github.com/conda-forge/numpy-feedstock/pull/262 ), and conda-forge is using [`numpy` version `1.22` or newer]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/f3a458a625609f36daaa4cad5ddc6a69615a9fab/recipe/conda_build_config.yaml#L701-L706 ), we can just rely on `run_exports` from `numpy` and drop `pin_compatible` from recipes

This is also important to prepare for building with `numpy` version `2.0`, which builds packages with `numpy` version `1.x` and `2.x` support. As such `pin_compatible` would be unnecessarily strict by restricting packages built with it  to `numpy` version `2.x` when `1.x` could have also been supported

As a result (based on discussion at the core meeting), we decided to add a piggyback migrator that would remove `{{ pin_compatible("numpy") }}`. This would help prepare for the NumPy 2 migrator. It also could couple with that migrator to upgrade to NumPy 2 and clean this pinning out at the same time

h-vetinari:
Fixes #2469; goes hand-in-hand with (the migrator filename used in) https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5790

This is a baby version of #2135 / #1668, where we've already extensively tested the used functionality. Strictly speaking, we could even avoid the whole dance with `_slice_into_output_sections` and just directly remove any lines matching the regex pattern for `pin_compatible("numpy",...)` in the meta.yaml. However, I left it in for now in case anyone wants to add some logic that takes into account existing host-/run-dependencies, because then we need to operate on a per-section basis.

### DIFF
--- a/conda_forge_tick/migrators/numpy2.py
+++ b/conda_forge_tick/migrators/numpy2.py
@@ -1,0 +1,47 @@
+import os
+import re
+
+from conda_forge_tick.migrators.core import MiniMigrator
+from conda_forge_tick.migrators.libboost import _replacer, _slice_into_output_sections
+
+# pin_compatible("numpy"...)
+# ^   ^           ^
+# p   c           n
+raw_pat_pcn = r".*\{\{\s*pin_compatible\([\"\']numpy[\"\'].*"
+pat_pcn = re.compile(raw_pat_pcn)
+
+
+def _process_section(name, attrs, lines):
+    """
+    Migrate requirements per section.
+
+    We want to migrate as follows:
+    - remove all occurrences of `{{ pin_compatible("numpy",...) }}`;
+      these will be taken care of henceforth by numpy's run-export
+    """
+    # _replacer take the raw pattern, not the compiled one
+    lines = _replacer(lines, raw_pat_pcn, "")
+    return lines
+
+
+class Numpy2Migrator(MiniMigrator):
+    def filter(self, attrs, not_bad_str_start=""):
+        lines = attrs["raw_meta_yaml"].splitlines()
+        has_pcn = any(pat_pcn.search(line) for line in lines)
+        # filter() returns True if we _don't_ want to migrate
+        return not (has_pcn)
+
+    def migrate(self, recipe_dir, attrs, **kwargs):
+        fname = os.path.join(recipe_dir, "meta.yaml")
+        if os.path.exists(fname):
+            with open(fname) as fp:
+                lines = fp.readlines()
+
+            new_lines = []
+            sections = _slice_into_output_sections(lines, attrs)
+            for name, section in sections.items():
+                # _process_section returns list of lines already
+                new_lines += _process_section(name, attrs, section)
+
+            with open(fname, "w") as fp:
+                fp.write("".join(new_lines))
--- a/conda_forge_tick/migrators/numpy2.py
+++ b/conda_forge_tick/migrators/numpy2.py
@@ -2,26 +2,13 @@ import os
 import re
 
 from conda_forge_tick.migrators.core import MiniMigrator
-from conda_forge_tick.migrators.libboost import _replacer, _slice_into_output_sections
+from conda_forge_tick.migrators.libboost import _replacer
 
 # pin_compatible("numpy"...)
 # ^   ^           ^
 # p   c           n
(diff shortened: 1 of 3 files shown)


beckermr (changes requested):
The `_slice_into_output_sections` can raise an error. This made sense for the stdlib work, but here we could iterate through the lines and do a straight replacement. 

Also, the code will fail if people do unusual but valid things in yaml like an explicit list. I don't think we should handle this last case maybe?

h-vetinari:
> Also, the code will fail if people do unusual but valid things in yaml like an explicit list. I don't think we should handle this last case maybe?

You mean like:
```
run: [{{ pin_compatible("numpy") }}, something_else]
```
?

> The `_slice_into_output_sections` can raise an error. This made sense for the stdlib work, but here we could iterate through the lines and do a straight replacement.

Fine by me.

beckermr:
Yes I mean like this `run: [{{ pin_compatible("numpy") }}, something_else]`.

conda_forge_tick/migrators/numpy2.py
```diff
@@ -0,0 +1,32 @@
...
+        if os.path.exists(fname):
+            with open(fname) as fp:
+                lines = fp.readlines()
+
+            new_lines = []
+            # _replacer take the raw pattern, not the compiled one
+            lines = _replacer(lines, raw_pat_pcn, "")
+
+            with open(fname, "w") as fp:
+                fp.write("".join(new_lines))
```
beckermr, line 32: This looks wrong?

```suggestion
            # _replacer take the raw pattern, not the compiled one
            lines = _replacer(lines, raw_pat_pcn, "")

            with open(fname, "w") as fp:
                fp.write("".join(lines))
```

h-vetinari: Yeah, getting tired. 😑

h-vetinari:
> Yes I mean like this `run: [{{ pin_compatible("numpy") }}, something_else]`.

It's already handled. The regular expression doesn't care what comes before the first `{`

beckermr:
Ah but if _replacer ignores the whole line, then we delete the entire run section?

h-vetinari:
> Ah but if _replacer ignores the whole line, then we delete the entire run section?

It doesn't ignore it, it deletes it. If that ends up creating an empty run-section that's not worth cleaning up automatically IMO - too many corner cases (comments, etc.). We also didn't do that for boost or stdlib either, e.g. they produced
```diff
     run:
-       - boost-cpp
```
or
```diff
     run_constrained:
-       - __osx >={{ MACOSX_DEPLOYMENT_TARGET|default("10.9") }}  # [osx and x86_64]
```
and that was fine?

beckermr:
Yep agreed! I was just checking that I understood the code.

beckermr (approved):

jakirkham:
Thanks Axel and Matt! 🙏

Have we seen this piggyback migrator applied in any PRs? Would be interesting to see how it is doing

h-vetinari:
> Have we seen this piggyback migrator applied in any PRs? Would be interesting to see how it is doing

It explicitly looks for the `numpy2.yaml` from https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5790. As long as that PR is not merged, the piggyback won't ever be applied.

jakirkham:
Interesting am curious is there a particular reason to wait for NumPy 2 to make this change?

As noted in issue ( https://github.com/regro/cf-scripts/issues/2469 ), this change should be ok to make with the current NumPy pinning

Please let me know if this is more to the story

h-vetinari:
> is there a particular reason to wait for NumPy 2 to make this change?

You're right that we _could_ already remove it based on the existing 1.x infrastructure, but it's a numpy-related change - IMO it makes sense to roll it out together with the numpy migrator, rather than just attach it to random migrators or version updates.

PR #2543, opened 2024-05-06
REF make libboost migrator more robust

beckermr:

### DIFF
--- a/conda_forge_tick/migrators/libboost.py
+++ b/conda_forge_tick/migrators/libboost.py
@@ -17,39 +17,51 @@ def _slice_into_output_sections(meta_yaml_lines, attrs):
     The result will always contain a "global" section (== everything
     if there are no other outputs).
     """
-    outputs = attrs["meta_yaml"].get("outputs", [])
-    output_names = [o["name"] for o in outputs]
-    # if there are no outputs, there's only one section
-    if not output_names:
-        return {"global": meta_yaml_lines}
-    # output_names may contain duplicates; remove them but keep order
-    names = []
-    [names := names + [x] for x in output_names if x not in names]
-    num_outputs = len(names)
-    # add dummy for later use & reverse list for easier pop()ing
-    names += ["dummy"]
-    names.reverse()
-    # initialize
-    pos, prev, seek = 0, "global", names.pop()
+    outputs_token_pos = None
+    re_output_start = None
+    re_outputs_token = re.compile(r"^\W*outputs:.*")
+    re_outputs_token_list = re.compile(r"^\W*outputs:\W*\[.*")
+    re_match_block_list = re.compile(r"^\W*-.*")
+
+    pos = 0
+    section_index = -1
     sections = {}
     for i, line in enumerate(meta_yaml_lines):
-        # assumes order of names matches their appearance in meta_yaml,
-        # and that they appear literally (i.e. no {{...}}) and without quotes
-        if f"- name: {seek}" in line:
-            # found the beginning of the next output;
-            # everything until here belongs to the previous one
-            sections[prev] = meta_yaml_lines[pos:i]
-            # update
-            pos, prev, seek = i, seek, names.pop()
-            if seek == "dummy":
-                # reached the last output; it goes until the end of the file
-                sections[prev] = meta_yaml_lines[pos:]
-    if len(sections) != num_outputs + 1:
-        raise RuntimeError("Could not find all output sections in meta.yaml!")
+        if re_outputs_token.match(line) and outputs_token_pos is None:
+            if re_outputs_token_list.match(line):
+                raise RuntimeError(
+                    "List-style outputs not supported for outputs slicing!"
+                )
+
+            outputs_token_pos = i
+            continue
+
+        if (
+            outputs_token_pos is not None
+            and re_output_start is None
+            and i > outputs_token_pos
+            and re_match_block_list.match(line)
+        ):
+            outputs_indent = len(line) - len(line.lstrip())
+            re_output_start = re.compile(r"^" + r" " * outputs_indent + r"-.*")
+
+        if (
+            outputs_token_pos is not None
+            and re_output_start is not None
+            and re_output_start.match(line)
+        ):
+            sections[section_index] = meta_yaml_lines[pos:i]
+            section_index += 1
+            pos = i
+            continue
+
+    if pos < len(meta_yaml_lines):
+        sections[section_index] = meta_yaml_lines[pos:]
+
     return sections
 
 
-def _process_section(name, attrs, lines):
+def _process_section(output_index, attrs, lines):
     """
     Migrate requirements per section.
 
@@ -60,13 +72,22 @@ def _process_section(name, attrs, lines):
       and remove it in run.
     """
     outputs = attrs["meta_yaml"].get("outputs", [])
-    if name == "global":
+    if output_index == -1:
         reqs = attrs["meta_yaml"].get("requirements", {})
     else:
-        filtered = [o for o in outputs if o["name"] == name]
-        if len(filtered) == 0:
-            raise RuntimeError(f"Could not find output {name}!")
-        reqs = filtered[0].get("requirements", {})
+        seen_names = set()
+        unique_outputs = []
+        for output in outputs:
+            _name = output["name"]
+            if _name in seen_names:
+                continue
+            seen_names.add(_name)
+            unique_outputs.append(output)
+
+        try:
+            reqs = unique_outputs[output_index].get("requirements", {})
+        except IndexError:
+            raise RuntimeError(f"Could not find output {output_index}!")
 
     build_req = reqs.get("build", set()) or set()
     host_req = reqs.get("host", set()) or set()
@@ -101,7 +122,7 @@ def _process_section(name, attrs, lines):
         lines = _replacer(lines, p_base + p_selector, "libboost-devel", max_times=1)
         # delete all other occurrences
         lines = _replacer(lines, "boost-cpp", "")
-    elif is_boost_in_host and name == "global" and outputs:
+    elif is_boost_in_host and output_index == -1 and outputs:
         # global build section for multi-output with no run-requirements;
         # safer to use the full library here
         lines = _replacer(lines, p_base + p_selector, "libboost-devel", max_times=1)
@@ -154,9 +175,9 @@ class LibboostMigrator(MiniMigrator):
 
             new_lines = []
             sections = _slice_into_output_sections(lines, attrs)
-            for name, section in sections.items():
+            for output_index, section in sections.items():
                 # _process_section returns list of lines already
-                new_lines += _process_section(name, attrs, section)
+                new_lines += _process_section(output_index, attrs, section)
 
             with open(fname, "w") as fp:
                 fp.write("".join(new_lines))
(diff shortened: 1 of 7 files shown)


beckermr:
@h-vetinari I've been seeing a lot of errors in this migrator so I refactored it to be a bit more robust.

conda_forge_tick/migrators/libboost.py
```diff
@@ -58,6 +58,15 @@ def _slice_into_output_sections(meta_yaml_lines, attrs):
     if pos < len(meta_yaml_lines):
         sections[section_index] = meta_yaml_lines[pos:]
 
+    # output_names may contain duplicates; remove them but keep order
```
beckermr, line 61:
```suggestion
    # double check length here to flag possible weird cases
    # this check will fail incorrectly for outputs with the same name
    # but different build strings.
```

beckermr:
@h-vetinari any comments here?

conda_forge_tick/migrators/libboost.py
```diff
@@ -12,44 +12,106 @@ def _slice_into_output_sections(meta_yaml_lines, attrs):
...
+            and re_output_start.match(line)
+        ):
+            sections[section_index] = meta_yaml_lines[pos:i]
+            section_index += 1
+            pos = i
+            continue
+
+    # the last output possibly needs to be added
+    if pos < len(meta_yaml_lines):
+        sections[section_index] = meta_yaml_lines[pos:]
```
beckermr, line 72:
```suggestion
    # the last output needs to be added
    sections[section_index] = meta_yaml_lines[pos:]
```

beckermr:
> AFAICT this still doesn't solve templated output-names either. I personally find the index formulation ("index to what?") harder to understand than just identifying a slice by output name, but again, I don't yet understand what problems we're solving here.

It definitely solves the templated output names issue. We no longer require the parser to look for ` name: <name of output>`. The output names are never used except to cut things down to a unique set.

beckermr:
> From looking at the tests, I don't see why these would/should have failed before (aside from the API change of course). What failures did you see? Can we add one of the failing recipes as a test?

Failures are in the bot CI jobs where this code is raising and preventing migrations from going through. I will add a test with those.

tests/test_libboost.py
```diff
@@ -60,3 +61,122 @@ def test_boost(feedstock, new_ver, tmpdir):
...
+        "package:",
+        "  name: blah",
+        "  version: 1.0.0",
+        "",
+        "requirements:",
+        "  host:",
+        "    - foo",
+        "",
+        "outputs:",
+        "  # comment",
```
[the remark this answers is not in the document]
beckermr: Writing these tests was such a pain. This comment would have made it a lot easier!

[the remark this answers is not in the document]
beckermr: done!

h-vetinari:
> It definitely solves the templated output names issue

OK sorry, that wasn't clear from the test, which is still using hardcoded names. How about having an explicit test à la:
```
outputs:
    - name: {{ name|lower }}
      requirements: [...]
    - name: {{ "lib" ~ name }}
      requirements: [...]
    - name: {{ name ~ "-py" }}
      requirements: [...]
```
There's a bunch of infrastructure for this in `tests/test_libboost.py` where you can put a before/after `meta.yaml` for a given feedstock.

beckermr:
Yep! I am adding some tests per your suggestion here: https://github.com/regro/cf-scripts/pull/2548

They fail with the changes on master and so are marked as xfail. 

The rucio-client one there is interesting. That feedstock failed the slicing but I cannot see why.

beckermr:
> I personally find the index formulation ("index to what?") harder to understand than just identifying a slice by output name, but again, I don't yet understand what problems we're solving here.

I don't disagree here but getting the templated names out of the recipe itself is problematic for a lot reasons including selectors, jinja2 conditional statements, the fact that things like intervening jinja2 statements can potentially change the actual output name even if the template looks the same (mean two unique outputs can have the same template), and that the values of the templated vars are not always easy to deduce.

Counting outputs that come out are ordered seems simpler and a bit more robust as a heuristic, but also has edge cases around jinja2 conditional statements. :(

h-vetinari:
> I don't disagree here but getting the templated names out of the recipe itself is problematic for a lot reasons including selectors

Yeah sure, once we solve that case the extra complexity is very likely justified.

beckermr:
OK. Templated names now pass. There are some unrelated bugs the additional test cases flushed out and there are some actual bugs in the tests themselves. Will push the fixes later.

beckermr:
Alright @h-vetinari, the new tests show the cases the libboost and stdlib migrators can now handle. This will unblock a several failing PRs each bot run, which is a pretty big effect. Templated names for outputs are not rare it turns out.

conda_forge_tick/migrators/libboost.py
```diff
@@ -12,44 +12,105 @@ def _slice_into_output_sections(meta_yaml_lines, attrs):
...
+            pos = i
+            continue
+
+    # the last output needs to be added
+    sections[section_index] = meta_yaml_lines[pos:]
+
+    # finally, if a block list at the same indent happens after the outputs section ends
+    # we'll have extra outputs that are not real. We remove them
+    # by checking if there is a name key in the dict
+    re_name = re.compile(r"^\s*(-\s+)?name:.*")
```
h-vetinari, line 76: I don't get the effort here. Why don't we just match for `- name: ...` directly, rather than determining the indent level, catching everything at that indent level, and _then_ matching on name?

conda_forge_tick/migrators/libboost.py
```diff
@@ -12,44 +12,105 @@ def _slice_into_output_sections(meta_yaml_lines, attrs):
...
+                )
+
+            outputs_token_pos = i
+            continue
+
+        # next we need to find the indent of the first output list element.
+        # it could be anything from the indent of the `outputs:` line to that number
+        # plus zero or more spaces. Typically, the indent is the same as the `outputs:`
+        # plus one of [0, 2, 4] spaces.
+        # ocne we find the first output list element, we build a regex to match any list
```
h-vetinari, line 46: ocne -> once

(though I'd like to avoid the indentation handling completely - I don't (yet?) see why it's necessary)

h-vetinari (commented):
some questions about the implementation

conda_forge_tick/migrators/libboost.py
beckermr: An output doesn't have to start with the name key as the first item in the dictionary. People do this as a custom but it is not required.

conda_forge_tick/migrators/libboost.py
beckermr: Thanks. I will fix the typo. 

The indentation handling is because yaml allows any consistent indent below a given section. The outputs are a list, is you can have

"""
outputs:
- blah
"""

Or 

"""
outputs:
  - blah
"""

Or w/e

beckermr, line 46:
```suggestion
        # once we find the first output list element, we build a regex to match any list
```

beckermr: Looking for list items at any indent is not robust ofc, so we have to find a list item at the correct indent. Even that is not robust since we don't know when the outputs section ends. Thus we clean that up later by requiring a `name: ` key in the output. This is a heuristic since in principle that string could appear in other parts of the recipe.

beckermr: We could detect the indent of the dict in each output, record it, and then find the right name key at the correct indent.

conda_forge_tick/migrators/libboost.py
h-vetinari: > An output doesn't have to start with the name key as the first item in the dictionary.

Thanks, I hate it. 🙈 

I've _never_ seen that though (in contrast to templated outputs). In fact, I'd _want_ such feedstocks to break, because doing
```
outputs:
  - build:
      skip: true  # [not linux]
    name: ...
```
should just not be supported at all (i.e. we should turn it into a hard linter error) IMO.

I certainly don't think it's worth the implementation effort to accommodate such an extreme corner case.

conda_forge_tick/migrators/libboost.py
h-vetinari: Personally, I just want to not care about the indent at all here, and let the regex portion of `\s*` take care of it.

conda_forge_tick/migrators/libboost.py
beckermr: Yeah right now the libboost and cstdlib migrators are blocking perfectly good version and migration prs from being made because they don't handle templated names. 

We have one of these options

- turn off these minimigrators
- swallow their errors so they don't block PRs
- merge this pr to help fix things
- merge another pr to help fix things
- go around telling people the bot cannot handle templated output names

The implementation effort is fundamentally what is required to do the job at some standard, not what we are willing to give. The current recipe format allows a lot of edge cases and that generates complicated code in the bot. That's just how it is.

conda_forge_tick/migrators/libboost.py
beckermr: That's less robust and leaves edge cases.

beckermr:
@h-vetinari I don't get the resistance to this PR. 

If it makes you feel better, you can defer all further issues with this code or the associated bot aspects of those migrations to me. That way you don't have any extra work associated with code you feel is too complex. 

The current code misses too many edge cases and that has to be fixed.

conda_forge_tick/migrators/libboost.py
h-vetinari: There was a misunderstanding here. I agree we should fix templated names, no question about that (even though I'd like to get away from them wherever possible).

What I meant is that I don't think we should fix "name of output does not appear as first line", because the implementation here has IMO ballooned way further than what it is necessary to support templated outputs - which would mainly be to relax `f"- name: {seek}"` -> `f"- name:` (i.e. not include the package name), and incorporate your array/index logic of course.

If you feel the added generality for catching this extreme corner case[^1] is worth it, I won't stand in the way, but it makes me way less likely to touch that code again because I feel it's a bad trade-off and would have to resist the urge to remedy that.

[^1]: again, not talking about templated names here, but about not relying on having the hyphen before `name:`, i.e. `name:` being the first line per output

beckermr: That's fine. I'll take care of this code from now on. Thanks for the help in getting it started!

h-vetinari:
> @h-vetinari I don't get the resistance to this PR.

The implementation complexity makes a big jump, I'm simply asking if it's worth it in all aspects. That's far from resistance (I said I won't stand in the way), just regular review process.

> The current code misses too many edge cases and that has to be fixed.

You should IMO add tests for all the cases you intend the implementation to cover, otherwise it's too easy for things to regress.

conda_forge_tick/migrators/libboost.py
h-vetinari: We might also want to move this to be an "official" (albeit smith-internal) utility, rather than being tucked away in the boost migrator.

beckermr:
Good point! Both the templated names and outputs that don't have name as the first dict item are in the tests here already.

h-vetinari:
> outputs that don't have name as the first dict item are in the tests here already.

I disagree. You have
```
outputs:
  # comment
    - name: blarg  # comment
      requirements:
```
but that's just a comment in between (same goes for the cases from #2548). IOW, every `name:` field still actually comes with `- ` in front (and so it would have been sufficient to detect section boundaries by `\s*-\s+name:.*`, at least for for the current test corpus). If your intention was to support something like
```
outputs:
  - build:
      skip: true  # [not linux]
    name: ...
```
then I suggest you add a test for that.

h-vetinari:
Nevermind, I just now saw
```
    - requirements:
        host:
          - baz
      name: blarg2
```
Sorry about the noise.

PR #2553, opened 2024-05-07
TST run tests mostly without containers

### LINKED ISSUE #2550: Docker fallout, part two
After #2460 / #2462, I still cannot run tests locally; for one, `test_stdlib.py` does not have the `HAVE_CONTAINERS` skip, but more importantly, even then that turns into a trap for me, because I actually have `docker` through WSL, only that it doesn't work with the call created by the tests:
```
cmd: ['docker',
 'run',
 '-e',
 'CF_TICK_IN_CONTAINER=true',
 '--security-opt=no-new-privileges',
 '--read-only',
 '--cap-drop=all',
 '--mount',
 'type=tmpfs,destination=/tmp,tmpfs-mode=1777,tmpfs-size=6000000000',
 '-m',
 '6000m',
 '--cpus',
 '1',
 '--ulimit',
 'nofile=1024:1024',
 '--ulimit',
 'nproc=2048:2048',
 '--rm',
 '-i',
 'conda-forge-tick:test',
 '/opt/conda/envs/cf-scripts/bin/python',
 '-u',
 '/opt/autotick-bot/docker/run_bot_task.py',
 'parse-meta-yaml',
 '--platform',
 'win',
 '--arch',
 '64',
 '--log-level',
 'warning']
```

I'm also failing to identify which docker image is necessary for that (though obviously, it needs `/opt/conda/envs/cf-scripts/bin/python` to be present).

I admit that I haven't understood why dockerization was necessary in the first place... But given that it likely was for good reason, could we at least add a debug option to disable it and just execute the respective tasks directly (rather than in docker)? For running local tests I find it overkill to run in docker TBH, and it's an impediment to development for me.

beckermr:
This PR changes the tests to run mostly without containers. It also turns off container tests unless docker is available and the right image has been built locally.

closes #2550

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,33 @@ If your migrator needs special configuration, you should write a new factory fun
 
 ## Developer Documentation
 
+### Useful Environment Variables
+
+- `CF_TICK_GRAPH_DATA_BACKENDS`: See [`LazyJson` Data Structures and Backends](#lazyjson-data-structures-and-backends) below.
+- `CF_TICK_GRAPH_DATA_USE_FILE_CACHE`: See [`LazyJson` Data Structures and Backends](#lazyjson-data-structures-and-backends) below.
+- `MONGODB_CONNECTION_STRING`: See [`LazyJson` Data Structures and Backends](#lazyjson-data-structures-and-backends) below.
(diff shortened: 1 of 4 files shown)


beckermr:
cc @h-vetinari this should make things automatic for you. If you ever want to debug containers on WSL, build the docker image with the test tag locally. I added docs in the readme.

PR #2573, opened 2024-05-10
REF make migrators in another job

beckermr:
This PR adds a separate job to make the migrators.

To do:
- [x] add function to load migrators
- [x] check that other code using initialize_migrators still works

Most of the diff here is a straight copy paste of code from one file to another.


beckermr:
Time for a YOLO merge.

PR #2615, opened 2024-05-16
PROD end migrators faster

beckermr:
This will end migrations faster by turning off solver checks after 30 days instead of 90.

### DIFF
--- a/conda_forge_tick/make_migrators.py
+++ b/conda_forge_tick/make_migrators.py
@@ -92,7 +92,7 @@ logger = logging.getLogger(__name__)
 PR_LIMIT = 5
 MAX_PR_LIMIT = 50
 MAX_SOLVER_ATTEMPTS = 50
-CHECK_SOLVABLE_TIMEOUT = 90  # 90 days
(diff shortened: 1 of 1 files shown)


beckermr:
cc @h-vetinari

h-vetinari:
correct me if I'm wrong, but this will not (yet) open PRs for those failing feedstocks after those 30 days? It'll simply mark the migration as done?

I don't yet know how much work it is, but I'd prefer to just open a PR (even if unresolvable) after ~10 days

beckermr:
This turns off solver checks and so any PR that can be made without a bot error will be made.

beckermr:
So one thing that happened in the past is that we'd make a migration and the bot would make a bunch of PRs. Then we'd find a bug/issue and have to go around manually marking PRs as bot-rerun in order to fix the migration issue. This was a huge pain and has motivated both the solver checks and the fact that the bot slowly ramps up migrations. 

For this reason, ~10 days is just too short a time IMHO for us to start ignoring the solver checks.

beckermr:
I am going to merge this. We should discuss making the time even lower in a future PR.

PR #2620, opened 2024-05-16
ENH ramp PRs better

beckermr:
This PR should ramp PRs for migrators better. It starts them slow, then bumps them up in rate for a while and then back down. Hopefully, this will emphasize newly merged migrators a bit over older ones that have gone through most of their work.

### DIFF
--- a/conda_forge_tick/make_migrators.py
+++ b/conda_forge_tick/make_migrators.py
@@ -214,7 +214,7 @@ def add_rebuild_migration_yaml(
     migration_yaml: str,
     config: dict,
     migration_name: str,
-    pr_limit: int = PR_LIMIT,
+    nominal_pr_limit: int = PR_LIMIT,
     max_solver_attempts: int = 3,
 ) -> None:
     """Adds rebuild migrator.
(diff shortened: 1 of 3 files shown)


beckermr:
cc @h-vetinari

conda_forge_tick/make_migrators.py
```diff
@@ -297,17 +297,54 @@ def add_rebuild_migration_yaml(
...
-        pr_limit=pr_limit,
+        pr_limit=nominal_pr_limit,
         top_level=top_level,
         cycles=cycles,
         piggy_back_migrations=piggy_back_migrations,
         max_solver_attempts=max_solver_attempts,
         **config,
     )
+
+    # adaptivewl set PR limits based on the number of PRs made so far
```
beckermr, line 313:
```suggestion
    # adaptively set PR limits based on the number of PRs made so far
```

h-vetinari:
> Hopefully, this will emphasize newly merged migrators a bit over older ones that have gone through most of their work.

So this happened while I slept, but just to emphasise: I don't care about the speed of one particular migration. The numpy2 migration was the cause for me to look a bit more under the hood of what the bot does, and what I do care about is distributing/using our cycles effectively.

I still think the biggest win would be to open PRs for "not solvable" _and_ "bot error" after some fixed timeout. Having a concrete PR to get notified & look at is both beneficial for the maintainers, keeps the bot from burning cycles retrying the same thing over and over, and progresses much faster through the migration.

In any case, thanks for the tweaks and tuning here!

PR #2622, opened 2024-05-17
ENH force issue PRs after many many solver attempts

beckermr:
OK @h-vetinari here is an initial attempt at issuing PRs after many failed solver attempts per feedstock. Right now I set this to 100 attempts. A bot job takes ~2 hours, so the bot will try a PR for about a week before issuing it. This matches the other timescales we give maintainers to fix things up. We don't actually notify folks of failed solver attempts, but if we did, the week would be the right timescale.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -270,6 +270,8 @@ def run(
     if (
         feedstock_ctx.feedstock_name != "conda-forge-pinning"
         and (base_branch == "master" or base_branch == "main")
+        # feedstocks that have problematic bootstrapping will not always be solvable
+        and feedstock_ctx.feedstock_name not in BOOTSTRAP_MAPPINGS
         and (
             (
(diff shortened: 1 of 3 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -284,6 +289,10 @@ def run(
         )
         # feedstocks that have problematic bootstrapping will not always be solvable
         and feedstock_ctx.feedstock_name not in BOOTSTRAP_MAPPINGS
+        # we try up to MAX_SOLVER_ATTEMPTS times and then we just skip
```
beckermr, line 292:
```suggestion
        # we try up to force_pr_after_solver_attempts times and then we just skip
```

beckermr:
I'm going to wait until this afternoon to merge to give folks time to respond in the core chat to my note.

beckermr:
YOLO!

PR #2643, opened 2024-05-21
Begin Building on a modular Git Backend

ytausch:
In my efforts about adding easier local debugging for the bot, we need a way to let the bot run without the need of having a conda-forge GitHub token available. Accomplishing this is not easy with the current codebase because the git and token logic is baked into it.

With this PR, I begin to decouple the git logic and make everything interfacing with GitHub replaceable by a dry run backend (that is currently still unused). To keep the PR small, not all git operations but only some now use the new object-oriented git approach. Further PRs will follow that replace more of the logic.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -192,13 +192,7 @@ def run(
 
     # TODO: run this in parallel
     feedstock_dir, repo = get_repo(
-        fctx=feedstock_ctx,
-        branch=branch_name,
-        feedstock=feedstock_ctx.feedstock_name,
-        protocol=protocol,
-        pull_request=pull_request,
-        fork=fork,
-        base_branch=base_branch,
+        fctx=feedstock_ctx, branch=branch_name, base_branch=base_branch
     )
     if not feedstock_dir or not repo:
         logger.critical(
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -29,8 +29,8 @@ from conda_forge_tick.feedstock_parser import BOOTSTRAP_MAPPINGS
 from conda_forge_tick.git_utils import (
     GIT_CLONE_DIR,
     comment_on_pr,
-    get_github_api_requests_left,
     get_repo,
+    github_backend,
     is_github_api_limit_reached,
     push_repo,
 )
@@ -680,7 +680,8 @@ def _run_migrator_on_feedstock_branch(
 
 def _is_migrator_done(_mg_start, good_prs, time_per, pr_limit):
     curr_time = time.time()
-    api_req = get_github_api_requests_left()
+    backend = github_backend()
+    api_req = backend.get_api_requests_left()
 
     if curr_time - START_TIME > TIMEOUT:
         logger.info(
@@ -1107,5 +1108,5 @@ def main(ctx: CliContext) -> None:
             #     ],
             # )
 
-    logger.info("API Calls Remaining: %d", get_github_api_requests_left())
+    logger.info("API Calls Remaining: %d", github_backend().get_api_requests_left())
     logger.info("Done")
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -604,7 +604,8 @@ def _run_migrator_on_feedstock_branch(
                 fctx.feedstock_name,
             )
 
-            if is_github_api_limit_reached(e):
+            if is_github_api_limit_reached():
+                logger.warning("GitHub API error", exc_info=e)
                 break_loop = True
 
     except VersionMigrationError as e:
--- a/conda_forge_tick/executors.py
+++ b/conda_forge_tick/executors.py
@@ -39,8 +39,8 @@ logger = logging.getLogger(__name__)
 class DaskRLock(DaskLock):
     """A reentrant lock for dask that is always blocking and never times out."""
 
-    def __init__(self, name: str):
-        super().__init__(name=name)
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         self._rcount = 0
         self._rdata = None
 
--- a/conda_forge_tick/executors.py
+++ b/conda_forge_tick/executors.py
@@ -72,7 +72,7 @@ def _init_process(lock):
 
 def _init_dask(lock):
     global DRLOCK
-    # it appears we have to construct the locak by name instead
+    # it appears we have to construct the lock by name instead
     # of passing the object itself
     # otherwise dask uses a regular lock
     DRLOCK = DaskRLock(name=lock)
--- a/conda_forge_tick/executors.py
+++ b/conda_forge_tick/executors.py
@@ -21,16 +21,30 @@ PRLOCK = DummyLock()
 DRLOCK = DummyLock()
 
 
+@contextlib.contextmanager
+def lock_git_operation():
+    """
+    A context manager to lock git operations - it can be acquired once per thread, once per process,
+    and once per dask worker.
+    Note that this is a reentrant lock, so it can be acquired multiple times by the same thread/process/worker.
+    """
+
+    with TRLOCK, PRLOCK, DRLOCK:
+        yield
+
+
 logger = logging.getLogger(__name__)
 
 
 class DaskRLock(DaskLock):
     """A reentrant lock for dask that is always blocking and never times out."""
 
-    def acquire(self):
-        if not hasattr(self, "_rcount"):
-            self._rcount = 0
+    def __init__(self, name: str):
+        super().__init__(name=name)
+        self._rcount = 0
+        self._rdata = None
 
+    def acquire(self, *args):
         self._rcount += 1
 
         if self._rcount == 1:
@@ -39,13 +53,13 @@ class DaskRLock(DaskLock):
         return self._rdata
 
     def release(self):
-        if not hasattr(self, "_rcount") or self._rcount == 0:
+        if self._rcount == 0:
             raise RuntimeError("Lock not acquired so cannot be released!")
 
         self._rcount -= 1
 
         if self._rcount == 0:
-            delattr(self, "_rdata")
+            self._rdata = None
             return super().release()
         else:
             return None
--- a/conda_forge_tick/executors.py
+++ b/conda_forge_tick/executors.py
@@ -16,9 +16,9 @@ class DummyLock:
         pass
 
 
-TRLOCK = TRLock()
-PRLOCK = DummyLock()
-DRLOCK = DummyLock()
+GIT_LOCK_THREAD = TRLock()
+GIT_LOCK_PROCESS = DummyLock()
+GIT_LOCK_DASK = DummyLock()
 
 
 @contextlib.contextmanager
@@ -29,7 +29,7 @@ def lock_git_operation():
     Note that this is a reentrant lock, so it can be acquired multiple times by the same thread/process/worker.
     """
 
-    with TRLOCK, PRLOCK, DRLOCK:
+    with GIT_LOCK_THREAD, GIT_LOCK_PROCESS, GIT_LOCK_DASK:
         yield
 
 
@@ -66,16 +66,16 @@ class DaskRLock(DaskLock):
 
 
 def _init_process(lock):
-    global PRLOCK
-    PRLOCK = lock
+    global GIT_LOCK_PROCESS
+    GIT_LOCK_PROCESS = lock
 
 
 def _init_dask(lock):
-    global DRLOCK
+    global GIT_LOCK_DASK
     # it appears we have to construct the lock by name instead
     # of passing the object itself
     # otherwise dask uses a regular lock
-    DRLOCK = DaskRLock(name=lock)
+    GIT_LOCK_DASK = DaskRLock(name=lock)
 
 
 @contextlib.contextmanager
@@ -84,8 +84,8 @@ def executor(kind: str, max_workers: int, daemon=True) -> typing.Iterator[Execut
 
     This allows us to easily use other executors as needed.
     """
-    global DRLOCK
-    global PRLOCK
+    global GIT_LOCK_DASK
+    global GIT_LOCK_PROCESS
 
     if kind == "thread":
         with ThreadPoolExecutor(max_workers=max_workers) as pool_t:
@@ -99,7 +99,7 @@ def executor(kind: str, max_workers: int, daemon=True) -> typing.Iterator[Execut
             initargs=(lock,),
         ) as pool_p:
             yield pool_p
-        PRLOCK = DummyLock()
+        GIT_LOCK_PROCESS = DummyLock()
     elif kind in ["dask", "dask-process", "dask-thread"]:
         import dask
         import distributed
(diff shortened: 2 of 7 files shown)


beckermr:
This PR in terms of scope and functionality is IMHO really great. It should be relative straightforward to review. I'll get to it by the end of the week. Thank you!

ytausch:
Thank you too!

conda_forge_tick/auto_tick.py
```diff
@@ -192,13 +192,7 @@ def run(
...
     # TODO: run this in parallel
     feedstock_dir, repo = get_repo(
-        fctx=feedstock_ctx,
-        branch=branch_name,
-        feedstock=feedstock_ctx.feedstock_name,
-        protocol=protocol,
-        pull_request=pull_request,
-        fork=fork,
-        base_branch=base_branch,
+        fctx=feedstock_ctx, branch=branch_name, base_branch=base_branch
```
beckermr, line 195: I get why the API was changed here, given that we never used non-default values and will never do so either. We should at some point clean up the keywords on the surrounding function as well, but not today!

conda_forge_tick/git_utils.py
```diff
@@ -88,77 +97,533 @@ def github_client() -> github.Github:
...
+        :param to_treeish: The treeish to reset to. Defaults to "HEAD".
+        :raises GitCliError: If the git command fails.
+        """
+        self._run_git_command(["reset", "--quiet", "--hard", to_treeish], git_dir)
+
+    def clone_repo(self, origin_url: str, target_dir: Path):
+        """
+        Clone a Git repository. If target_dir already exists, it will be reset to the state of the git index
+        with `git reset --hard HEAD`. This may be unexpected behavior because target_dir could contain a
+        different repository. Use with caution.
```
beckermr, line 168: If you did this at the CLI, git would raise an error. We should do the same I think.

conda_forge_tick/git_utils.py
```diff
@@ -88,77 +97,533 @@ def github_client() -> github.Github:
...
+                )
+
+                # not sure why this is needed, but it was in the original code
+                self.checkout_new_branch(
+                    target_dir,
+                    base_branch,
+                    start_point=f"upstream/{base_branch}",
+                )
+
+        # not sure why this is needed, but it was in the original code
```
beckermr, line 321: Thanks for leaving this as is! The code here has years of weird bugs and edge cases. Not all of them were documented. :/

conda_forge_tick/git_utils.py
```diff
@@ -88,77 +97,533 @@ def github_client() -> github.Github:
...
+                f"It seems branch {new_branch} does not exist. Creating it.",
+            )
+            self.checkout_new_branch(target_dir, new_branch, start_point=base_branch)
+
+
+class GitPlatformBackend(ABC):
     """
-    gh = github3_client()
+    A backend for interacting with a git platform (e.g. GitHub).
+    Note that this class is not thread-safe, you should create a new instance for each thread.
```
beckermr, line 339: I think the lack of thread-safety is related to git and the backend more than the class in and of itself. In other words, making git commands from multiple threads on the same repo is likely not a good idea either. Similarly, making multiple API calls to GitHub trying to fork a repo from multiple threads is not a great idea, though it may in fact work. 

Should we protect calls to methods on this class (and its non-abstract children) and the class above via an RLock (https://docs.python.org/3/library/threading.html#rlock-objects) or some similar scheme?

conda_forge_tick/git_utils.py
```diff
@@ -88,77 +97,533 @@ def github_client() -> github.Github:
...
+
+        Additionally, implementations may print diagnostic information about the API limit.
+        """
+        return self.get_api_requests_left() in (0, None)
+
+
+class GitHubBackend(GitPlatformBackend):
+    """
+    A git backend for GitHub, using both PyGithub and github3.py as clients.
+    It is unclear why both clients are used, in the future, this should be refactored to use only one client.
```
beckermr, line 467:
```suggestion
    Both clients are used for historical reasons. In the future, this should be refactored to use only one client.
```

conda_forge_tick/git_utils.py
```diff
@@ -88,77 +97,533 @@ def github_client() -> github.Github:
...
+    def __init__(self):
+        super().__init__(GitCli())
+        self._repos: set[str] = set()
+
+    def get_api_requests_left(self) -> Bound:
+        return Bound.INFINITY
+
+    def does_repository_exist(self, owner: str, repo_name: str) -> bool:
+        if owner == self._USER:
+            return repo_name in self._repos
```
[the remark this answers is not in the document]
beckermr: I think that's the point though. This method tests if the fork exists so will return False if it does not.

conda_forge_tick/git_utils.py
```diff
@@ -308,63 +684,35 @@ def get_repo(
...
-    # first, let's grab the feedstock locally
-    upstream = feedstock_url(fctx=fctx, protocol=protocol)
-    origin = fork_url(upstream, gh_username)
-    feedstock_reponame = feedstock_repo(fctx=fctx)
+    try:
+        backend.fork("conda-forge", feedstock_repo_name)
+    except RepositoryNotFoundError:
+        logger.warning(f"Could not fork conda-forge/{feedstock_repo_name}")
+        with fctx.attrs["pr_info"] as pri:
+            pri["bad"] = f"{fctx.package_name}: does not match feedstock name\n"
```
beckermr, line 695: We need to make an issue to remind us to remove the package_name attribute from the feedstock context object and just use feedstock_name to avoid this.

tests/test_git_utils.py
```diff
@@ -1,4 +1,842 @@
...
+        with pytest.raises(GitCliError):
+            cli.clone_repo(
+                "https://github.com/conda-forge/this-repo-does-not-exist.git", dir_path
+            )
+
+
+def test_git_cli_clone_repo_success():
+    cli = GitCli()
+
+    git_url = "https://github.com/conda-forge/duckdb-feedstock.git"
```
beckermr, line 169: I am pretty sure this feedstock has been archived based on other PRs I looked at today. Maybe pick another one? Sometimes I used conda-smithy or similar since we likely won't ever archive that one.

tests/test_git_utils.py
```diff
@@ -1,4 +1,842 @@
...
+
+        # delete the README.md file
+        readme_file.unlink()
+
+        assert not readme_file.exists()
+
+        # clone the repo again (in the same directory!) - this should reset the repo
+        cli.clone_repo(git_url, dir_path)
+
+        assert readme_file.exists()
```
beckermr, line 187: If we make this command error, then we need to change the tests here.

beckermr (changes requested):
I left a few comments. The tests are exhaustive which is great. I did notice some tests where the code was tested live and it had tests with mocks. I am all for using mocks where needed, but tests of the code working live are more valuable. Thus in cases where we have live tests and mock tests, I'd delete the tests with mocks (assuming they are purely duplicative).

conda_forge_tick/git_utils.py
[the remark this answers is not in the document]
ytausch: Yes, exactly. Note that this does not prevent us from calling `does_repository_exist` with `regro-cf-autotick-bot` (the actual bot user) as repository owner since `_USER` is a special placeholder value.

All forks are created by the virtual `_USER` user, simulating GitHub behaviour.

conda_forge_tick/git_utils.py
```diff
@@ -178,17 +643,14 @@ def feedstock_url(fctx: FeedstockContext, protocol: str = "ssh") -> str:
...
     return url
 
 
 def feedstock_repo(fctx: FeedstockContext) -> str:
     """Gets the name of the feedstock repository."""
-    repo = fctx.feedstock_name + "-feedstock"
-    if repo.endswith(".git"):
-        repo = repo[:-4]
-    return repo
+    return fctx.feedstock_name + "-feedstock"
```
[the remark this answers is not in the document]
ytausch: Just found out that `fork_url` is not used anyway, so I removed it.

tests/test_git_utils.py
```diff
@@ -1,4 +1,842 @@
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        dir_path = Path(tmpdir)
+
+        init_temp_git_repo(dir_path)
+        cli._run_git_command(
+            ["commit", "--allow-empty", "-m", "Initial commit"],
+            working_directory=dir_path,
+        )
+
+        cli.reset_hard(dir_path)
```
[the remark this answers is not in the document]
ytausch: The point of this test was to check if reset hard fails if the repository is already reset - the actual reset functionality is tested in `test_git_cli_reset_hard`. I've now added an inspection of `git log` to additionally check if all commits are preserved. This should be quite complete now IMO.

conda_forge_tick/git_utils.py
ytausch: Agreed.

ytausch: The logic is now moved into clone_fork_and_branch

conda_forge_tick/git_utils.py
ytausch: I am a bit confused now. I understand that
- `package_name` is the networkx node name of the feedstock
- Here, `feedstock_repo` uses the `feedstock_name` to generate the git URL
- So if the git URL does not exist, the value of `feedstock_name` is invalid.

Why is the error message "package_name does not match feedstock_name" then? (I just copied this from the existing code)

Are there any instances where `feedstock_name` is actually different from `package_name`? Or is this exactly the point you want to make?

beckermr: Yes this is the point I am making. Only feedstock_name matters. We should remove package_name all together. The names of the packages can differ from the feedstock, but bot's graph nodes use the feedstock_name.

tests/test_git_utils.py
ytausch: Actually I used this feedstock exactly because it has been archived. My criteria for a testing feedstock was:

1. The feedstock should not be deleted in the near future.
2. Its content should somehow be consistent, i.e. the README file should not disappear because it is used in some tests.
3. It should have as little commits as possible to make testing faster.

conda-smithy and other popular feedstocks satisfy all of these except 3. Less popular feedstocks might violate 2 because we don't really have control about what happens in these.

This is why I chose an archived feedstock: Its content will not change, and also the number of commits is fixed. I did this with the assumption that conda-forge does not delete archived feedstocks - what do you think?

conda_forge_tick/git_utils.py
beckermr: xref: https://github.com/regro/cf-scripts/pull/2660

tests/test_git_utils.py
beckermr: Oh I didn't realize github let you clone archived repos.

ytausch:
I would actually like to keep the mocked tests, as they

- test if the git methods correctly translate to git CLI arguments, which is pretty trivial to do with mocks and pretty hard if we only rely on inspecting the state of the git repository afterwards
- test complicated edge cases that are a lot harder to test with the real git cli, see e.g. `test_git_cli_clone_fork_and_branch_mock` -> there, the real git cli is only used for a minimal test

Since they are not redundant, it's worth keeping them IMO. WDYT?

beckermr:
Sure ok.

conda_forge_tick/git_utils.py
ytausch: Good point (and my comment in the code is indeed wrong). An ever better solution is probably per-repository locking for the CLI as well as for GitHub. I will try to come up with a good solution for it.

beckermr: Ok. We don't need this to be too complex. The bot doesn't normally checkout more than one repo at a time and a single global lock would likely be ok.

beckermr: The executors.py file has a set of global locks used to ensure the hdel operations in the lazy json backends don't cause an error. Maybe use those?

beckermr: Actually looking at the code, you have to use those since they cover multiprocessing, dask, and python threads.

ytausch: You're right that a single global lock is probably sufficient for now - however I noticed that TRLOCK (thread-scoped) and PRLOCK (process-scoped) are reentrant locks while DLOCK (for dask) is not, which would make implementation a little bit more complex.

Since we don't use dask anyway, are you fine with removing it in a separate PR?

beckermr: We've used dask in the past and should keep it. Maybe dask provides a reentrant lock?

beckermr: I bet we could make the dask lock reentrant easily enough by subclassing and overriding the acquire and release methods to test on self._locked in the right way before calling the parent methods. I'll make a pr later.

beckermr: OK dask locking is fixed up and tested!

conda_forge_tick/executors.py
```diff
@@ -21,16 +21,30 @@ def __exit__(self, *args, **kwargs):
...
 class DaskRLock(DaskLock):
     """A reentrant lock for dask that is always blocking and never times out."""
 
-    def acquire(self):
-        if not hasattr(self, "_rcount"):
-            self._rcount = 0
+    def __init__(self, name: str):
+        super().__init__(name=name)
+        self._rcount = 0
+        self._rdata = None
```
beckermr, line 45: This init signature is wrong. Please see the dask documentation. I specifically avoided overriding the init so I didn't have to keep the init here up to date with changes in dask. Please roll back the changes to this class and use the `getattr` and `setattr` approach.

ytausch: The only place where `DaskRLock` is instantiated is our own code (in `_init_dask`) - so there should be no problem with changing the `__init__` signature. There are readability benefits when all attributes are defined in `__init__`.

conda_forge_tick/git_utils.py
ytausch: I added it to our git methods

conda_forge_tick/executors.py
beckermr: Please back these changes out. We've initiated the dask lock with a client in the past.

ytausch: How about this (see my latest commit)?

beckermr: Yes that will work. Thank you!

ytausch:
@beckermr I think the test failures are not related to my changes here. Could you have a look, please?

beckermr (changes requested):
I'll need to go through more of the code later this week, but here is a comment to start.

conda_forge_tick/executors.py
```diff
@@ -21,16 +21,30 @@ def __exit__(self, *args, **kwargs):
 DRLOCK = DummyLock()
 
 
+@contextlib.contextmanager
+def lock_git_operation():
```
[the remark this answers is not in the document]
ytausch: But isn't this problematic? Let's say we also want to restrict parallelism for "eating bananas" in the future. Then using a single lock for everything would prevent us to "eat bananas" in thread/process/dask worker A and to run git operations in thread/process/dask worker B at the same time although we most likely want that, which defeats multithreading/-processing.

It would be better to use more locks for different things, considering that one global lock for all git-related operations is a big compromise already.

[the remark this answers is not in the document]
beckermr: Sure so rename the TRLOCK, PRLOCK and DRLOCK variables then.

beckermr (changes requested):
Just need to resolve the comment above and then we can merge, Nice work!

conda_forge_tick/executors.py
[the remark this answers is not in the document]
ytausch: Done.

ytausch:
Done!

beckermr (approved):
LGTM and thank you!

beckermr:
We are seeing bot errors due to this and need to revert.

PR #2649, opened 2024-05-22
Use version prefix from github URL

ZzEeKkAa:
Pick up release prefix from the archive URL.

After this changes version at is picked up properly:
```
> CF_TICK_IN_CONTAINER=true python -m conda_forge_tick.cli --debug --online update-upstream-versions --job=1 --n-jobs=1 spirv-headers

2024-05-22 15:26:16,058 DEBUG    conda_forge_tick.update_upstream_versions || Fetching latest version for spirv-headers from Github...
2024-05-22 15:26:16,058 DEBUG    conda_forge_tick.update_sources || Found version prefix from url: vulkan-sdk-
2024-05-22 15:26:16,058 DEBUG    conda_forge_tick.update_upstream_versions || Using URL https://github.com/khronosgroup/spirv-headers/releases.atom
2024-05-22 15:26:16,801 DEBUG    conda_forge_tick.update_upstream_versions || Found version 1.3.283.0 on Github
2024-05-22 15:26:16,801 INFO     conda_forge_tick.update_upstream_versions || # 0     - spirv-headers - 1.3.268.0 - 1.3.283.0
```

Fixes: #2648

### DIFF
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -563,10 +563,14 @@ class Github(VersionFromFeed):
         self.ver_prefix_remove = [self.version_prefix] + self.ver_prefix_remove
 
     def get_version_prefix(self, version: str, split_url: list[str]):
+        """Returns prefix for the first split that contains version. If prefix
+        is empty - returns None."""
         r = re.compile(rf"^(.*){version}")
         for split in split_url:
             match = r.match(split)
-            if match is not None and match.group(1) != "":
+            if match is not None:
+                if match.group(1) == "":
+                    return None
                 return match.group(1)
 
         return None
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -553,13 +553,14 @@ class IncrementAlphaRawURL(BaseRawURL):
 
 class Github(VersionFromFeed):
     name = "Github"
+    version_prefix = None
 
     def set_version_prefix(self, version: str, split_url: list[str]):
-        version_prefix = self.get_version_prefix(version, split_url)
-        if version_prefix is None:
+        self.version_prefix = self.get_version_prefix(version, split_url)
+        if self.version_prefix is None:
             return
-        logger.debug(f"Found version prefix from url: {version_prefix}")
-        self.ver_prefix_remove = [version_prefix] + self.ver_prefix_remove
+        logger.debug(f"Found version prefix from url: {self.version_prefix}")
+        self.ver_prefix_remove = [self.version_prefix] + self.ver_prefix_remove
 
     def get_version_prefix(self, version: str, split_url: list[str]):
         r = re.compile(rf"^(.*){version}")
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -555,7 +555,6 @@ class Github(VersionFromFeed):
     name = "Github"
 
     def set_version_prefix(self, version: str, split_url: list[str]):
-        logger.debug(f"dbg: {split_url}")
         version_prefix = self.get_version_prefix(version, split_url)
         if version_prefix is None:
             return
@@ -566,7 +565,7 @@ class Github(VersionFromFeed):
         r = re.compile(rf"^(.*){version}")
         for split in split_url:
             match = r.match(split)
-            if match is not None:
+            if match is not None and match.group(1) != "":
                 return match.group(1)
 
         return None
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
The missed lines in the coverage indicate to me that while the PR works, it may not be acting as intended. It appears to always find a version prefix even though some github urls do not have any prefixes (e.g., `https://github.com/regro/conda-forge-feedstock-check-solvable/archive/refs/tags/0.5.2.tar.gz`) and others may have bits that seem like a version prefix but are actually a package name (e.g., `https://github.com/simplejson/simplejson/releases/download/{{ version }}/simplejson-{{ version }}.tar.gz`). 

Can you add some tests so we can see what is happening and how we might fix it?

ZzEeKkAa:
I guess it is due to the fact that prefix can be empty string. This scenario then won't never happen, but let me add those tests to fix the coverage.

beckermr:
I actually think it is the opposite. The code never finds an empty string for the prefix which is suspicious.

ZzEeKkAa:
I've run it through debugger and it was a case for pairs like this: ('1.2.3', 'https://example.com/archs/1.2.3.tar.gz'). It finds '1.2.3' in '1.2.3.tar.gz' and gets prefix as empty string. Simply checking for empty string should fix coverage issue, since this part of the code will be reachable.

ZzEeKkAa:
@beckermr how to regenerate coverage?

ZzEeKkAa:
Okay, now it is wise versa, haha. So I gonna add test now

tests/test_upstream_versions.py
```diff
@@ -1643,3 +1643,41 @@ def test_main(
...
+        ("https://example.com/archs/example-1.2.3.tar.gz", "1.2.3", None),
+        (
+            "https://github.com/example/example/archive/refs/tags/1.2.3.tar.gz",
+            "1.2.3",
+            None,
+        ),
+        (
+            "https://github.com/example/example/archive/refs/tags/example-1.2.3.tar.gz",
+            "1.2.3",
+            "example-",
```
beckermr, line 1662: This isn't always version prefix even for github. Some github urls like this

> https://github.com/simplejson/simplejson/releases/download/{{ version }}/simplejson-{{ version }}.tar.gz

are actually the package name.

beckermr: I think for urls with `/refs/tags` in them, the test case is correct.

conda_forge_tick/update_sources.py
```diff
@@ -553,11 +553,30 @@ def get_url(self, meta_yaml) -> Optional[str]:
...
+                return match.group(1)
+
+        return None
 
     def get_url(self, meta_yaml) -> Optional[str]:
         if "github.com" not in meta_yaml["url"]:
             return None
         split_url = meta_yaml["url"].lower().split("/")
+        version = meta_yaml["version"]
+        self.set_version_prefix(version, split_url)
```
beckermr, line 579:
```suggestion
        if "/archive/refs/tags/" in meta_yaml["url"]:
            self.set_version_prefix(version, split_url)
```

Maybe we can do this for now? This would only search for prefixes in the most common case of using a url of the source archive from a GH release.

tests/test_upstream_versions.py
ZzEeKkAa: I've just added support to find first split with version. There still could be examples like:
```
https://github.com/simplejson/simplejson/releases/download/my-name-1.2.3/simplejson-1.2.3.tar.gz
```

And currently it won't find it because of the original issue. Prefix is `my-name-` that is not predefined and contains `a` in name. So it will be treated as alpha version and ignored.

conda_forge_tick/update_sources.py
ZzEeKkAa: See my other comment. I guess I've added better solution.

ZzEeKkAa:
@beckermr  I'm now worrying about cases when product release may have subtools with different version. Something like this: `https://github.com/foo/bar/releases/download/foo-1.2.3/bar-2.3.4.src.tar.xz`

But it is out of scope of this PR.

beckermr:
The first split is a heuristic, but probably not the worst one. Will merge once tests pass.

beckermr (approved):

ZzEeKkAa:
@beckermr when does it get to production? I've tried to triggered the script from here: https://github.com/conda-forge/spirv-headers-feedstock/issues/13 but it could not fetch new version

ZzEeKkAa:
Thank you @beckermr !

PR #2664, opened 2024-05-27
BUG fix some errors in the pydantic models

beckermr:
This PR fixes some logical errors in the pydantic models.

### DIFF
--- a/conda_forge_tick/models/pr_info.py
+++ b/conda_forge_tick/models/pr_info.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from enum import StrEnum
-from typing import Annotated, Any, Literal, Self
+from typing import Annotated, Any, Literal, Optional, Self
 
 from pydantic import (
     UUID4,
@@ -14,7 +14,6 @@ from conda_forge_tick.models.common import (
     CondaVersionString,
     PrJsonLazyJsonReference,
     StrictBaseModel,
-    ValidatedBaseModel,
     before_validator_ensure_dict,
 )
 from conda_forge_tick.models.pr_json import PullRequestData, PullRequestState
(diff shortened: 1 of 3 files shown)


conda_forge_tick/models/pr_info.py
```diff
@@ -269,7 +295,7 @@ class PrInfoValid(StrictBaseModel):
     The Azure token errors should be removed from the graph, e.g. by parsing the model and re-serializing it.
     """
 
-    pre_pr_migrator_status: dict[str, str]
+    pre_pr_migrator_status: Optional[dict[str, str]] = None
```
ytausch, line 298: What is the difference between this field being `None` and being an empty dictionary? If there is none, you should use an empty dict as value instead and make this non-optional.

https://github.com/pydantic/pydantic/discussions/8062#discussioncomment-7517322

@beckermr

conda_forge_tick/models/pr_info.py
```diff
@@ -285,7 +311,7 @@ class PrInfoValid(StrictBaseModel):
     If a migration is eventually successful, the corresponding key is removed from the dictionary.
     """
 
-    pre_pr_migrator_attempts: dict[str, int]
+    pre_pr_migrator_attempts: Optional[dict[str, int]] = None
```
ytausch, line 314: see above

conda_forge_tick/models/pr_info.py
```diff
@@ -297,45 +323,13 @@ class PrInfoValid(StrictBaseModel):
 
     @model_validator(mode="after")
     def check_pre_pr_migrations(self) -> Self:
-        if self.pre_pr_migrator_status.keys() != self.pre_pr_migrator_attempts.keys():
+        ppms = getattr(self, "pre_pr_migrator_status", {}) or {}
+        ppma = getattr(self, "pre_pr_migrator_attempts", {}) or {}
```
ytausch, line 327: This is not needed if you make the fields non-optional as I suggested

conda_forge_tick/models/version_pr_info.py
```diff
@@ -17,35 +17,29 @@ class VersionPrInfoValid(ValidatedBaseModel):
     auto_tick run.
     """
 
-    new_version_attempts: dict[str, int | float]
+    new_version_attempts: Optional[dict[str, int | float]] = None
```
ytausch, line 20: see my other comment

conda_forge_tick/models/version_pr_info.py
```diff
@@ -17,35 +17,29 @@ class VersionPrInfoValid(ValidatedBaseModel):
...
     """
     Mapping (version -> number of attempts) to describe the number of attempts that were made to update the versions in
     the PR. The value can be a float since failing with a solving error is counted as a partial attempt, which only
     increases the count by a fraction that is specified in auto_tick.run (03/2024: 0.2).
     """
 
-    new_version_errors: dict[str, str]
+    new_version_errors: Optional[dict[str, str]] = None
```
ytausch, line 27: see my other comment

conda_forge_tick/models/version_pr_info.py
```diff
@@ -17,35 +17,29 @@ class VersionPrInfoValid(ValidatedBaseModel):
...
     """
 
     @model_validator(mode="after")
     def check_new_version_error_keys(self):
-        wrong_versions = [
-            version
-            for version in self.new_version_errors
-            if version not in self.new_version_attempts
-        ]
+        nve = getattr(self, "new_version_errors", {}) or {}
```
ytausch, line 35: not needed if you apply my suggestion

conda_forge_tick/models/pr_info.py
beckermr: The keys are missing in some json blobs and that is ok. I thought this was the proper syntax for a field that could be one of missing, set to None, an empty dictionary, or a dictionary with values.

ytausch: I know that the JSON files do not differentiate between empty and absent values.

However, one of the goals of the Pydantic model was to use it in the future for actually parsing the JSON files in production code. Pydantic uses the default value you supply if a field is absent, which does not need to be None. In case of container types, an empty container (here: dict) does actually represent what we want, so we should use that instead.

This would also remove some of the ugly code that checks for None values.

beckermr: I am confused here. If the field is not optional then pydantic throws an error if it is not there. So are you saying the default value should be an empty dictionary? Would that still allow a value of None?

ytausch: > If the field is not optional then pydantic throws an error if it is not there.

No, Pydantic throws an error if a field *which does not have a default value* is not there.

> So are you saying the default value should be an empty dictionary?

Yes, and because of how Python works, you have to create it using a `default_factory` (see my link above).

> Would that still allow a value of None?

If the field is non-optional, `None` would be an invalid value for it. All JSON files that do not have `bad` set do not have `None` as a value in their container fields, which is why I originally created a separate model for `bad` objects. If you really want to merge the "good" and "bad" models and "bad" models use None for empty container values, you should convert `None` values explicitly like I did with `NoneIsEmptyList` in common.py

It's probably easiest to see if you appreciate that `None` does not play a special role in Pydantic and is just a value like any other value.

beckermr: > All JSON files that do not have bad set do not have None as a value in their container fields

This is not true. The bot's data model is very flexible and None is a fine value even if bad is set to False. 

Let me look at the other comments.

beckermr: I see. So the Optional adds None as a possible value and using the default as an empty dict is helpful to eliminate some of the code. 

However, since the bot allows None, we still need code to coerce that to an empty dict. I'll check the list thing in common.py to understand that too.

ytausch: > This is not true. The bot's data model is very flexible and None is a fine value even if bad is set to False.

It might be that you consider None a "fine value" but still, all non-bad JSON files passed the validation even before you added None as a default value. This is what I meant with my statement.

> So the thing in common.py does the coercion but it seems to indicate that None is not a valid value when it actually is. We can go through and migrate the bot schema and code to eliminate this, but as it stands now None is ok.

Adding the type coercion from None to empty dict to Pydantic does not indicate that None is "invalid" IMO, because the model still **accepts** None. It would indicate that we, in the future, want to interpret `None` values as empty as soon as we use the model to read the data. Indicating this intent in the Pydantic model does not mean we have to change our existing code.

At least this is how I see things.

In general, the Pydantic model is designed with future usability in mind. See e.g. `validate_platform_info` in `node_attributes.py`. This method moves platform-dependent data in a hierarchical structure because it will be a lot easier and less error-prone to use than concatenating strings dynamically to obtain dictionary keys.

Are you sharing my goal of preserving this idea?

beckermr: Oh hmmm. I was looking through the model testing and json blobs that were not marked otherwise as bad but had missing keys were failing. Did you see that or maybe that is new?

ytausch: Hmm, I just checked out revision 127df4d9 (before you did the Pydantic model changes), ran it on the current graph and found that all relevant errors were one of two types:

- The `version` field must match the `package.version` field in the `meta_yaml` field.
- The `platform_info` field must contain all build platforms in the `platforms` field.

With one exception though, which is `version_pr_info/ucx-split` - this one had missing keys. So I am fine with setting a default value like you did.

PR #2665, opened 2024-05-27
Replace doctr_run with our own token-preserving function

ytausch:
This PR removed `doctr_run` and the corresponding dependency from our repo.

Why?
- In the very near future, `GitCli` should be able to run commands that require token hiding in outputs. However, I cannot easily use `doctr_run` for it because I need more control over the subprocess `CompletedProcess` return object. 
- With this change, I have `run_command_hiding_token` which is equivalent to `doctr_run` as well as `print_subprocess_output_strip_token`, allowing me to run the subprocess with token hiding logic manually.
- doctr is quite a big dependency just to get a 10-line function, so probably it's a good idea to remove it anyway.

### DIFF
--- a/conda-lock.yml
+++ b/conda-lock.yml
@@ -3085,51 +3085,6 @@ package:
     version: 2.6.1
   - category: main
     dependencies:
-      cryptography: ''
-      python: '>=3.5'
-      pyyaml: ''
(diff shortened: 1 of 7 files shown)


beckermr (approved):
Thank you!

PR #2698, opened 2024-06-06
BUG make sure to add newline to CBC

beckermr:

Hopefully this fixes #2696 

cc @minrk

### DIFF
--- a/conda_forge_tick/migrators/cstdlib.py
+++ b/conda_forge_tick/migrators/cstdlib.py
@@ -250,4 +250,4 @@ class StdlibMigrator(MiniMigrator):
             )
 
             with open(fname, "w") as fp:
-                fp.write("".join(cbc_lines))
+                fp.write("".join(cbc_lines) + "\n")


minrk (approved):
Thanks!

minrk:
Unfortunately, this [did not](https://github.com/conda-forge/petsc-feedstock/pull/204/files#diff-ff61408cdc05bc9667deeadb55e4aaceb1371972076b6bf6934f9008920f2bd2) fix the removal of trailing newlines.

conda_forge_tick/migrators/cstdlib.py
```diff
@@ -250,4 +250,4 @@ def migrate(self, recipe_dir, attrs, **kwargs):
             )
 
             with open(fname, "w") as fp:
-                fp.write("".join(cbc_lines))
+                fp.write("".join(cbc_lines) + "\n")
```
minrk, line 253: I think this adds a newline every time the file is rewritten, even if there is one already, so it would grow on each migration.

I also think this code also didn't have a problem stripping newlines since `cbc_lines` would have trailing newlines preserved, so I don't think it was related to #2696

PR #2778, opened 2024-06-11
Readd git backend

ytausch:
This reapplies #2643 after #2705 reverted the change. Turns out the default connection mode must be HTTPS, not SSH.


beckermr:
Should we remove the ssh urls all together? I don't expect we'll ever use them.

ytausch:
Good point, I'll do that later

beckermr:
Shall we try to merge again and do it in another PR or do you want to do it here?

ytausch:
I've done it here, should work now

PR #2794, opened 2024-06-13
Also try mangling sdist urls to match pep625

bollwyvl:

This adds _yet another_ URL to try, namely the basically-now-ubiquitous [PEP625](https://peps.python.org/pep-0625/) form:

```
old    package-name-0.0.0.tar.gz
              V
new    package_name-0.0.1.tar.gz
```

### DIFF
--- a/conda_forge_tick/url_transforms.py
+++ b/conda_forge_tick/url_transforms.py
@@ -1,8 +1,10 @@
 import os
+import re
+
 from itertools import permutations
 
 EXTS = [".tar.gz", ".zip", ".tar", ".tar.bz2", ".tar.xz", ".tgz"]
-
+PYPI_URLS = ["https://pypi.io", "https://files.pythonhosted.org"]
 
 def _ext_munger(url):
     for old_ext, new_ext in permutations(EXTS, 2):
@@ -48,22 +50,28 @@ def _v_munger(url):
 
 
 def _pypi_domain_munger(url):
-    for old_d, new_d in permutations(
-        ["https://pypi.io", "https://files.pythonhosted.org"],
-        2,
-    ):
+    for old_d, new_d in permutations(PYPI_URLS, 2):
         yield url.replace(old_d, new_d, 1)
 
 
 def _pypi_name_munger(url):
     bn = os.path.basename(url)
-    if (
-        url.startswith("https://pypi.io")
-        or url.startswith("https://files.pythonhosted.org")
-    ) and ("{{ version }}" in bn and "{{ name" not in bn):
-        yield os.path.join(os.path.dirname(url), "{{ name }}-{{ version }}.tar.gz")
+    dn = os.path.dirname(url)
+    dist_bn = os.path.basename(os.path.dirname(url))
+    is_sdist = url.endswith(".tar.gz")
+    is_pypi = any(url.startswith(pypi) for pypi in PYPI_URLS)
+    has_version = re.search(r"\{\{\s*version", bn)
 
+    # try the original url first, as a fallback (probably can't be removed?)
     yield url
+
+    if not (is_sdist and is_pypi and has_version):
+        return
+
+    # try static PEP625 name with PEP345 distribution name (_ not -)
+    yield os.path.join(dn, '%s-{{ version }}.tar.gz' % dist_bn.replace("-", "_"))
+    # many legacy names had - which would match the name
+    yield os.path.join(dn, '%s-{{ version }}.tar.gz' % dist_bn)
 
 
 def _pypi_munger(url):
@@ -141,7 +149,9 @@ def gen_transformed_urls(url):
     url : str
         The URL to transform.
     """
-    yield from _gen_new_urls(
+    yielded = []
+
+    for new_url in _gen_new_urls(
         url,
         [
             _ext_munger,
@@ -154,4 +164,7 @@ def gen_transformed_urls(url):
             _pypi_name_munger,
             _github_munger,
         ],
-    )
+    ):
+        if new_url not in yielded:
+            yield new_url
+            yielded.append(new_url)
--- a/conda_forge_tick/url_transforms.py
+++ b/conda_forge_tick/url_transforms.py
@@ -1,6 +1,5 @@
 import os
 import re
-
 from itertools import permutations
 
 EXTS = [".tar.gz", ".zip", ".tar", ".tar.bz2", ".tar.xz", ".tgz"]
--- a/conda_forge_tick/url_transforms.py
+++ b/conda_forge_tick/url_transforms.py
@@ -61,9 +61,13 @@ def _pypi_name_munger(url):
     is_sdist = url.endswith(".tar.gz")
     is_pypi = any(url.startswith(pypi) for pypi in PYPI_URLS)
     has_version = re.search(r"\{\{\s*version", bn)
+    has_name = re.search(r"\{\{\s*name", bn)
 
     # try the original URL first, as a fallback (probably can't be removed?)
     yield url
+
+    if is_pypi and has_version and not has_name:
+        yield os.path.join(dn, "{{ name }}-{{ version }}.tar.gz")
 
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
Can we add some tests? I know the tests are thing already. :/

bollwyvl:
> add some tests? I know the tests are thing already. :/

how do we feel about moving a bunch of the `black`-violating tests to `.txt`-based fixtures?

beckermr:
What do you mean by that? W/e it is, a new PR please.

conda_forge_tick/url_transforms.py
```diff
@@ -48,23 +50,29 @@ def _v_munger(url):
...
+    has_version = re.search(r"\{\{\s*version", bn)
 
+    # try the original url first, as a fallback (probably can't be removed?)
     yield url
 
+    if not (is_sdist and is_pypi and has_version):
+        return
+
+    # try static PEP625 name with PEP345 distribution name (_ not -)
+    yield os.path.join(dn, '%s-{{ version }}.tar.gz' % dist_bn.replace("-", "_"))
```
jaimergp, line 72: I think the PEP normalization also transforms periods to underscores. Check with ruamel.yaml to confirm but in that case we will need one more yield here.

conda_forge_tick/url_transforms.py
```diff
@@ -141,7 +149,9 @@ def gen_transformed_urls(url):
     url : str
         The URL to transform.
     """
-    yield from _gen_new_urls(
+    yielded = []
```
jaimergp, line 152: I guess this could be a set.

conda_forge_tick/url_transforms.py
```diff
@@ -154,4 +164,7 @@ def gen_transformed_urls(url):
             _pypi_name_munger,
             _github_munger,
         ],
-    )
+    ):
+        if new_url not in yielded:
+            yield new_url
+            yielded.append(new_url)
```
jaimergp, line 170: And in that case this would be `.add`

beckermr:
I honestly don't get how this PR passes or works. Here is the existing tests for url transforms  w/pypi (which is way more complete than I recall!):

```
def test_url_transform_pypi():
    urls = set(list(gen_transformed_urls("https://pypi.io/{{ name }}/{{ name }}-barf")))
    assert urls == {
        "https://files.pythonhosted.org/{{ name }}/{{ name }}-barf",
        "https://files.pythonhosted.org/{{ name }}/{{ name.replace('-', '_') }}-barf",
        "https://files.pythonhosted.org/{{ name }}/{{ name.replace('_', '-') }}-barf",
        "https://pypi.io/{{ name }}/{{ name }}-barf",
        "https://pypi.io/{{ name }}/{{ name.replace('-', '_') }}-barf",
        "https://pypi.io/{{ name }}/{{ name.replace('_', '-') }}-barf",
    }

    urls = set(
        list(
            gen_transformed_urls(
                "https://pypi.io/{{ name }}/{{ name.replace('_', '-') }}-barf",
            ),
        ),
    )
    assert urls == {
        "https://files.pythonhosted.org/{{ name }}/{{ name }}-barf",
        "https://files.pythonhosted.org/{{ name }}/{{ name.replace('_', '-') }}-barf",
        "https://pypi.io/{{ name }}/{{ name }}-barf",
        "https://pypi.io/{{ name }}/{{ name.replace('_', '-') }}-barf",
    }

    urls = set(
        list(
            gen_transformed_urls(
                "https://pypi.io/{{ name }}/{{ name.replace('_','-') }}-barf",
            ),
        ),
    )
    assert urls == {
        "https://files.pythonhosted.org/{{ name }}/{{ name }}-barf",
        "https://files.pythonhosted.org/{{ name }}/{{ name.replace('_','-') }}-barf",
        "https://pypi.io/{{ name }}/{{ name }}-barf",
        "https://pypi.io/{{ name }}/{{ name.replace('_','-') }}-barf",
    }

    urls = set(
        list(
            gen_transformed_urls(
                "https://pypi.io/{{ name }}/{{ name|replace('_','-') }}-barf",
            ),
        ),
    )
    assert urls == {
        "https://files.pythonhosted.org/{{ name }}/{{ name }}-barf",
        "https://files.pythonhosted.org/{{ name }}/{{ name|replace('_','-') }}-barf",
        "https://pypi.io/{{ name }}/{{ name }}-barf",
        "https://pypi.io/{{ name }}/{{ name|replace('_','-') }}-barf",
    }

    urls = set(
        list(
            gen_transformed_urls(
                'https://pypi.io/{{ name }}/{{ name.replace("_", "-") }}-barf',
            ),
        ),
    )
    assert urls == {
        "https://files.pythonhosted.org/{{ name }}/{{ name }}-barf",
        'https://files.pythonhosted.org/{{ name }}/{{ name.replace("_", "-") }}-barf',
        "https://pypi.io/{{ name }}/{{ name }}-barf",
        'https://pypi.io/{{ name }}/{{ name.replace("_", "-") }}-barf',
    }

```

We already test at least part of the `-` to `_` transform, right?

beckermr:
Oh you are munging hard-coded names. duh. Needs tests then for sure.

bollwyvl:
> fixtures

> url transforms w/ pypi (which is way more complete than I recall!):

Made #2797. Might do a little work there before coming back to here... and indeed, this (draft) PR needs some better stress tests, especially of bad/special/magic cases that categorically _did_ change over time (PEP420 with `.`, `poetry` and `flit` epochs, etc. ).

beckermr:
We have pre-commit so it should be easy to solve linter errors.

beckermr:
pre-commit.ci autofix

bollwyvl:
I am struggling to find examples of any in-the-wild examples that normalizes `.` to `_`, or apply case normalization.

beckermr:
That's fine. As long as something made up is in the tests, it will be fine I think.

conda_forge_tick/url_transforms.py
bollwyvl: edcfef82

conda_forge_tick/url_transforms.py
```diff
@@ -48,23 +50,38 @@ def _v_munger(url):
...
         yield url.replace(old_d, new_d, 1)
 
 
 def _pypi_name_munger(url):
     bn = os.path.basename(url)
-    if (
-        url.startswith("https://pypi.io")
-        or url.startswith("https://files.pythonhosted.org")
-    ) and ("{{ version }}" in bn and "{{ name" not in bn):
-        yield os.path.join(os.path.dirname(url), "{{ name }}-{{ version }}.tar.gz")
```
beckermr, line 64: It looks like you removed this variant that we try. We should yield it too.

conda_forge_tick/url_transforms.py
bollwyvl: [4face02](https://github.com/regro/cf-scripts/pull/2794/commits/4face02266bc5b2d2035c48ee5707bb08956e037)

beckermr (changes requested):

conda_forge_tick/url_transforms.py
bollwyvl: Hm... it doesn't check to see if that variable is even set... seems like _adding_ variables when not asked could potentially make things _worse_...

beckermr: If the url doesn't work it tries the next one, so it doesn't matter. Also there is a test if name is in the url in the first place. That means it is likely there in the recipe. 

In general though, please do not remove features. I've seen this more than once in PRs others have made to the bot and I don't really understand it. 

Everything is here for a reason, even if we do not recall why. Unless we know what happened, we should leave things in place. 

Let's add this back and then we can merge.

bollwyvl: 2bea2c2f

beckermr (approved):

beckermr:
Thank you!

PR #2795, opened 2024-06-13
Readd Git backend, again

ytausch:
This should resolve the errors we observed in https://github.com/regro/cf-scripts/actions/runs/9465733947/job/26075864527

We erroneously returned the directory name of the feedstock and not the full feedstock path in `get_repo`.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -30,8 +30,8 @@ from conda_forge_tick.feedstock_parser import BOOTSTRAP_MAPPINGS
 from conda_forge_tick.git_utils import (
     GIT_CLONE_DIR,
     comment_on_pr,
-    get_github_api_requests_left,
     get_repo,
+    github_backend,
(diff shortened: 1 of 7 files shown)


beckermr:
Is there a better way to test this PR before we merge again?

ytausch:
Problem is, there are no real tests for the `auto_tick` module - and I think it's currently not worth building them because I want to change some more git-related things about it.
Running locally is also not really possible currently. So I guess we should just try again - feel free to review the code again if you think that makes sense.

beckermr:
OK. I will merge this PR when I have a chance to baby sit the bot and see what happens.

ytausch:
Just as a note: I would appreciate if we could use "rebase" instead of "merge" to keep this PR up-to-date with the base branch for now. I have three branches that all depend on this, and rebasing helps me to keep a clean history.

ytausch:
@beckermr I would appreciate if we can merge this finally.

PR #2802, opened 2024-06-19
consistently ensure trailing newline in recipe conda_build_config.yaml

minrk:
#2698 didn't fix #2696, as seen [here](https://github.com/conda-forge/petsc-feedstock/pull/204/files#diff-ff61408cdc05bc9667deeadb55e4aaceb1371972076b6bf6934f9008920f2bd2). It also adds an _additional_ trailing newline, even if there is one.

I found 3 places where trailing newlines could be mishandled:

- yaml migrator (only affects conda-forge-pinning, I think)
- cstdlib (#2698 adds an _additional_ trailing newline rather than ensuring one exists)
- mpi_pin_run_as_build - I'm pretty sure this is actually the one that was affecting the petsc feedstock, which prompted #2696 . I was able to run the migrator on petsc and it re-stripped the trailing newline.

reissue and extend #2696. GitHub didn't give me a reopen button, perhaps because the branch had been deleted.

### DIFF
--- a/conda_forge_tick/migrators/cstdlib.py
+++ b/conda_forge_tick/migrators/cstdlib.py
@@ -245,9 +245,12 @@ class StdlibMigrator(MiniMigrator):
             # that in the conda_build_config, and remove all `__osx` constraints in
             # the meta.yaml (see further up).
             # this line almost always has a selector, keep the alignment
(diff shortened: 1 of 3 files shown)


beckermr:
Thanks @minrk !

PR #2812, opened 2024-06-26
Extend new Git Backend

ytausch:
This PR builds upon #2795 to extend the git backend and use it in even more places - and completely decouples the `auto-tick` command from the old git logic, allowing its local execution.

`#2795` should be merged first. To easily compare what's new in this PR, click [here](https://github.com/ytausch/cf-scripts/compare/git-backend-attempt-3...ytausch:cf-scripts:continue-git-backend).


conda_forge_tick/auto_tick.py
```diff
@@ -177,307 +497,138 @@ def run(
...
-            gdiff = eval_cmd(
-                ["git", "diff", "--name-only", f"{head_ref.strip()}...HEAD"]
-            )
-
-            diffed_files = [
-                _
-                for _ in gdiff.split()
-                if not (
-                    _.startswith("recipe")
-                    or _.startswith("migrators")
```
ytausch, line 301: @isuruf @beckermr While designing test cases, I came across this piece of code that was introduced in #516.

What is the purpose of excluding files from a `migrators` directory when checking if a migration rerender is not a trivial MigrationYaml change (at least this is how I called it in my refactored code)? I am not aware of any feedstocks that have a top-level `migrators` directory. The closest to my understanding of this code is probably the `.ci_support/migrations` (sic!) directory, which is not considered here.

I personally would guess that this is is a mix of horrendously outdated and never-caught typo but please enlighten me!

conda_forge_tick/auto_tick.py
```diff
@@ -139,35 +145,349 @@ def _get_pre_pr_migrator_attempts(attrs, migrator_name, *, is_version):
...
+
+    recipe_dir = context.local_clone_dir / "recipe"
+    migrators_dir = context.local_clone_dir / "migrators"
+
+    nontrivial_migration_yaml_changes = any(
+        not file.is_relative_to(recipe_dir)
+        and not file.is_relative_to(migrators_dir)
+        and not file.name.startswith("README")
+        for file in changed_files
+    )
```
ytausch, line 293: @isuruf @beckermr This is where it appears in the refactored code.

conda_forge_tick/auto_tick.py
beckermr: So this does indeed look like a bug in terms of the path, however the purpose of the code is as follows. When we run the bot, we run a series of feedstock cleanup tasks on top of the main migrators. These tasks are called minimigrators or piggy-pack migrations. They make changes throughout the feedstock, including in the recipe. Additionally, the README can change for various reasons that don't involve a full rebuild. So this code is meant to make the bot not issue PRs for trivial changes introduced by the rerender.

ytausch: Yep, that's clear from my side. How should we proceed with the bug then? Remove the `migrators` directory check because we did not need it anyway so far?

ytausch: I removed it in 37809fa since the `migrators` directory check did not do anything until now, so apparently we don't need it.

beckermr: I suspect we need to use `.ci_support/migrations` instead. We can do that in a new PR though.

beckermr (changes requested):
We need to break this one up into manageable chunks around 500 lines or so. The refactors of auto_tick.py especially need to be done bit by bit slowly since the testing of that code is non-existent.

ytausch:
I'll see what I can do, probably there will also more testing for `auto_tick` this time so things don't break

ytausch:
The changes of this PR are split into different smaller PRs that are intended to be merged in their order below:

- #2872 
- #2873
- tbc

ytausch:
All of these changes have been merged in (Git-1) to (Git-6).

Issue #261, opened 2018-07-10
true integration testing


CJ-Wright:
We may need to run true integration (with a dev graph to boot) so we don't blow up the graph by accident and can run CIs with close to 100% coverage.

CJ-Wright:
It would be good to separate the actual operation (the pulling of data and updating of files) from the pushing back to repos.

viniciusdc:
@CJ-Wright  Indeed, we should really rethink this ! Is there any new development ? I can think about this if possible (I can add to my GSoC program as a form of debug/test codes)

CJ-Wright:
There weren't any recent developments on this front to the best of my knowledge

ytausch:
I want to revive this issue and came up with the following concept:

# GitHub Accounts and Repositories
For a proper integration test strategy, we must mimic the relevant GitHub accounts and repositories with which the bot interacts. I propose the following accounts ("test accounts") and repositories ("test repositories"):

- `conda-forge-bot-staging` (organization) mimics the [conda-forge](https://github.com/conda-forge) organization and will contain a selection of test feedstocks (see below how we create them)
- `regro-cf-autotick-bot-staging` (user) mimics the [regro-cf-autotick-bot](https://github.com/regro-cf-autotick-bot) account and is a test environment in which the bot will create forks of the `conda-forge-bot-staging` repositories
- `regro-staging` (organization) (named after the [regro](https://github.com/regro) account) contains a special version of the [cf-graph-countyfair](https://github.com/regro/cf-graph-countyfair) which the bot uses during testing. See below how we prepare the graph for testing.

I am aware this requires us to manage three additional GitHub entities. However, since production also uses three accounts this way, we should stick to this architecture and mirror it as closely as possible, preventing future headaches.

# Integration Test Definition
To define test cases, we use the following directory structure:

```text
definitions/
├── pydantic/
│   ├── resources/
│   │   ├── recipe_before.yml
│   │   └── ... (entirely custom)
│   ├── version_update.py
│   ├── aarch_migration.py
│   ├── some_other_test_case.py
│   └── ...
├── llvmdev/
│   ├── resources/
│   └── test_case.py
└── ...
```

As shown, there are different test cases for different feedstocks. Each test case is represented by a Python module (file). Each test case module must define a `prepare()` and a `check_after()` method.

[...]

PR #2813, opened 2024-06-26
Run `auto-tick` with a single package only

ytausch:
This PR depends on #2812 and adds an argument to the `auto-tick` command allowing it to be run with a single package only for easier debugging. `#2812` needs to be merged first.

To easily compare the changes that are new with this PR, click [here](https://github.com/ytausch/cf-scripts/compare/continue-git-backend...ytausch:cf-scripts:auto-tick-single-package)


ytausch:
Since it's still very burdensome to add a fully-fledged integration/end-to-end test for the auto-tick command, I decided to skip it for now. When we deal with #261, we might also have a look at testing the dry-run mode.

conda_forge_tick/lazy_json_backends.py
```diff
@@ -630,6 +630,17 @@ def get_all_keys_for_hashmap(name):
...
 
+def does_key_exist_in_hashmap(name: str, key: str) -> bool:
+    """
+    Check if a key exists in a hashmap, using the primary backend.
+    :param name: The hashmap name.
+    :param key: The key to check.
+    :return: True if the key exists, False otherwise.
+    """
+    backend = LAZY_JSON_BACKENDS[CF_TICK_GRAPH_DATA_PRIMARY_BACKEND]()
+    return backend.hexists(name, name)
```
beckermr, line 641: This looks wrong. I think you need `hexists(name, key)`.

conda_forge_tick/cli.py
```diff
@@ -151,11 +151,20 @@ def update_upstream_versions(
 
 
 @main.command(name="auto-tick")
+@click.argument(
+    "package",
+    required=False,
```
beckermr, line 156:
```suggestion
    required=False,
    default=None,
    type=str,
```

This will help things be a bit more explicit on what happens if nothing is supplied and what type is expected. 

We also need to add a help string.

Finally, this name should be changed from `package` to `feedstock` and it should specify if the argument is `blah-feedstock` or just `blah`.

beckermr (changes requested):
I made some initial comments. `package` needs to be changed to `feedstock` everywhere. After the existing comments are addressed, I'll take another look.

conda_forge_tick/lazy_json_backends.py
ytausch: Done

conda_forge_tick/cli.py
ytausch: I applied your suggestion consistently across all CLI commands and method signatures.
The only help string we have for arguments is the docstring below, I made sure it is explained there.

conda_forge_tick/cli.py
```diff
@@ -131,31 +131,46 @@ def make_graph(
...
-    "package",
+    "feedstock",
     required=False,
+    default=None,
+    type=str,
 )
 @pass_context
 def update_upstream_versions(
-    ctx: CliContext, job: int, n_jobs: int, package: Optional[str]
+    ctx: CliContext, job: int, n_jobs: int, feedstock: Optional[str]
```
pavelzw, line 141:
```suggestion
    ctx: CliContext, job: int, n_jobs: int, feedstock: str | None
```

tests/test_upstream_versions.py
```diff
@@ -1223,15 +1223,15 @@ def test_include_node_bad_pull_request_upstream(caplog):
...
 
     caplog.set_level(logging.DEBUG)
     update_upstream_versions(
         nx.DiGraph(),
-        package=package_name,
+        feedstock=feedstock_name,
     )
 
-    assert "Package nonexistent-package not found in graph" in caplog.text
+    assert "Feedstock nonexistent-feedstock not found in graph" in caplog.text
```
pavelzw, line 1234:
```suggestion
    assert f"Feedstock {feedstock_name}-feedstock not found in graph" in caplog.text
```

ytausch:
pre-commit.ci autofix

conda_forge_tick/auto_tick.py
```diff
@@ -1182,6 +1267,11 @@ def _remove_closed_pr_json():
 
     # at this point, any json blob referenced in the pr info is state != closed
     # so we can remove anything that is empty or closed
+    if feedstock:
+        logger.info(
+            "Since you requested a run for a specific package, we are not removing closed pr_json files."
```
pavelzw, line 1272:
```suggestion
            "Since you requested a run for a specific feedstock, we are not removing closed pr_json files."
```

pavelzw (approved):
no additional comments from my side

PR #2837, opened 2024-07-10
Add a `GithubReleases` version provider

jaimergp:

Comes from an [Element conversation](https://matrix.to/#/!SOyumkgPRWoXfQYIFH:matrix.org/$17206004316459BODch:matrix.org?via=matrix.org&via=gitter.im&via=staffchat.ethz.ch). Maintainers publish tags, and then the GH release. They only want the GH releases caught by the bot. [The `releases.atom` feed sadly lists tags along with the releases](https://github.com/orgs/community/discussions/35990). This alternative method guarantees only releases are caught by leveraging the `/releases/latest` endpoint.

I think I need to update the pydantic model for the `version_sources` enum, but I can't find it.


beckermr:
How do we manage the issue of running out of api requests? We are going to very quickly exhaust our token with this. That's why we use the atom feed in the first place.

jaimergp:
That's a public endpoint afaik. There's no API, just a HTTP redirect. See https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release.

beckermr (changes requested):
A question on tokens. Also, if in testing we exhaust the bot token, this will cause the bot to fail since it won't be able to open prs.

tests/test_upstream_versions.py
```diff
@@ -1694,3 +1696,63 @@ def test_github_version_prefix(url, version, version_prefix, tmpdir):
...
+        (
+            "https://github.com/spglib/spglib/archive/v2.3.0.tar.gz",
+            "2.3.0",
+            "2.4.0",
+            "2.5.0",
+        ),
+    ],
+)
+@flaky
+def test_github_tag_not_released(
```
jaimergp, line 1735: Since this has passed on CI now, I'd say we remove this. The moment upstream releases 2.5.0 this will start failing and it's just going to be noise.

WDYT?

beckermr (approved):

tests/test_upstream_versions.py
beckermr: yep sounds good to me!

PR #2841, opened 2024-07-12
fix: `HAVE_CONTAINERS` check

Hofer-Julian:
- The original threw an exception if `docker` was missing.
- Not all tests using `docker` used `HAVE_CONTAINERS`


beckermr:
pre-commit.ci autofix

PR #2845, opened 2024-07-16
feat: Extend feedstock functions to allow for `recipe_yaml`

Hofer-Julian:
- Some error handling is moved from `populate_feedstock_attributes` to `load_feedstock_local`
- This allowed to handle bad responses earlier and a bit less awkward
- Special case a few places which require `meta_yaml` to be present

### DIFF
--- a/conda_forge_tick/feedstock_parser.py
+++ b/conda_forge_tick/feedstock_parser.py
@@ -224,6 +224,9 @@ def populate_feedstock_attributes(
 
     from conda_forge_tick.chaindb import ChainDB, _convert_to_dict
 
+    if meta_yaml is None and recipe_yaml is None:
+        raise ValueError("Either `meta_yaml` or  `recipe_yaml` needs to be given.")
+
     sub_graph.update({"feedstock_name": name, "parsing_error": False, "branch": "main"})
 
     if mark_not_archived:
--- a/conda_forge_tick/feedstock_parser.py
+++ b/conda_forge_tick/feedstock_parser.py
@@ -530,8 +530,9 @@ def load_feedstock_local(
 def load_feedstock_containerized(
     name: str,
     sub_graph: typing.MutableMapping,
-    meta_yaml: Optional[str] = None,
-    conda_forge_yaml: Optional[str] = None,
+    meta_yaml: str | None = None,
+    recipe_yaml: str | None = None,
+    conda_forge_yaml: str | None = None,
     mark_not_archived: bool = False,
 ):
     """Load a feedstock into subgraph based on its name. If meta_yaml and/or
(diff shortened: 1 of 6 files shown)


conda_forge_tick/feedstock_parser.py
```diff
@@ -624,14 +633,13 @@ def load_feedstock(
         The sub_graph, now updated with the feedstock metadata
     """
     in_container = os.environ.get("CF_TICK_IN_CONTAINER", "false") == "true"
-    if use_container is None:
-        use_container = not in_container
```
Hofer-Julian, line 636: Since `use_container` has to `bool` this will never be run.

Hofer-Julian:
Thanks, @beckermr!
I've adapted the docker functions. Error handling didn't need to be adapted as far as I can tell, since the docker functions also call `load_feedstock_local` behind the scenes.

conda_forge_tick/feedstock_parser.py
beckermr: Ah that's a bug. The default value should be `None`. Thanks for catching!

Hofer-Julian: Okay, I've adapted the code accordingly!

beckermr (changes requested):
Have you run this code on a actual rattler build recipe? 

The block of code you skipped where it calls `parse_meta_yaml` is what produces the data from the actual rendered recipe. That data is what is used to build the graph. Even in the new recipe format some rendering is required. So you need to add the rattler-build equivalent call.

Also, this PR could use tests in several spots including 

- [ ] rattler build recipes are rendered and parsed correctly
- [ ] the changes to the docker stuff work properly

Hofer-Julian:
> Have you run this code on a actual rattler build recipe?

No, you are right I forgot to specify my intention in the PR description. In my mind, having small, contained PRs that work towards enabling that goal eases review for you.

I see that you prefer a bigger scope than I intended, so I will extend the PR accordingly.

beckermr:
I agree with the intent for small PRs and prefer small ones too. However, this pr leaves things not functional which is why I am hesitating.

Hofer-Julian:
> However, this pr leaves things not functional which is why I am hesitating.

Do you mean that this PR introduces things that are not yet used?
Or that it breaks things?

If it is the latter, could you please point me to the place you suspect is breaking?
Because my expectation is that everything just works as is.

>  Add tests that the changes to the docker stuff work properly

Which one do you mean specifically? The biggest change I see is that I add an optional flag `--recipe-yaml` to `parse-feedstock`. I thought that the best time to add tests for that is when there's actually functionality behind it.

beckermr:
Right. I meant the first thing, not the second.

And yes, I was referring to testing the new flag. 

I am happy to merge as is if you prefer. I was confused as you described above about whether or not you inteded these changes to work. Let me know what you'd like to do.

Hofer-Julian:
> Right. I meant the first thing, not the second.
And yes, I was referring to testing the new flag.

Understood

> I am happy to merge as is if you prefer. I was confused as you described above about whether or not you inteded these changes to work. Let me know what you'd like to do.

My preference would be to merge the current PR.
I will then follow up with a PR that actually builds the graph from `recipe.yaml` and adds tests.

beckermr (approved):
Thank you!

PR #2850, opened 2024-07-17
fix: pyproject.toml metadata

Hofer-Julian:
Never [uv](https://github.com/astral-sh/uv) versions (which pixi uses behind the scenes) seem to be more strict than pip here

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "conda-forge-tick"
+name = "cf-scripts"
 authors = [
     {name = "conda-forge-tick development team", email = "condaforge@gmail.com"},
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "cf-scripts"
 authors = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [project]
-name = "conda-forge-tick"
+name = "cf-scripts"
 authors = [
     {name = "conda-forge-tick development team", email = "condaforge@gmail.com"},
```
jaimergp, line 12:
```suggestion
    {name = "cf-scripts development team", email = "condaforge@gmail.com"},
```

For consistency?

Hofer-Julian: Sure :)

beckermr (changes requested):
I don't follow this change. Cf-scripts is the repo name, but the python project is conda_forge_tick. Whatever tool is forcing these to be the same is IMHO conflating two different things.

Hofer-Julian:
> I don't follow this change. Cf-scripts is the repo name, but the python project is conda_forge_tick. Whatever tool is forcing these to be the same is IMHO conflating two different things.

You are right, I should have checked more closely if the error message suggested the correct solution.
It was an error in my local `pixi.toml`.

PR #2851, opened 2024-07-17
Build graph from `recipe.yaml`

Hofer-Julian:
Tests:
- [ ] CLI tests for `parse-feedstock`
- [ ] `recipe.yaml` is rendered and parsed correctly
- [ ] Returned graph contains expected data


conda_forge_tick/migrators_types.pyi
```diff
@@ -58,7 +58,7 @@ class MetaYamlOutputs(TypedDict, total=False):
     # TODO: Not entirely sure this is right
     build: BuildRunExportsDict
 
-class MetaYamlTypedDict(TypedDict, total=False):
+class RecipeTypedDict(TypedDict, total=False):
```
beckermr, line 61: Why was this type renamed?

Hofer-Julian: In my mind, both the parsed `recipe.yaml` and `meta.yaml` would return a dictionary that conforms to the same interface. For that, the name `MetaYamlTypedDict` doesn't make too much sense. 
Or do you think that is a bad idea?

beckermr: Yeah totally fine with the change. I just didn't understand the rationale.

beckermr (commented):
You likely know this already, but the recipe needs to be rendered for all of the various different `.ci_support` files in the feedstock.

conda_forge_tick/utils.py
```diff
@@ -20,6 +20,7 @@
 import jinja2.sandbox
 import networkx as nx
 import ruamel.yaml
+from conda_recipe_v2_schema.model import SimpleRecipe
```
beckermr, line 23: This package will need to be added to conda-forge and added to the environment.yaml file here. Not sure what the status is for the first thing.

conda_forge_tick/utils.py
```diff
@@ -517,30 +527,145 @@ def parse_recipe_yaml_local(
...
         stdout=subprocess.PIPE,
-        stderr=subprocess.DEVNULL,
         text=True,
         input=text,
         check=True,
     )
     return json.loads(res.stdout)
 
 
+def _validate_parsed_recipes(
```
beckermr: One thing to be aware of here is that conda-forge has tons of invalid recipe bits. I don't see this changing even with the new format. So we cannot run validation that would make the bot error during its processing. We also need the code below to be robust to invalid recipes. It can mark the node as a parsing error if we want, but it needs to continue working. IDK what the code does right now, but I did want to make the comment before we got much further.

conda_forge_tick/utils.py
Hofer-Julian: Yes, that's what I did :)
Do I have to update the lockfile as well?

conda_forge_tick/utils.py
```diff
@@ -517,30 +527,145 @@ def parse_recipe_yaml_local(
...
+    parsed_recipes: list[dict[str, Any]],
+) -> list[SimpleRecipe]:
+    return [SimpleRecipe.model_validate(recipe["recipe"]) for recipe in parsed_recipes]
+
+
+def _parse_validated_recipes(
```
beckermr, line 572: To make this robust validation errors, we likely need to rewrite this code to work on the recipe as a dictionary and not as a model object.

conda_forge_tick/utils.py
beckermr: yes there is a script that will relock things.

conda_forge_tick/utils.py
Hofer-Julian: Thanks for giving the feedback early. I was hoping that we could error out when we encounter invalid `recipe.yaml`. If that's not an option, I will adapt the code.

> It can mark the node as a parsing error if we want, but it needs to continue working

Is there already something for that in place.

conda_forge_tick/utils.py
Hofer-Julian: Yeah, will do so.

conda_forge_tick/utils.py
beckermr: Yes the way we handle errors in downloads of the repos.

conda_forge_tick/utils.py
```diff
@@ -657,17 +653,26 @@ def _parse_validated_recipes(
...
         "requirements": requirements_data,
         "source": source_data,
         "test": None,
         "outputs": output_data,
     }
 
+    return _remove_none_values(parsed_recipes)
+
+
+def _remove_none_values(d):
```
beckermr, line 670: I wrote a small utility function to account for this weird edge case. No need to use it but I thought i'd mention in case it was helpful. 

https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/utils.py#L967

The docs are missing which is definitely my bad. The key bit is a pattern like `d.get("key", {}) or {}`. This inserts an empty dict when the key is `None`.

jezdez:
This PR title seems to bury the led that this is adding rattler-build to conda-forge?

beckermr:
Well "adding to conda-forge" is a bit misleading since full support in *all* of our tools will take a long, long time. 

It is true that with this PR, we will be using rattler-build as the current parser for the new format when building the bot's graph.

However, that is a design choice we can change later should other parsers come along.

beckermr (commented):
I noticed this doesn't handle the for_pinning option yet. That will likely need to be a future PR.

Hofer-Julian:
Thanks for the review comments during the process @beckermr, they were very helpful!

This is the current status of the PR:
- `rattler-build` is used to render the recipe, if it exists with a failure an exception is thrown, please let me know if you prefer a different behavior
- then, it is parsed to reflect the conda build's schema
- the biggest difference is that I don't fill the `test` field, according to @wolfv the two recipe formats differ too much here for that to make sense
- `.ci_support files` should be accounted for, I haven't added tests for that one though
- as you mentioned `for_pinning` is not supported yet, and will be implemented in a follow-up
- I've moved to `pathlib` for functions that I've touched, please let me know if you want me to refrain from these drive-by improvements in the future
- I haven't added specific tests for the added CLI command yet, as far as I can tell the meta-yaml also doesn't have them. If you have specific suggestions, I am happy to implement them

conda_forge_tick/utils.py
```diff
@@ -380,6 +380,300 @@ def _render_meta_yaml(text: str, for_pinning: bool = False, **kwargs) -> str:
...
 
 
+def parse_recipe_yaml(
+    text: str,
+    for_pinning=False,
+    platform=None,
+    arch=None,
+    cbc_path=None,
+    log_debug=False,
+    use_container: bool | None = True,
```
beckermr, line 390: This default should be None too. I think this is a bug I put into these functions more generally. Let's change this one and I will submit a PR fixing any others I can find.

tests/test_recipe_yaml_parsing.py
```diff
@@ -0,0 +1,30 @@
...
+    assert package_data["name"] == "ipywidgets"
+    assert package_data["version"] == "8.1.2"
+
+
+def test_parse_validated_recipes():
+    text = TEST_RECIPE_YAML_PATH.joinpath("mplb.yaml").read_text()
+    recipe_yaml_dict = parse_recipe_yaml_local(text)
+
+    text = TEST_META_YAML_PATH.joinpath("mplb.yaml").read_text()
+    meta_yaml_dict = parse_meta_yaml_local(text)
```
beckermr, line 27: We can remove `_local` here so that we test the container stuff too. Or we can add a parametrized test that does both.

beckermr (changes requested):
A few nits, only one of which should be addressed here. 

It appears the code has had inconsistent handling of parsing errors for a long time. I think we want to move to never raising and always making the graph node if we can. This is something to be done in a followup PR I think.

conda_forge_tick/utils.py
Hofer-Julian: Done

tests/test_recipe_yaml_parsing.py
Hofer-Julian: Done

PR #2856, opened 2024-07-18
Simplify `_parse_meta_yaml_impl`

Hofer-Julian:
`for_pinning` of  `_render_meta_yaml` already has the default value `False`, so the if condition didn't do anything.

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -730,10 +730,7 @@ def _parse_meta_yaml_impl(
 
     logger.debug("jinja2 environmment:\n%s", pprint.pformat(cfg_as_dict))
 
-    if for_pinning:
-        content = _render_meta_yaml(text, for_pinning=for_pinning, **cfg_as_dict)
-    else:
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
This function sometimes gets called with for pinning set to true. This change would not allow that to work.

Hofer-Julian:
> This function sometimes gets called with for pinning set to true. This change would not allow that to work.

Ah, is the idea here to override `cfg_as_dict`?

beckermr:
Cfg_as_dict doesn't have that value present iirc.

Hofer-Julian:
> Cfg_as_dict doesn't have that value present iirc.

Then I don't understand from where you are coming from, but happy to close that PR if you think it isn't correct :)

beckermr:
If you read the code, you'll see what happens. The change in this pr will break the bot and so cannot be merged.

Hofer-Julian:
> If you read the code, you'll see what happens.

I did and I don't see it. It was just a drive-by-PR, so I am closing it.

PR #2878, opened 2024-07-30
feat: Add `for_pinning` support for `recipe.yaml`

Hofer-Julian:
- Replaces and quotes `pin_subpackage` and `pin_compatible` as done with `meta.yaml`
- Adds tests for newly introduced functions
- Includes missing pieces for `cbc` support from graph PR
- Ignore ruff lint for long lines, since formatting is enforced anyway

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -382,11 +382,11 @@ def _render_meta_yaml(text: str, for_pinning: bool = False, **kwargs) -> str:
 
 def parse_recipe_yaml(
     text: str,
-    for_pinning=False,
-    platform=None,
-    arch=None,
-    cbc_path=None,
-    log_debug=False,
+    for_pinning: bool = False,
+    platform: str | None = None,
+    arch: str | None = None,
+    cbc_path: str | None = None,
+    log_debug: bool = False,
     use_container: bool | None = None,
 ) -> "RecipeTypedDict":
     """Parse the recipe.yaml.
@@ -427,6 +427,7 @@ def parse_recipe_yaml(
             for_pinning=for_pinning,
             platform=platform,
             arch=arch,
+            cbc_path=cbc_path,
             log_debug=log_debug,
         )
     else:
@@ -435,17 +436,18 @@ def parse_recipe_yaml(
             for_pinning=for_pinning,
             platform=platform,
             arch=arch,
+            cbc_path=cbc_path,
             log_debug=log_debug,
         )
 
 
 def parse_recipe_yaml_containerized(
     text: str,
-    for_pinning=False,
-    platform=None,
-    arch=None,
-    cbc_path=None,
-    log_debug=False,
+    for_pinning: bool = False,
+    platform: str | None = None,
+    arch: str | None = None,
+    cbc_path: str | None = None,
+    log_debug: bool = False,
 ) -> "RecipeTypedDict":
     """Parse the recipe.yaml.
 
@@ -496,11 +498,11 @@ def parse_recipe_yaml_containerized(
 
 def parse_recipe_yaml_local(
     text: str,
-    for_pinning=False,
-    platform=None,
-    arch=None,
-    cbc_path=None,
-    log_debug=False,
+    for_pinning: bool = False,
+    platform: str | None = None,
+    arch: str | None = None,
+    cbc_path: str | None = None,
+    log_debug: bool = False,
 ) -> "RecipeTypedDict":
     """Parse the recipe.yaml.
 
@@ -526,14 +528,17 @@ def parse_recipe_yaml_local(
         for some errors. Have fun.
     """
 
-    rendered_recipe = _render_recipe_yaml(text)
-    parsed_recipes = _parse_validated_recipes(rendered_recipe)
+    rendered_recipes = _render_recipe_yaml(text, cbc_path=cbc_path, platform=platform)
+    if for_pinning:
+        rendered_recipes = _process_recipe_for_pinning(rendered_recipes)
+    parsed_recipes = _parse_recipes(rendered_recipes)
     return parsed_recipes
 
 
 def _render_recipe_yaml(
     text: str,
-    cbc_path=None,
+    platform: str | None = None,
+    cbc_path: str | None = None,
 ) -> list[dict[str, Any]]:
     """
     Renders the given recipe YAML text using the `rattler-build` command-line tool.
@@ -542,6 +547,8 @@ def _render_recipe_yaml(
     ----------
     text : str
         The recipe YAML text to render.
+    platform : str, optional
+        The platform (e.g., 'linux', 'osx', 'win').
     cbc_path : str, optional
         The path to global pinning file.
 
@@ -551,8 +558,11 @@ def _render_recipe_yaml(
         The rendered recipe as a dictionary.
     """
     variant_config_flags = [] if cbc_path is None else ["--variant-config", cbc_path]
+    build_platform_flags = [] if platform is None else ["--build-platform", platform]
     res = subprocess.run(
-        ["rattler-build", "build", "--render-only"] + variant_config_flags,
+        ["rattler-build", "build", "--render-only"]
+        + variant_config_flags
+        + build_platform_flags,
         stdout=subprocess.PIPE,
         text=True,
         input=text,
@@ -561,7 +571,27 @@ def _render_recipe_yaml(
     return [output["recipe"] for output in json.loads(res.stdout)]
 
 
-def _parse_validated_recipes(
+def _process_recipe_for_pinning(recipes: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    def replace_name_key(d: dict[str, Any]) -> Any:
+        for key, value in d.items():
+            if isinstance(value, dict):
+                if key in ["pin_subpackage", "pin_compatible"] and "name" in value:
+                    # Create a new dictionary with 'package_name' first
+                    new_value = {"package_name": value.pop("name")}
+                    new_value.update(value)
+                    d[key] = _munge_dict_repr(new_value)
+                else:
+                    replace_name_key(value)
+            elif isinstance(value, list):
+                for item in value:
+                    if isinstance(item, dict):
+                        replace_name_key(item)
+        return d
+
+    return [replace_name_key(recipe) for recipe in recipes]
+
+
+def _parse_recipes(
     validated_recipes: list[dict[str, Any]],
 ) -> "RecipeTypedDict":
     first = validated_recipes[0]
@@ -585,6 +615,17 @@ def _parse_validated_recipes(
             "summary": about.get("summary"),
         }
     )
+    if run_exports := requirements.get("run_exports"):
+        weak = run_exports.get("weak")
+        if isinstance(weak, list) and isinstance(weak[0], dict):
+            run_exports["weak"] = [value for entry in weak for value in entry.values()]
+
+        strong = run_exports.get("strong")
+        if isinstance(strong, list) and isinstance(strong[0], dict):
+            run_exports["strong"] = [
+                value for entry in strong for value in entry.values()
+            ]
+
     build_data = (
         None
         if build is None or requirements is None
--- a/docker/run_bot_task.py
+++ b/docker/run_bot_task.py
@@ -394,6 +394,7 @@ def _parse_recipe_yaml(
     for_pinning,
     platform,
     arch,
+    cbc_path,
     log_debug,
 ):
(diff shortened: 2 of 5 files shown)


beckermr (changes requested):
Looking good! Two comments.

conda_forge_tick/utils.py
```diff
@@ -585,6 +615,17 @@ def _parse_validated_recipes(
             "summary": about.get("summary"),
         }
     )
+    if run_exports := requirements.get("run_exports"):
```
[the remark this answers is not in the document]
Hofer-Julian: Good point, I extracted this block into a function with comments.
Hope that makes things clearer.

conda_forge_tick/utils.py
```diff
@@ -435,17 +436,18 @@ def parse_recipe_yaml(
             for_pinning=for_pinning,
             platform=platform,
             arch=arch,
+            cbc_path=cbc_path,
             log_debug=log_debug,
         )
 
 
 def parse_recipe_yaml_containerized(
```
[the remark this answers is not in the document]
Hofer-Julian: Nice find. Fixed!

conda_forge_tick/utils.py
```diff
@@ -667,6 +715,42 @@ def _parse_validated_recipes(
     return _remove_none_values(parsed_recipes)
 
 
+def parse_recipe_yaml_requirements(requirements):
```
beckermr, line 718: I'd suggest using an `_` at the start of the function name here in order to not create more APIs we have ot think about. This one is clearly supposed to be private.

conda_forge_tick/utils.py
```diff
@@ -667,6 +715,42 @@ def _parse_validated_recipes(
...
+    When rendering the recipe by rattler build,
+    `requirements["run_exports"]["weak"]` gives a list looking like:
+    [
+        {
+          "pin_subpackage": {
+            "name": "slepc",
+            "lower_bound": "x.x.x.x.x.x",
+            "upper_bound": "x.x"
+          }
+        }
```
beckermr, line 731: This syntax appears to have an extra `{...}`?

conda_forge_tick/utils.py
```diff
@@ -667,6 +715,42 @@ def _parse_validated_recipes(
...
+    """
+    if run_exports := requirements.get("run_exports"):
+        weak = run_exports.get("weak")
+        if isinstance(weak, list) and isinstance(weak[0], dict):
+            run_exports["weak"] = [value for entry in weak for value in entry.values()]
+
+        strong = run_exports.get("strong")
+        if isinstance(strong, list) and isinstance(strong[0], dict):
+            run_exports["strong"] = [
+                value for entry in strong for value in entry.values()
```
beckermr, line 750: I would have expected `.keys()` here (and above) given the doc string? I also think this might interact weirdly with the parsing for pinning since it comes after and is undoing dicts.

Let's add a unit test for this function to ensure it works properly and help document it's behavior.

conda_forge_tick/utils.py
Hofer-Julian: Good point!

Done in https://github.com/regro/cf-scripts/pull/2878/commits/2716239b004f3bb74b0b8aa4d72c658932283eba

Hofer-Julian:
As usual, your reviews are much appreciated @beckermr!
I expect that I will be able to address them on Monday.

conda_forge_tick/utils.py
Hofer-Julian: Double-checked, and the output of rattler-build is indeed structured that way

conda_forge_tick/utils.py
Hofer-Julian: I've added three

Hofer-Julian:
> The code needs more tests, both unit and integration.
> 
> I think it'd be worth translating more of the recipes I test with [here](https://github.com/regro/cf-scripts/blob/master/tests/test_parse_munged_run_export.py).

Most of your comments should be addressed now.
I've added more comments, more unit tests, and logic to handle more edge cases.

I've also ported another recipe as you suggested. The others were too complicated to be ported by `conda-recipe-manager`. If there's a specific case that you want to be handled, let me know, and I will try to make a recipe for that.

tests/test_recipe_yaml/slepc_weak_strong.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+requirements:
+  run:
+    - petsc
+    - suitesparse
+  run_exports:
+    strong:
+      - ${{ pin_subpackage('slepc', upper_bound='x') }}
+    weak:
+      - ${{ pin_subpackage('slepc', upper_bound='x.x') }}
```
beckermr, line 30: The build strings and comments are important for this test as they catch an important edge case.

beckermr (changes requested):
A few more nits. We should add one test with recipe outputs.

tests/test_recipe_yaml/slepc_weak_strong.yaml
Hofer-Julian: I've re-added the comments. However, I cannot add the build strings.
This combination of `pin_subpackage` and build strings is not supported by rattler-build

wolfv: @beckermr - the way `pin_subpackage` and `pin_compatible` work in rattler-build it doesn't really allow any extra characters on the `pin_subpackage` line. We basically lazy evaluate the function at a later time in order to not have to do the crazy recursive eval of conda-build (which makes it slow). 

It might actually be something to test in `rattler-build`. I see two options to cover this case:

- we could add a new keyword argument to the function (`build: "foo_*"`)
- we could extend the parsing in a way that we parse the rest of the string as the build argument

I think the first of the two options would be much easier to implement. 

Maybe @h-vetinari has opinions because he triggered some larger re-shuffeling of the pin-function arguments :)

wolfv: For more context (@h-vetinari): the current requirement looks something like:

```yaml
- ${{ pin_subpackage(foo, max_pin="x.x") real_*  # some comment
```

conda_forge_tick/utils.py
```diff
@@ -493,6 +495,9 @@ def parse_recipe_yaml_containerized(
     if arch is not None:
         args += ["--arch", arch]
 
+    if cbc_path is not None:
+        args += ["--cbc-path", cbc_path]
```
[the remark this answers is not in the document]
Hofer-Julian: I've added a test `test_parse_cudnn_recipe_yaml`

conda_forge_tick/utils.py
```diff
@@ -680,6 +728,58 @@ def _parse_validated_recipes(
...
+        original = run_exports.get(strength)
+        if (
+            isinstance(original, list)
+            and len(original) > 0
+            and isinstance(original[0], dict)
+        ):
+            result = []
+            for entry in original:
+                if isinstance(entry, str):
+                    result.append(entry)
```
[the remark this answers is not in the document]
Hofer-Julian: Yes, it was an oversight that this line wasn't run. Fixed!

Hofer-Julian:
@beckermr summary of the commits that I added since your last review:
- Added test `test_parse_cudnn_recipe_yaml` to verify `cbc_path` works as expected
- Changed logic, so a line that should have been run is actually run
- Remove unused flags `arch` (part of platform with `rattler-build`) and `log_debug`

tests/test_recipe_yaml/slepc_weak_strong.yaml
beckermr: Ahhhh I see. We'll need something in the v2 spec for this as it is a use case we care about. Agreed that a `build` keyword is probably best.

h-vetinari: Yeah, I think a new kwarg sounds like a good idea! The previous reshuffling was about adding expressiveness (and reducing confusion because things didn't work as might be expected with the possibility to ask for contradictory things), so being able to to handle a case like [slepc](https://github.com/conda-forge/slepc-feedstock/blob/fdc04765da95151ad337532c24d72dc46932ff81/recipe/meta.yaml#L41-L44) (pin on additional constraints, but not as tight as `exact=True`) without workarounds sounds like a good addition to me.

How would we write this? 
```
  run_exports:
    - if: cuda_compiler_version == "None"
      then: ${{ pin_subpackage('slepc', upper_bound='x.x', build=scalar ~ '_*') }}
      else: ${{ pin_subpackage('slepc', upper_bound='x.x', build='cuda' ~ cuda_major ~ '_*') }}
```
Both `scalar` & `cuda_major` are variables; one from CBC, one derived from `cuda_compiler_version`. I'm assuming they are in context for when the function call happens? Or can we define a variable that corresponds to the build string (depending on the various combinations) and then just do:
```
  run_exports:
    - ${{ pin_subpackage('slepc', upper_bound='x.x', build=some_variable) }}
```

I think repeated application of `- if X: then: Y` makes the recipe pretty unreadable, so I'd definitely want to avoid a quadruple-if just for that run-export. Sidenote: are multi-line `then:` & `else:` clauses possible? I'm assuming they pytorch example here is a mechanical translation, because 
```
        - if: "cuda_compiler_version != \"None\""
          then: cudnn
        - if: "cuda_compiler_version != \"None\""
          then: nccl
        - if: "cuda_compiler_version != \"None\""
          then: magma
```
should really be
```
        - if: cuda_compiler_version != "None"
          then: cudnn
                nccl      # how do I write this?
                magma
```

wolfv: Yes, multi-line if-else – as you've typed – is possible. `conda-recipe-manager` doesn't do it nicely (for humans) yet.

Hofer-Julian: Personally, I don't mind the discussion here, but I wonder if we should open a new issue for this at https://github.com/prefix-dev/rattler-build so we can find it again later on

wolfv: I just created an issue: https://github.com/prefix-dev/rattler-build/issues/1012

docker/run_bot_task.py
```diff
@@ -503,30 +501,22 @@ def parse_meta_yaml(
     "--platform",
     type=str,
     default=None,
-    help="The platform (e.g., 'linux', 'osx', 'win').",
+    help="The platform (e.g., 'linux-64', 'osx-arm64', 'win-64').",
```
beckermr, line 504: Just an FYI in conda platform is linux, osx, win and arch is x86_64 (aka 64), arm64, ppc64le, etc. IDK what rattler-build is doing, but the doc string and usage here is confusing. I'd rename this option to `--platform-arch` if that is how it is supposed to be used.

beckermr: We should also double check the code to ensure the correct value is being passed here.

conda_forge_tick/utils.py
```diff
@@ -524,9 +509,7 @@ def parse_recipe_yaml_local(
     for_pinning : bool, optional
         If True, render the recipe.yaml for pinning migrators, by default False.
     platform : str, optional
-        The platform (e.g., 'linux', 'osx', 'win').
-    arch : str, optional
-        The CPU architecture (e.g., '64', 'aarch64').
+        The platform (e.g., 'linux-64', 'osx-arm64', 'win-64').
     cbc_path : str, optional
         The path to global pinning file.
     log_debug : bool, optional
```
beckermr, line 515: The doc strings need to be updated for removed options.

conda_forge_tick/utils.py
```diff
@@ -524,9 +509,7 @@ def parse_recipe_yaml_local(
     for_pinning : bool, optional
         If True, render the recipe.yaml for pinning migrators, by default False.
     platform : str, optional
-        The platform (e.g., 'linux', 'osx', 'win').
-    arch : str, optional
-        The CPU architecture (e.g., '64', 'aarch64').
+        The platform (e.g., 'linux-64', 'osx-arm64', 'win-64').
```
beckermr, line 512: See my comment on platform vs arch below. We should call this one `platform_arch` if this is what we intend or similar.

beckermr (changes requested):
A few more items. Sorry for the slow drip of PR comments and review here. The bot code base may confuse platform vs arch in other spots. Let's fix this one and I will make an issue to check the rest of the repo.

Hofer-Julian:
@beckermr I've renamed everything to `platform_arch`.
I've also adapted one spot where `platform` instead of `platform_arch` was passed.

beckermr (approved):

PR #2880, opened 2024-07-30
fix: function signature to match baseclass

Hofer-Julian:

### DIFF
--- a/conda_forge_tick/migrators/version.py
+++ b/conda_forge_tick/migrators/version.py
@@ -219,7 +219,7 @@ class Version(Migrator):
                 )
             )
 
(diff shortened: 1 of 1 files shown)


conda_forge_tick/migrators/version.py
```diff
@@ -219,7 +219,7 @@ def migrate(
                 )
             )
 
-    def pr_body(self, feedstock_ctx: FeedstockContext) -> str:
+    def pr_body(self, feedstock_ctx: FeedstockContext, add_label_text=True) -> str:
```
beckermr, line 222:
```suggestion
    def pr_body(self, feedstock_ctx: FeedstockContext, add_label_text=False) -> str:
```

beckermr (approved):

PR #2883, opened 2024-07-31
Adapt version migrator for recipe.yaml

Hofer-Julian:


conda_forge_tick/update_recipe/build_number.py
```diff
@@ -51,3 +58,14 @@ def update_build_number(raw_meta_yaml, new_build_number, build_patterns=None):
         raw_meta_yaml = "\n".join(lines) + "\n"
 
     return raw_meta_yaml
+
+
+def update_build_number_recipe_yaml(
```
beckermr, line 63: You are going to find that people parameterize the build number in the `context` section a lot and so this code is going to fail in those cases. I think you know this, but I wanted to be sure, so I thought i'd comment.

conda_forge_tick/update_recipe/version.py
```diff
@@ -379,7 +379,9 @@ def _try_to_update_version(cmeta: Any, src: str, hash_type: str):
     return updated_version, errors
 
 
-def update_version(raw_meta_yaml, version, hash_type="sha256"):
```
beckermr, line 382: The API changes in this file are going to break the admin commands here: https://github.com/conda-forge/webservices-dispatch-action/blob/main/webservices_dispatch_action/version_updater.py

OFC I wrote that integration with no versioning of the bot code base or expectations of API statbility, so the breakage is clearly my fault! 

That all being said, I think at this point we need a versioning policy for the bot code base and we need to distribute it properly. 

I've had some discussion in the past and calver was generally agreed to be the least work given the fluid nature of the bot's APIs. We will need to automate updates of services that depend on the bot's version as well.

cc @jaimergp @ytausch for viz

xref: https://github.com/regro/cf-scripts/issues/2894

beckermr: I'd honestly prefer no API changes and this is the approach I've taken in the past. However, it is very hard for me to keep this approach going given how many PRs are being made on the bot.

Hofer-Julian: I appreciate the difficulty here, and are happy to compromise on nice function names in order to avoid breakage.

I will rename this function again, and try my best, to avoid renaming others that are not private.

If you find something that I missed, please let me know.

conda_forge_tick/update_recipe/build_number.py
Hofer-Julian: Good point, nice to have it here as a todo-entry for me :)

Hofer-Julian: Done in https://github.com/regro/cf-scripts/pull/2883/commits/f8c8e66f46c5d6f0ec7f82d25c010a4649fdd904

conda_forge_tick/update_recipe/version.py
```diff
@@ -527,3 +529,28 @@ def update_version(raw_meta_yaml, version, hash_type="sha256"):
...
+
+    Returns
+    -------
+    updated_recipe_yaml : str or None
+        The updated meta.yaml. Will be None if there is an error.
+    errors : set of str
+        A set of strings giving any errors found when updating the
+        version. The set will be empty if there were no errors.
+    """
+    raise NotImplementedError()
```
Hofer-Julian, line 556: This still needs to be implemented.
`update_version` the `meta.yaml` pendant contains *a lot* of logic. One has to have a close look which parts of it are not necessary for `recipe.yaml` and how to adapt the rest that is.

tests/test_migrators.py
```diff
@@ -551,6 +553,94 @@ def run_test_migration(
...
+    cf_yml_path = tmp_path.parent / "conda-forge.yml"
+    cf_yml = cf_yml_path.read_text() if cf_yml_path.exists() else "{}"
+
+    name = yaml.safe_load(in_yaml)["package"]["name"]
+    pmy = populate_feedstock_attributes(
+        name, sub_graph={}, recipe_yaml=in_yaml, conda_forge_yaml=cf_yml
+    )
+
+    # these are here for legacy migrators
+    pmy["version"] = pmy["meta_yaml"]["package"]["version"]
```
Hofer-Julian, line 582: This function started out as a copy of `run_test_migration` the `meta.yaml` pendant. Only up to this point, I am fairly sure that nothing needs to be adapted anymore. The rest should be evaluated if it is needed or if it needs to be adapted.

conda_forge_tick/update_recipe/build_number.py
```diff
@@ -51,3 +58,28 @@ def update_build_number(raw_meta_yaml, new_build_number, build_patterns=None):
...
+    cases = [
+        ("build", "number"),
+        ("context", "build_number"),
+        ("context", "build"),
+    ]
+
+    for case in cases:
+        replace_build_number(recipe, *case)
+
+    return yaml.dump(recipe, sort_keys=False)
```
Hofer-Julian, line 85: `yaml` sorts the keys per default when deserializing. I don't think we ever want that, so let's remember to set the argument also for other cases of `yaml.dump` in this and other PRs.

Hofer-Julian (commented):
I leave these comments, so I or even better @wolfv can more easily pick up this PR later on.

Hofer-Julian:
Closing in favor of https://github.com/regro/cf-scripts/pull/2920

PR #2886, opened 2024-07-31
List paused migrators separately

### LINKED ISSUE #2885: How to disambiguate `closed` from `paused` migrators?
There's a feature request for the conda-forge.org/status page, where users would like to see "Paused migrators" separate from the actually "Closed ones". AFAIK, this information is not exposed in cf-graph-countyfair/status, so I'm looking into the necessary changes that would be needed.

My findings so far:

* `conda_forge_tick.status_report.main()` gathers the existing migrators from conda-forge-pinning and the payloads from cf-graph-countyfair.
* Then it diffs the old "total" list with the new "total" list. Anything not in the new total is considered closed and included in the "closed" JSON for two weeks.

So, if I manage to identify a `paused` migrator from [`mgconf` object here](https://github.com/regro/cf-scripts/blob/f6b95457760d93157f6cda751cbbfc2a3d7ef493/conda_forge_tick/status_report.py#L429-L433), then I can include it in a separate list and report it on its own section. Cool!

My question is how to better expose the `paused` boolean, because as far as I can tell, these are ignored in the [`migration_factory()`](https://github.com/regro/cf-scripts/blob/f6b95457760d93157f6cda751cbbfc2a3d7ef493/conda_forge_tick/make_migrators.py#L469-L470). What should I write in that hypothetical `else:` clause? Shall I add a `paused` boolean to the base `Migrator` class? Then what? 😬 

Any feedback or pointers will be welcome, thanks!

jaimergp:

Closes #2885

### DIFF
--- a/conda_forge_tick/status_report.py
+++ b/conda_forge_tick/status_report.py
@@ -351,7 +351,7 @@ def _compute_recently_closed(total_status, old_closed_status, old_total_status):
     # grab any new stuff
     closed_status = {m: now for m in set(old_total_status) - set(total_status)}
 
-    # grab anything rcent from previous stuff
+    # grab anything recent from previous stuff
     for m, nm in old_closed_status.items():
         tm = int(dateutil.parser.parse(nm.split(" closed at ", 1)[1]).timestamp())
         if m not in total_status and now - tm < two_weeks:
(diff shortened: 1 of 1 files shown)


beckermr (approved):
Lgtm! This should work.

jaimergp:
Thanks! We'll iterate on this if needed once the frontend work starts.

PR #2918, opened 2024-08-07
misc: disable recipe.yaml for migrators

nichmor:
Description:

This PR prevents situations where new feedstocks would somehow trip up migrations. 

Checklist:

Cross-refs, links to issues, etc:

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -6,6 +6,7 @@ import itertools
 import json
 import logging
 import os
+from pathlib import Path
 import pprint
 import subprocess
 import sys
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -6,7 +6,6 @@ import itertools
 import json
 import logging
 import os
-from pathlib import Path
 import pprint
 import subprocess
 import sys
--- a/conda_forge_tick/migrators/arch.py
+++ b/conda_forge_tick/migrators/arch.py
@@ -10,7 +10,6 @@ from conda_forge_tick.make_graph import (
     make_outputs_lut_from_graph,
 )
 from conda_forge_tick.migrators.core import GraphMigrator, _sanitized_muids
-from conda_forge_tick.os_utils import pushd
 from conda_forge_tick.utils import (
     as_iterable,
     frozen_to_json_friendly,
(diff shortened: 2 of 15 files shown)


conda_forge_tick/utils.py
```diff
@@ -6,6 +6,7 @@
 import json
 import logging
 import os
+from pathlib import Path
 import pprint
```
wolfv, line 10: this is a leftover?

nichmor: yes, thanks

beckermr:
Can you fill out the PR description to provide the motivation for this one? I can think of easier ways to do this and am curious what the global plan is.

wolfv:
@beckermr - we wanted to prevent a situation where new feedstocks would somehow trip up migrations. There are a couple places where the `meta.yaml` file is unconditionally loaded. 

Any hints / ideas welcome!

beckermr:
Thanks @wolfv! That makes sense. Let me think on this a bit.

wolfv:
@beckermr any new ideas? :) Sorry for the turmoil

beckermr (changes requested):
No worries on the turmoil! Thanks for the hard work. So I had a few ideas but I ended up hating all of them. Here is what we should do.

1. This PR has missed some of the migrators (the graph / ABI ones). We should fix that unless it was purposeful in which case we should discuss.

2. Let's make a single tracking issue that has a list of the skipped migrators to figure out if we will enable for v2 and if we do, we can link to the PR.

nichmor:
> No worries on the turmoil! Thanks for the hard work. So I had a few ideas but I ended up hating all of them. Here is what we should do.
> 
> 1. This PR has missed some of the migrators (the graph / ABI ones). We should fix that unless it was purposeful in which case we should discuss.
> 2. Let's make a single tracking issue that has a list of the skipped migrators to figure out if we will enable for v2 and if we do, we can link to the PR.

Hey @beckermr ! sorry if I'm being slow - could you please point me where it is the graph or abi migrator located?

beckermr:
It is in the file called `core.py` in the same module. There appear to a lot of other migrators we have missed. Are those OK? We need to check all of them.

wolfv:
@beckermr - I searched but didn't find anything in `core.py`.

beckermr (changes requested):
So I don't get why this PR doesn't just add statements in the filter functions to bypass v1 recipes.

PR #2920, opened 2024-08-07
add v2 recipe editing capabilities for bumping version & build number

wolfv:
Hi @beckermr - this is the initial PR to port over the functionality for the recipe editing for v2 recipes. 

I'll add some tests now.


wolfv:
@beckermr - I ported all recipe editing here. It's a bit involved because the tool does the following:

- evaluate the variables in the context using Jinja
- find all the sources, irregardless of `if/else` branches
- render the version _with_ the variables loaded from the context
- edit the recipe to insert the new hashes & version

lmk what you think! :)

beckermr:
Yep. If you look at the existing code it detects a bunch of errors and outputs an error message etc. we need to match that api.

beckermr:
Also the hashing bits should use the util in the repo. We've had issues with hashing stalling for some URLs and killing the bot. I'll leave more comments later. Sorry for the terse comments.

wolfv:
@beckermr - I just tackled the API of `update_version`, will make sure that the build number update also matches. Reused the existing hashing function.

conda_forge_tick/migrators/version.py
```diff
@@ -195,23 +195,31 @@ def migrate(
...
-                with open("meta.yaml", "w") as fp:
-                    fp.write(updated_meta_yaml)
-                self.set_build_number("meta.yaml")
+            if len(errors) == 0 and updated_meta_yaml is not None:
+                meta_yaml.write_text(updated_meta_yaml)
+                self.set_build_number(meta_yaml)
 
-            return super().migrate(recipe_dir, attrs)
-        else:
+                return super().migrate(recipe_dir, attrs)
```
beckermr, line 214: This return clause needs be added for v2 or combined into a single set of statements at the end of the method.

conda_forge_tick/update_recipe/v2/build_number.py
```diff
@@ -0,0 +1,61 @@
...
+    Returns:
+    --------
+    The updated recipe as a string.
+    """
+    data = _load_yaml(file)
+    build_number_modified = _update_build_number_in_context(data, new_build_number)
+    if not build_number_modified:
+        _update_build_number_in_recipe(data, new_build_number)
+
+    return _dump_yaml_to_str(data)
```
beckermr, line 61: People can and have mixed setting build numbers in both the jinja2 vars / context and explicitly in the recipe. I know it is awful. We should update the code to do both and check if the build number in the non-context part is templated or not before updating.

conda_forge_tick/update_recipe/v2/context.py
```diff
@@ -0,0 +1,19 @@
+import jinja2
+
+
+def load_recipe_context(
+    context: dict[str, str], jinja_env: jinja2.Environment
+) -> dict[str, str]:
+    """
+    Load all string values from the context dictionary as Jinja2 templates.
+    Use linux-64 as default target_platform, build_platform, and mpi.
```
beckermr, line 9: I'm concerned that assuming linux as the default will result in invalid updates for some recipes.

conda_forge_tick/update_recipe/v2/jinja/filters.py
```diff
@@ -0,0 +1,28 @@
...
+    major = parts[0] if len(parts) > 0 else ""
+    minor = parts[1] if len(parts) > 1 else ""
+    return f"{major}{minor}"
+
+
+def _bool(value: str) -> bool:
+    return bool(value)
+
+
+def _split(s: str, sep: str = " ") -> list[str]:
```
beckermr, line 26: Setting this default to `None` would match standard python behavior. IDK if we should do that or not, but it seems like it might be a good idea.

conda_forge_tick/update_recipe/v2/jinja/jinja.py
```diff
@@ -0,0 +1,94 @@
...
+    Target platform, build platform, and mpi are set to linux-64 by default.
+    """
+    env = SandboxedEnvironment(
+        variable_start_string="${{",
+        variable_end_string="}}",
+        trim_blocks=True,
+        lstrip_blocks=True,
+        autoescape=jinja2.select_autoescape(default_for_string=False),
+        undefined=_MissingUndefined,
+    )
```
beckermr, line 43: Can you add a comment on these options, why they are set, and how that relates to the supported features for the v2 recipe?

Also, will this parser fail on `if` conditions, `for` loops, or other non-supported jinja2 filters / syntax?

conda_forge_tick/update_recipe/v2/jinja/jinja.py
```diff
@@ -0,0 +1,94 @@
...
+    return env
+
+
+def load_recipe_context(
+    context: dict[str, str], jinja_env: jinja2.Environment
+) -> dict[str, str]:
+    """
+    Load all string values from the context dictionary as Jinja2 templates.
+    Use linux-64 as default target_platform, build_platform, and mpi.
```
beckermr, line 84: Again I am concerned about the defaults here for platform etc.

conda_forge_tick/update_recipe/v2/source.py
```diff
@@ -0,0 +1,50 @@
...
+        return
+
+    outputs = visit_conditional_list(outputs, None)
+    for output in outputs:
+        sources = output.get("source", None)
+        sources = typing.cast(ConditionalList[Source], sources)
+        if sources is None:
+            continue
+        source_list = visit_conditional_list(sources, None)
+        yield from source_list
```
beckermr, line 50: Somehow it escaped me that outputs can have their own sources now. That will be fun.

conda_forge_tick/update_recipe/v2/version.py
```diff
@@ -0,0 +1,135 @@
...
+        return f"{self.hash_type}: {self.hash_value}"
+
+
+def _has_jinja_version(url: str) -> bool:
+    """Check if the URL has a jinja `${{ version }}` in it."""
+    pattern = r"\${{\s*version"
+    return re.search(pattern, url) is not None
+
+
+def update_hash(source: Source, url: str, hash_: Hash | None) -> None:
```
beckermr, line 53: Some people will set the hash in the context section. yay edge cases!

conda_forge_tick/update_recipe/v2/version.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if not _has_jinja_version(url):
+            continue
+
+        template = env.from_string(url)
+        rendered_url = template.render(context_variables)
+        try:
+            update_hash(source, rendered_url, hash_)
+        except HashError:
+            return None, set(f"Could not retrieve hash for '{rendered_url}'")
```
beckermr, line 133: Instead of an early return, the bot should collect all of the errors and return them once. These error strings appear on the status page.

conda_forge_tick/update_recipe/v2/version.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __init__(self, message: str = "Could not update version") -> None:
+        self.message = message
+        super().__init__(self.message)
+
+
+class HashError(Exception):
+    def __init__(self, url: str) -> None:
+        self.url = url
+        self.message = f"Could not hash {url}"
```
beckermr, line 34: It is useful to add the templated string that was rendered and what version in that string was tried to this error so that folks can diagnose issues.

conda_forge_tick/update_recipe/v2/version.py
```diff
@@ -0,0 +1,135 @@
...
+    for source in get_all_sources(data):
+        # render the whole URL and find the hash
+        if "url" not in source:
+            continue
+
+        url = source["url"]
+        if isinstance(url, list):
+            url = url[0]
+
+        if not _has_jinja_version(url):
```
beckermr, line 125: URLs can have other templated items that change (like version subparts) and so will need to updated.

conda_forge_tick/update_recipe/v2/yaml.py
```diff
@@ -0,0 +1,22 @@
+import io
+from pathlib import Path
+
+from ruamel.yaml import YAML
+
+yaml = YAML()
+yaml.preserve_quotes = True
+yaml.width = 4096
```
beckermr, line 8: This is pretty big. The bot uses 320 elsewhere - https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/recipe_parser/_parser.py#L46

conda_forge_tick/update_recipe/v2/version.py
```diff
@@ -0,0 +1,135 @@
...
+            continue
+
+        template = env.from_string(url)
+        rendered_url = template.render(context_variables)
```
beckermr, line 129: The bot tries a series of url template modifications in order to get version updates to work. People do things like add / remove `v`, there are multiple pypi urls etc. We need to use those here.

beckermr (changes requested):
I left a bunch of comments. One more is that the tests are not complete in that they don't touch a lot of the edge cases. 

This PR is getting pretty big, so my suggestion is that we split to three or four PRs: 1) add the parser + tests for it, 2) the build number bump, 3) the version change + basic tests, and 4) version tests for edge cases.

conda_forge_tick/update_recipe/v2/context.py
wolfv: I understand where you are coming from but we do ignore the selectors in the version updating code (e.g. we go down both branches, `then` and `else`.

However, we do use the value for templating, so people could theoretically do something like 

```yaml
source:
  url: https://foo.${{ "linux" if linux else "bla" }}/...
  sha256: ${{ "foo" ...
```

But that doesn't seem very likely and also buggy so I don't think we'll see that a lot in the wild.

conda_forge_tick/update_recipe/v2/jinja/filters.py
wolfv: We should match the `minijinja` behavior which is to split on any whitespace (and also the same as the default in Python I think). So that's a good point! Thanks!

conda_forge_tick/update_recipe/v2/source.py
wolfv: Wasn't this always the case? I think it was always the case.

conda_forge_tick/update_recipe/v2/version.py
wolfv: Yeah it's pretty discouraged by the schema right now and I don't see the benefit for the users to set the hash in the context. We just inject it in the source and the old hash will persist in the context for now. 
If someone gets annoyed we can modify this.

conda_forge_tick/update_recipe/v2/jinja/jinja.py
beckermr: Is there a way we can make it fail and output an error?

conda_forge_tick/update_recipe/v2/yaml.py
wolfv: I was under the impression that we would never want word wrapping where it wasn't done by the user, but sure, can use the suggestion :)

conda_forge_tick/update_recipe/v2/version.py
wolfv: Yeah, that would all be done in the `context` or in the rendering process. There is no functional difference here.

wolfv:
Thanks for the comments @beckermr - I'll try to get to it asap.

conda_forge_tick/update_recipe/v2/version.py
beckermr: Many recipes have more than once source.

conda_forge_tick/update_recipe/v2/version.py
beckermr: This could break some recipes and people will be even more annoyed by that. We need to fix it.

conda_forge_tick/update_recipe/v2/yaml.py
beckermr: I tuned this carefully after several complaints to do enough wrapping to make recipes legible, but not enough to annoy folks. :)

conda_forge_tick/update_recipe/v2/version.py
wolfv: true! let me double check that case.

conda_forge_tick/update_recipe/v2/version.py
wolfv: Wait, you mean "people" as in upstream maintainers doing that?

beckermr: I mean upstream changes the urls and the bot tries to account for it. See here: https://github.com/regro/cf-scripts/blob/master/conda_forge_tick/url_transforms.py

beckermr:
No rush!

conda_forge_tick/update_recipe/v2/version.py
wolfv: Which also lets me wonder if we need to adjust the code somewhere else where the different versions are being tried out :)

conda_forge_tick/update_recipe/v2/jinja/filters.py
wolfv: Done!

conda_forge_tick/update_recipe/v2/context.py
wolfv: I believe that doesn't work in the context section. However you _can_ use templating and inline Jinja expressions, such as:

```
context:
  version: "1.2.3"
  version_string: "v${{ version if linux else "foo" }}"
```

conda_forge_tick/update_recipe/v2/version.py
beckermr: The version ticker thing does this too, but it does not record which url variant worked. Hence it is done here again. It is possible that the recipe can change between when the version is found and when we update the recipe as well.

conda_forge_tick/update_recipe/v2/context.py
beckermr: right so we need to make this correct.

conda_forge_tick/migrators/version.py
wolfv: thanks, fixed.

conda_forge_tick/update_recipe/v2/context.py
wolfv: I think it's bad practice in many ways. 

Can I instead add a lint on the sha256 hash being in the proper spot (NOT in context)?
Then people would not / could not do many shenanigans with the URL either.

beckermr: That won't change other items that might be platform dependent in the context section.

wolfv: But the SHA hash will make it impossible, thankfully :)

beckermr: Nope. Folks could define parts of urls in the context section that are platform dependent that are not hashes. So when the bot comes through to render those urls and parts, it will compute the wrong hash since it might have a valid, but incorrect url for that platform.

wolfv: @beckermr - i sorta get where you come from but is it really necessary to care about all of these edge cases on day one?

It is also not a very clever way to write the recipe. You would have to "double-if" the source.

E.g.

```yaml
context:
  foo: ${{ "bla" if linux else "foo" }}

source:
  - if: linux
    then:
      url: ${{ foo }}
      sha256: 1abcd
    else:
      url: ${{ foo }}
      sha256: 2bcde
```

So I don't think many people would wanna write their sources like this (doesn't make much logical sense).

conda_forge_tick/update_recipe/v2/version.py
wolfv: I made a lint to prevent this in conda-smithy: https://github.com/conda-forge/conda-smithy/pull/2028

Still needs tests etc. but I think that's the better way forward.

conda_forge_tick/update_recipe/v2/context.py
beckermr: Yes we should care. My suggested procedure for this work was to abstract the current version update algorithm to be independent of the recipe format backend and then use that. It should handle most, if not all, of the cases I am bringing up.

conda_forge_tick/update_recipe/v2/version.py
wolfv: @beckermr how would it break a recipe though. You would have:

```
context:
  sha256: foo

source:
  url: bla
  sha256: "1234..."
```

At some point someone would come and clean up the left-over hash from the context.

It would only break if the SHA hash was part of the URL. But in that case, the autotick updater has basically no chance anyways :)

wolfv:
I wanted to quickly type out what a v2 recipe looks like:

```yaml
context:
  version: "0.9.6"

package:
  name: bottom
  version: ${{ version }}
  
source:
  url: https://github.com/ClementTsang/bottom/archive/refs/tags/${{ version }}.tar.gz
  sha256: 202130e0d7c362d0d0cf211f6a13e31be3a02f13f998f88571e59a7735d60667

build:
  script:
    - if: unix
      then:
        - cargo install --locked --root $PREFIX --path . --no-track
      else:
        - cargo install --locked --root %PREFIX% --path . --no-track

requirements:
  build:
    - ${{ compiler("rust") }}

tests:
  - script:
      - btm --help 
```

Now if you had multiple sources, or conditional sources, you are still forced to spell them out. E.g.

```yaml
source:
  - if: linux
    then:
      url: https://github.com/ClementTsang/bottom/archive/refs/tags/${{ version }}-linux.tar.gz
      sha256: 202130e0d7c362d0d0cf211f6a13e31be3a02f13f998f88571e59a7735d60667
  - if: win
    then:
      url: https://github.com/ClementTsang/bottom/archive/refs/tags/${{ version }}-win.tar.gz
      sha256: 8a86c4eecf12446ff273afc03e1b3a09a911d0b7981db1af58cb45c439161295
  - if: osx
    then:
      url: https://github.com/ClementTsang/bottom/archive/refs/tags/${{ version }}-osx.tar.gz
      sha256: 688787d8ff144c502c7f5cffaafe2cc588d86079f9de88304c26b0cb99ce91c6
```

I looked at the code for the selector evaluation and it makes a lot of sense for meta.yamls'. 

But for the new recipe format it doesn't really make sense. The use case doesn't really exist because people wouldn't write it that way. You wouldn't even really use any platform specific jinja variables in the URLs etc. since you can just insert the proper URL (just include the version).

So IMO we can ignore this for the new recipes. I also do not (really not) see a good use case for templating the SHA hash.

Can someone come up with a good use case for templating that hash (for the new recipe format)? Maybe I am blind! :)

wolfv:
Regarding the version variations - that makes sense. We only support a subset of what is being tried in the current munger but that should be fine. I'll try to get to it. There is also an item about the R-version needing some slight adjustment that we should handle.

beckermr:
Yeah it's not really about what you think is a good or bad use of the format. If it is allowed, it will happen and the bot will have to deal with it. :/

mgorny:
Hope you don't me asking but where are we standing here? Is there anything I could do to help here? I'm new to this codebase, but I think I can spend some time on it; just don't want to step on anyone's toes.

wolfv:
Hi @mgorny - sorry for the slow response! 

This was working, but Isuru and Matt wanted to cover a number of special cases (e.g. templated SHA256 hashes).

I have the following prototype to render out the sources here: https://github.com/prefix-dev/rattler-build-conda-compat/pull/57

The idea is:

- we render against all variant configs and extract all URL sources
- check for updates and replace URLs and sha256 hashes as "text replacements" so that any comments stay intact, etc.

If you have some cycles to spend on this, that would be amazing. I added some tests in the PR linked that should show the URL source extraction to be working.

mgorny:
Do I understand correctly that you're suggesting that I take the prototype code from https://github.com/prefix-dev/rattler-build-conda-compat/pull/57 and try to integrate it into this pull request? Or do you have something else in mind?

ytausch:
Is this superseded by #3525?

wolfv:
yep!

PR #2924, opened 2024-08-10
Enable jpeg turbo for new recipes

wolfv:
Making a new PR for some changes to make Jpeg Turbo Migrator work with the new format.


nichmor:
I see that in this PR, there are also disabled migrators that are disabled here: https://github.com/regro/cf-scripts/pull/2918

what PR we should keep ?

wolfv:
yes, this is sitting on top of the other PR :)

beckermr (changes requested):
This needs review and clean up before merging.

PR #2949, opened 2024-08-23
feat: disable dependency checking by default

### LINKED ISSUE #2929: Dependency checker far from correct, or at least not helpful.
In this comment https://github.com/conda-forge/cds-crtools-feedstock/pull/83#issue-2456948976, the dependency checker complains about a large number of dependencies, most of which are produced in the package itself.   

The handful of flagged external packages, root_base and nds2_client in particular, are dependencies of at least some of the packages produced by the feedstock.

The feedstock creates a large number of packages.  Perhaps the bot is complaining that some of them don't have these dependencies, but if so, that isn't mentioned anywhere in the comment.

beckermr:
#### Description:

This PR disables dependency checking by default for version migrations. Folks can still opt-in if they want.

#### Checklist:

closes #2929

### DIFF
--- a/conda_forge_tick/migrators/version.py
+++ b/conda_forge_tick/migrators/version.py
@@ -326,7 +326,7 @@ class Version(Migrator):
             feedstock_ctx.attrs,
             ["conda-forge.yml", "bot", "inspection"],
             {},
-            "hint",
+            "disabled",
         )
         logger.info("bot.inspection: %s", update_deps)
 


ytausch:
From a user perspective, this leads to many people not seeing the dependency update info anymore since most users will not enable the dependency checking manually.

What is the rationale behind this? I understand that the dependency checks produce erroneous output in #2929, in a feedstock that is far from trivial. In my opinion, the better solution would be to disable the version checks for this single feedstock.

I found the default grayskull hints helpful for the average feedstock.

beckermr:
The dep updates are incorrect enough that they are more noise than signal now. I don't think it's worth our time to bother folks with this info.

ytausch:
So you saw incorrect hints for different feedstocks too? The PR text already says that the inspection might be inaccurate.

ytausch:
I just want to make sure that it's clear to us that disabling this feature by default is somewhat equal to removing it, as no one will go through the `conda-forge.yml` docs to find out how to enable it.

beckermr:
Yep. I am aware. It is just too flaky right now. Maybe we can improve it later. A lot of the false positives are due to how we deal with namespace packages and testing / dev code.

PR #2954, opened 2024-08-28
feat: use new action

beckermr:
#### Description:

This PR moves the relock job to the new GHA.

#### Checklist:


ytausch:
Since you moved the relock script to the other repo, what is the recommended way to relock locally? Having a local copy of the `beckermr/relock-conda` repo and running the script from there?

Issue #2961, opened 2024-08-29
Migrator bot opens PRs ignoring number of children?


h-vetinari:
There are open PRs in the [3.13](https://conda-forge.org/status/migration/?name=python313) migration with 0(!) children, while feedstocks with >2000 children are awaiting PR. This seems quite suboptimal as a strategy, if we don't tackle the biggest bottlenecks the earliest.

Bottom of the table filtered to "In PR" currently looks like this:

![image](https://github.com/user-attachments/assets/b2a3014e-dab0-42bd-bf99-b96e247b52f2)

h-vetinari:
Actually, it seems to disregard dependencies at least in some cases? It opened a [PR](https://github.com/conda-forge/numpy-feedstock/pull/329) for numpy without ever opening a PR for cython, even though numpy depends on that explicitly.

CC @beckermr

jakirkham:
It's worth noting we needed to pause the migrator<sup>1</sup> due to an absence of `cross-python`<sup>2</sup>

1: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6361
2: https://github.com/conda-forge/cross-python-feedstock/pull/79

h-vetinari:
I know, but that has no bearing on this issue.

beckermr:
Yeah I have no idea here. Someone will have to load the thing and look at the graph. I won't get to it for a while, but feel free to go ahead.

beckermr:
So numpy is marked as part of a cycle in the graph for python. See this [file](https://raw.githubusercontent.com/regro/cf-graph-countyfair/master/migrators/f/0/2/b/1/python313.json) and this entry

```json
 "cycles": {
   "__set__": true,
   "elements": [
    "fenics-basix",
    "fenics-basix-meta",
    "gnuradio",
    "gnuradio-soapy",
    "heavydb-ext",
    "intel-compiler-repack",
    "numpy",
    "omniscidb",
    "pyopencl"
   ]
  },
```

Feedstocks in cycles always go first no matter what. The surprising thing is that the solver checks passed.

beckermr:
The bot also skips solver checks for nodes in cycles. So this PR is expected.

beckermr:
The final bit here is that the bot only opens PRs for things it can solve. So even if it tried a feedstock with more downstream packages, if that one does not solve, it will not make the PR. This explains the rest of the things. I am going to close this issue. Feel free to reopen!

h-vetinari:
[...]

PR #2984, opened 2024-09-03
fix: do not set bad for rerender errors

beckermr:
#### Description:

This PR fixes a bug where rerender errors were making nodes be flagged as bad. It happened because we moved rerendering to containers so the error string changed.

#### Checklist:

xref: #2961


beckermr:
pre-commit.ci autofix

PR #2992, opened 2024-09-04
Link to migration on conda-forge.org in PR description

duncanmmacleod:
#### Description:

This PR modifies the PR description for migrations to include a link to the migration page on conda-forge.org. This may help (will help me) maintainers understand the status of the whole migration and where their package update sits in the grand scheme of things.

#### Checklist:

### DIFF
--- a/conda_forge_tick/migrators/migration_yaml.py
+++ b/conda_forge_tick/migrators/migration_yaml.py
@@ -285,10 +285,11 @@ class MigrationYaml(GraphMigrator):
 
     def pr_body(self, feedstock_ctx: ClonedFeedstockContext) -> str:
         body = super().pr_body(feedstock_ctx)
+        url = f"https://conda-forge.org/status/migration/?name={self.name}"
         if feedstock_ctx.feedstock_name == "conda-forge-pinning":
             additional_body = (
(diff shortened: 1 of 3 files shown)


conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -285,10 +285,11 @@ def migrate(
 
     def pr_body(self, feedstock_ctx: ClonedFeedstockContext) -> str:
         body = super().pr_body(feedstock_ctx)
+        url = f"https://conda-forge.org/status/migration/?name={self.name}"
```
beckermr, line 288: It is weird, but you need to use `get_migrator_name` from `conda_forge_tick.utils` here. I should clean this up in another PR.

beckermr (changes requested):
Thanks for the PR! We need one change which you would have had no way of knowing about. :(

conda_forge_tick/migrators/migration_yaml.py
duncanmmacleod: @beckermr I have made this change (hopefully correctly).

beckermr (approved):
Thank you!

PR #2994, opened 2024-09-04
feat: move locks to cf-graph-countyfair

beckermr:
#### Description:

This PR moves the lockfile to cf-graph-countyfair. This will reduce the number of bot code base releases.

#### Checklist:


ytausch:
I understand why you did this, but I'd appreciate a note about this in the README.

PR #3002, opened 2024-09-13
Version updates: report version for pending updates, remove `errored` set (redundant with `errors.keys()`)

jaimergp:
#### Description:

Update `version_status.json`:

- Turn `queued` set into a dictionary that maps the package name to the pending version update.
- Remove `errored` from the JSON payload because it's redundant with the keys of `errors`.

#### Checklist:

Part of https://github.com/conda-forge/conda-forge.github.io/issues/2289

### DIFF
--- a/conda_forge_tick/status_report.py
+++ b/conda_forge_tick/status_report.py
@@ -67,10 +67,9 @@ def _ok_version(ver):
 def write_version_migrator_status(migrator, mctx):
     """write the status of the version migrator"""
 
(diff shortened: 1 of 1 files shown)


beckermr:
We'll need to either keep both schemas here for a bit or ensure the status page can read either schema. Otherwise there will be an outage when they mismatch for a bit.

jaimergp:
Good point. I kept the old file around. The new one will be suffixed with `.v2.json`.

PR #3003, opened 2024-09-16
(Git-5) Refactor Solvability Check in auto_tick

ytausch:
#### Description:

Building upon #3000, this PR refactors the solvability checking code in auto_tick.

#### Checklist:

- [x] Pydantic model updated or no update needed

#### Cross-refs, links to issues, etc:

#3000 needs to be merged first.

PR #3004, opened 2024-09-16
(Git-6) Use new Git Backend in Entire auto_tick flow

ytausch:
#### Description:

Building upon #3003, this replaces the remaining auto_tick logic to use the new git backend.
Notably, this PR ensures that the new backend is used for opening PRs and commenting on PRs.

#### Checklist:

#3003 should be merged first.


ytausch:
@0xbe7a I think this is worth reviewing, even with limited knowledge of the codebase

conda_forge_tick/models/pr_json.py
```diff
@@ -48,7 +52,7 @@ class PullRequestState(StrEnum):
     """
 
 
-class PullRequestInfoHead(StrictBaseModel):
+class PullRequestInfoHead(ValidatedBaseModel):
```
0xbe7a, line 55: Why did you relax this?

ytausch: In `GitHubBackend.create_pull_request`, the pr_json pydantic model is directly validated on the JSON returned by the GitHub API. However, we do not want to keep all the information the API returns, mainly since these pydantic models also correspond to the data stored in cf-graph. Hence, I relaxed this to ignore any additional data returned by GitHub.
Since the direct usage of the model also prevents the code from introducing additional extra fields to the pr_json data, it's OK that the model validation test can no longer detect such extra fields when processing the cf-graph data.

conda_forge_tick/git_utils.py
```diff
@@ -134,37 +168,68 @@ class GitCli:
...
+                git_command,
+                check=check_error,
+                cwd=working_directory,
+                stdout=subprocess.PIPE,
+                **stderr_args,
+                text=True,
             )
         except subprocess.CalledProcessError as e:
-            raise GitCliError("Error running git command.") from e
+            e.stdout = replace_tokens(e.stdout, self.__hidden_tokens)
```
0xbe7a, line 209: I'm not a huge fan of encoding the credentials in the remote URL and then censoring the logs. I think a better approach would be similar to how GitHub handles it in the checkout action, by setting up the `'http.https://github.com/.extraheader'` git config to include the authorization header.

beckermr: Yeah putting the credentials on disk is not so great too right, or did I miss something? In any case, I don't actually care either way, but if we do put credentials on disk, we'll want to ensure they get cleaned up before the python process exits. On CI, GitHub will do this for us but potentially people running locally with a token might not.

ytausch: The feedstock cloning runs within a tmpdir context anyway. So if per-repository tokens are set, they are cleaned up automatically already.
So my proposal would be to implement it as @0xbe7a suggested and remove the token censoring entirely.

ytausch:
Matt, FYI: Bela is no longer mentoring me, therefore I ping @pavelzw for an additional review.

conda_forge_tick/git_utils.py
ytausch: Sorry for the delay - this is now implemented. For now, I only set the token for pushing and clear it immediately afterwards as an additional defense-in-depth measure.

conda_forge_tick/git_utils.py
```diff
@@ -764,13 +1049,75 @@ def _sync_default_branch(self, upstream_owner: str, upstream_repo: str):
...
+        target_owner: str,
+        target_repo: str,
+        base_branch: str,
+        head_branch: str,
+        title: str,
+        body: str,
+    ) -> PullRequestDataValid:
+        debug_out = textwrap.dedent(
+            f"""
+            ==============================================================
```
pavelzw, line 1063: this is being done multiple times. Could we move this into a separate method `debug_output(title, <dict>)`?

conda_forge_tick/models/pr_json.py
pavelzw, line 1: line 114

```
    # TODO: add a serializer for this field
```

this is already done, right?

conda_forge_tick/git_utils.py
```diff
@@ -739,7 +1018,13 @@ def get_remote_url(
 
         return super().get_remote_url(upstream_owner, repo_name, connection_mode)
 
-    @lock_git_operation()
+    def push_to_repository(
+        self, owner: str, repo_name: str, git_dir: Path, branch: str
+    ):
+        logger.debug(
+            f"Dry Run: Pushing changes from {git_dir} to {owner}/{repo_name} on branch {branch}."
```
pavelzw, line 1025: see below, should we add this into a `debug_output` method?

conda_forge_tick/utils.py
```diff
@@ -1232,12 +1232,35 @@ def change_log_level(logger, new_level):
         logger.setLevel(saved_logger_level)
 
 
+@overload
```
ytausch, line 1235: These changes are not needed anymore because at most one token is hidden

tests/test_git_utils.py
```diff
@@ -762,6 +948,75 @@ def test_git_cli_clone_fork_and_branch_non_existing_remote_existing_target_dir(c
...
 
+@pytest.mark.parametrize(
+    "backend", [GitHubBackend(MagicMock(), MagicMock(), ""), DryRunBackend()]
+)
+@mock.patch(
+    "conda_forge_tick.git_utils.GitHubBackend.user", new_callable=mock.PropertyMock
+)
+@mock.patch("conda_forge_tick.git_utils.GitCli.clone_fork_and_branch")
+def test_git_platform_backend_clone_fork_and_branch(
+    convenience_method_mock: MagicMock,
```
pavelzw, line 959:
```suggestion
    clone_fork_and_branch_mock: MagicMock,
```

pavelzw (approved):

conda_forge_tick/auto_tick.py
```diff
@@ -554,67 +558,64 @@ def run(
...
-        pr_json = {
-            "state": "closed",
-            "merged_at": "never issued",
-            "id": str(uuid4()),
-        }
+        pr_data = get_spoofed_closed_pr_info()
     else:
-        # push up
+        # push and PR
+        git_backend.push_to_repository(
```
beckermr, line 575: Why are we no longer catching errors on the repo push?

conda_forge_tick/contexts.py
```diff
@@ -55,6 +55,13 @@ def git_repo_owner(self) -> str:
     def git_repo_name(self) -> str:
         return f"{self.feedstock_name}-feedstock"
 
+    @property
+    def git_href(self) -> str:
```
beckermr, line 59: Maybe `get_http_ref` here or something? `get_href` confused me when I first read it.

conda_forge_tick/git_utils.py
```diff
@@ -614,6 +815,22 @@ def does_repository_exist(self, owner: str, repo_name: str) -> bool:
...
+        self.cli.add_token(git_dir, self.GIT_PLATFORM_ORIGIN, self.__token)
+
+        remote_url = self.get_remote_url(
+            owner,
+            repo_name,
+            GitConnectionMode.HTTPS,
+        )
+        self.cli.push_to_url(git_dir, remote_url, branch)
+
+        self.cli.clear_token(git_dir, self.GIT_PLATFORM_ORIGIN)
```
beckermr, line 832: This needs to use a try...finally block for safety.

conda_forge_tick/utils.py
ytausch: Yes, this was a note for myself :D
code related to multi-token hiding removed in 29b75a07589ef0326a5ff250e70ae6d7f6599b8b

conda_forge_tick/models/pr_json.py
ytausch: jup, removed in 1a6a13ac09be998f38f6cf675d037537dfa385d5

tests/test_git_utils.py
ytausch: b030c93bcd47e7902f66fd65dbca067a0dedf8a8

conda_forge_tick/contexts.py
ytausch: agreed. 0ffe783fd9c3038043d9e4986cb83f01edcdfceb

conda_forge_tick/git_utils.py
ytausch: Done in 1a99cebbb5c491c95333344821ffd8a66b490a42.

conda_forge_tick/git_utils.py
ytausch: done in 0448fcd8ace879dc571364938603ecca7f2b168d, I called it `print_dry_run_message`

conda_forge_tick/git_utils.py
ytausch: done

conda_forge_tick/auto_tick.py
ytausch: In the old code, the try-except block looks like this:

https://github.com/regro/cf-scripts/blob/621910b1438b5ed2f3510fc72c7f692ac4e2ed5a/conda_forge_tick/auto_tick.py#L581-L601

In `push_repo`, the only place where a `github3.GitHubError` can be raised is:

https://github.com/regro/cf-scripts/blob/621910b1438b5ed2f3510fc72c7f692ac4e2ed5a/conda_forge_tick/git_utils.py#L1115

`push_to_repository` only executes `git push`, the PR is created in `create_pull_request` below, which is protected by a try-else block catching duplicate PRs. To my understanding, this should all be equivalent to the old code.

ytausch:
Thank you! I just verified that no removed function is used in the regro or conda-forge organization via a quick GitHub search.

conda_forge_tick/auto_tick.py
beckermr: ah got it. SGTM and thanks for checking!

beckermr (approved):

beckermr:
We're seeing errors in pushing to repos with this so I am going to revert.

PR #3008, opened 2024-09-24
feat: add crates.io as a source for auto version bump

mbhall88:
#### Description:

On the gitter for conda-forge someone pointed me at this script as the source for automatic version bumps. The conversation was in relation to allowing [crates.io](https://crates.io/) (Rust package repository) to be detected for auto version bumps. Crates *can* be used to build packages, but there is no auto bump functionality (as far as I can tell).

I hope what I have contributed here is okay? Not sure if I've missed anything as I am new to this codebase.

#### Checklist:

- [x] Pydantic ~~model updated~~ or no update needed


conda_forge_tick/update_sources.py
```diff
@@ -616,20 +618,6 @@ def get_version(self, url: str) -> Optional[str]:
         return latest
 
 
-class LibrariesIO(VersionFromFeed):
```
beckermr, line 619: Why was this deleted?

beckermr (changes requested):
See the comment above.

conda_forge_tick/update_sources.py
mbhall88: It doesn't seem to be used anywhere. Also, it uses a method `self.package_name()` which doesn't exist? Looking through the git history it seemed like maybe it was copy/pasted from a while ago and the code base had evolved since then without this class being updated (as it wasn't being used anywhere).

Happy to add it back though if I have this all wrong?

beckermr: You don't have it wrong per say, but let's keep the dead code around just in case someone who wanted it comes asking about it.

beckermr (approved):

mbhall88:
Fixed the failing test and added a full integration test for the new crates.io source

beckermr:
pre-commit.ci auotfix

mbhall88:
@beckermr does the version bump infrastructure get the sha256 by just downloading the new version tarball and running sha256sum on that downloaded tarball? The reason I ask is that the crates.io API provides the checksum for the tarball which seems like a better place to get it than from the downloaded tarball in case there were issues with the download correct? (I've likely got this all wrong though sorry)

beckermr:
Yes, the update code does a download and checksum. Unfortunately, we have no way to feed a new checksum into the version bump code for the recipe. Further, not all version update sources supply a checksum. So while you are of course correct, in practice we don't support this.

PR #3009, opened 2024-09-24
feat: add build number editing for v1 recipes

wolfv:


wolfv:
Note: I added a fix to be able to pass a callable for `new_build_number` in #3011 

IMO this could still be reviewed/merged separately and I could take care of managing the two PRs, as you prefer.

conda_forge_tick/update_recipe/v1/build_number.py
```diff
@@ -0,0 +1,61 @@
...
+logger = logging.getLogger(__name__)
+
+HashType = Literal["md5", "sha256"]
+
+
+def _update_build_number_in_context(
+    recipe: dict[str, Any], new_build_number: int
+) -> bool:
+    for key in recipe.get("context", {}):
+        if key.startswith("build_") or key == "build":
```
beckermr, line 20: Which specific `build_*` keys are you targeting here? We should use an allowlist to ensure we don't stomp on any key that starts with `build_*`.

beckermr (changes requested):
I left a few comments. Also the module names of `v1` and `yaml` are easy to stomp on. Let's 

- rename the `v1` module to something longer like `v1_recipes` or w/e.
- rename the yaml module to something else and possibly combine those routines into the ones in the utils module or the recipe parser module

conda_forge_tick/update_recipe/v1/build_number.py
wolfv: changed to only target `build` and `build_number`.

wolfv:
I made the changes:

- only touch `build`, `build_number` or `number` keys in context (and also only touches the first of them)
- merged changes from the other PR to allow callables for build number
- renamed to `v1_recipe` and got rid of the `yaml` module by re-using old code

Let me know if there is anything else

conda_forge_tick/update_recipe/v1/build_number.py
wolfv: I changed this.

conda_forge_tick/update_recipe/v1/yaml.py
```diff
@@ -0,0 +1,22 @@
+import io
+from pathlib import Path
+
+from ruamel.yaml import YAML
+
+yaml = YAML()
```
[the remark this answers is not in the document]
wolfv: Using the existing YAML parser thingy now.

conda_forge_tick/update_recipe/v1_recipe/build_number.py
```diff
@@ -0,0 +1,103 @@
...
+        for output in recipe["outputs"]:
+            if "build" in output and "number" in output["build"]:
+                output["build"]["number"] = new_build_number
+                is_modified = True
+
+    return is_modified
+
+
+def _load_yaml(file: Path):
+    yaml = _get_yaml_parser(typ=None)
```
beckermr, line 66:
```suggestion
    yaml = _get_yaml_parser(typ="safe")
```

beckermr (approved):

wolfv:
@beckermr - the 'safe' loader doesn't seem to be able to preserve the quotes - which one do we care about more?

wolfv:
> The initial idea behind ruamel.yaml was its [RoundTripConstructor](https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/main.py#l137) , which is also a subclass of the [SafeConstructor](https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/constructor.py#l1011). You used to get this using the now deprecated round_trip_load function and nowadays via the .load() method after using YAML(typ='rt'), this is also the default for a YAML() instance without typ argument. That RoundTripConstructor does not registers any tags or have any code that interprets tags above and beyond the normal tags that the SafeConstructor uses.

https://stackoverflow.com/a/71299116

PR #3019, opened 2024-10-11
feat: enable arch migrators for v1 recipes

wolfv:


beckermr (changes requested):
We need to figure out which of the minimigrators that get attached to the osx-arm64 PRs we need to enable as well. The list of them is here: https://github.com/regro/cf-scripts/blob/main/conda_forge_tick/make_migrators.py#L194-L202

wolfv:
Started this issue to track the work on minimigrators for v1 recipe: https://github.com/regro/cf-scripts/issues/3642

PR #3151, opened 2024-11-16
fix: try munged pypi urls with middle section left alone

beckermr:
#### Description:

I am trying to catch more pypi errored versions.

#### Checklist:


beckermr:
pre-commit.ci autofix

Issue #3191, opened 2024-11-22
Py313 migration stuck on networkx - wrongly still "Awaiting parents"?


h-vetinari:
The 3.13 [migration](https://conda-forge.org/status/migration/?name=python313) has `networkx` as the top missing dependency, yet, paradoxically, as still "Awaiting parents". I've checked the "Done" feedstocks as well for whether they're parents of `networkx` (and whether they were built without spurious CI failures), but I cannot tell where this is hanging.

The problem is that even a manual PR to `networkx` will fix this, as the bot still thinks it needs to wait for something.

CC @beckermr

beckermr:
Hrmmmm ok. We need to check the node attrs and see what they look like. Could be a v1 parsing error.

beckermr:
I think the issue is here: https://github.com/regro/cf-scripts/blob/main/conda_forge_tick/migrators/migration_yaml.py#L615

It appears that v1 recipes parsed by rattler build do not come out with the outputs build section filled in. On networkx, the output has a build section of `null` in the json and so the output is wrong assumed to not be noarch. 

What is the semantics for v1 recipes? Does the parent build section get overrides from the outputs and then get used for the output? We can add w/e logic is correct to the function.

h-vetinari:
> What is the semantics for v1 recipes? Does the parent build section get overrides from the outputs and then get used for the output? We can add w/e logic is correct to the function.

Pinging @wolfv @baszalmstra @nichmor :)

PR #3195, opened 2024-11-23
fix: ensure build section is filled in for noarch

beckermr:
#### Description:

This PR properly makes the build section for recipes when it is empty.

#### Checklist:


beckermr:
xref: #3191

PR #2459, opened 2024-04-18
add stdlib piggyback to version_migrator

h-vetinari:
This was already brought up in #2372, and discussed in today's core call as well (no objection to turning it on despite the limitation about templated output names). Since we haven't had any notable issues I can remember over the last ~2 weeks, I think it's a good time to add it to the version migrator as well now.

Also pick up a drive-by fix for making the sysroot regex pattern catch some more relevant cases.

PR #3225, opened 2024-11-27
fix: use mini-migrators for everything

beckermr:
#### Description:

This PR adds a set of default mini-migrators for every thing we can.

#### Checklist:

### DIFF
--- a/conda_forge_tick/make_migrators.py
+++ b/conda_forge_tick/make_migrators.py
@@ -56,6 +56,7 @@ from conda_forge_tick.migrators import (
     LicenseMigrator,
     MigrationYaml,
     Migrator,
+    MiniMigrator,
     MPIPinRunAsBuildCleanup,
     NoarchPythonMinMigrator,
     NoCondaInspectMigrator,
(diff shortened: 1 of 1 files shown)


beckermr:
cc @h-vetinari this might help with not missing additions in the future

conda_forge_tick/make_migrators.py
```diff
@@ -97,6 +98,29 @@
 MAX_PR_LIMIT = 50
 MAX_SOLVER_ATTEMPTS = 50
 CHECK_SOLVABLE_TIMEOUT = 90  # 90 days
+DEFAULT_MINI_MIGRATORS = [
```
h-vetinari, line 101: Thanks for this!

What was the reason to not add `StdlibMigrator` to the default set? It's in use for both types of migrators (after #2372 & #2459)

conda_forge_tick/make_migrators.py
```diff
@@ -777,22 +796,13 @@ def initialize_migrators(
...
-                DependencyUpdateMigrator(python_nodes),
-                PyPIOrgMigrator(),
-                StdlibMigrator(),
-            ],
+            piggy_back_migrations=_make_mini_migrators_with_defaults(
+                extra_mini_migrators=[
+                    PipWheelMigrator(),
+                    DependencyUpdateMigrator(python_nodes),
+                    StdlibMigrator(),
+                ],
```
h-vetinari, line 804: I see it gets readded here anyway, but that just sounds to me like it should be in the default set? 🤔

beckermr: I tried to separate formatting versus recipe behavior changes, with the goal of future additions to the migrators preferring one over the other depending on context.

h-vetinari: OK, fair enough. 🤷

PR #3259, opened 2024-12-02
fix: dump PR json right away to avoid race condition

beckermr:
#### Checklist:


conda_forge_tick/auto_tick.py
```diff
@@ -748,6 +768,7 @@ def _run_migrator_on_feedstock_branch(
                     if "PRed" not in pri:
                         pri["PRed"] = []
                     pri["PRed"].append(d)
+
```
beckermr, line 771:
```suggestion
```

PR #3326, opened 2024-12-06
Separate Bot User GitHub Tokens from regro GitHub Tokens

ytausch:
#### Description:

Our proposed integration tests for #261 need to interact with the GitHub API with multiple different identities in a staging environment. Since nowadays, fine-grained personal access tokens are [recommended by GitHub whenever possible](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#types-of-personal-access-tokens), they clearly separate access to different GitHub users/organizations.

This means the bot needs to support using different access tokens depending on which GitHub user/org it is interacting with. This PR contributes the necessary changes.

Essentially, `BOT_TOKEN` can now be filled with a fine-grained access token for `regro-cf-autotick-bot` and `REGRO_TOKEN` with a fine-grained token for `regro`.

Note that these changes do not prevent from continuing to run the bot with a classic access token in production.

I carefully reviewed the source code to verify that the correct tokens are available and used everywhere. It would be helpful if you, @beckermr, could double-check.

I leave it up to you if you also want to update the production tokens to fine-grained tokens together with these changes. It involves more risk of breaking the bot but would ensure that future source code changes correctly differentiate between the two tokens. We can also wait for the integration tests to have more assurance here. 

@pavelzw

#### Checklist:

### DIFF
--- a/.github/workflows/bot-bot.yml
+++ b/.github/workflows/bot-bot.yml
@@ -58,6 +58,7 @@ jobs:
           conda-forge-tick auto-tick
         env:
           BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
+          REGRO_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
           RUN_ID: ${{ github.run_id }}
           MEMORY_LIMIT_GB: 7
           CF_TICK_GRAPH_DATA_BACKENDS: "${{ vars.CF_TICK_GRAPH_DATA_BACKENDS }}"
@@ -79,7 +80,7 @@ jobs:
           export RUN_URL="https://github.com/regro/cf-scripts/actions/runs/${RUN_ID}"
           conda-forge-tick deploy-to-github
         env:
-          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
+          REGRO_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
           RUN_ID: ${{ github.run_id }}
 
       - name: bump on fail
@@ -88,7 +89,7 @@ jobs:
           export ACTION_URL="https://github.com/regro/cf-scripts/actions/runs/${RUN_ID}"
           python cf-scripts/autotick-bot/bump_bot_team.py
         env:
-          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
+          REGRO_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
           RUN_ID: ${{ github.run_id }}
           ACTION_NAME: ${{ github.workflow }}
 
--- a/.github/workflows/bot-bot.yml
+++ b/.github/workflows/bot-bot.yml
@@ -45,8 +45,6 @@ jobs:
         if: success() && ! env.CI_SKIP
         run: |
           source cf-scripts/autotick-bot/install_bot_code.sh
-        env:
-          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
 
       - name: run migrations
         if: success() && ! env.CI_SKIP
--- a/.github/workflows/bot-events.yml
+++ b/.github/workflows/bot-events.yml
@@ -83,6 +83,7 @@ jobs:
             --event='${{ inputs.event }}' \
             --uid='${{ inputs.uid }}'
         env:
+          BOT_TOKEN : ${{ secrets.AUTOTICK_BOT_TOKEN }}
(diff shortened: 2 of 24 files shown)


.github/workflows/bot-bot.yml
```diff
@@ -45,8 +45,6 @@ jobs:
         if: success() && ! env.CI_SKIP
         run: |
           source cf-scripts/autotick-bot/install_bot_code.sh
-        env:
-          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
```
ytausch, line 49: I removed `BOT_TOKEN` for `install_bot_code.sh` entirely everywhere because I could not find where it uses the token.

.github/workflows/bot-bot.yml
```diff
@@ -81,7 +80,7 @@ jobs:
           export RUN_URL="https://github.com/regro/cf-scripts/actions/runs/${RUN_ID}"
           conda-forge-tick deploy-to-github
         env:
-          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
```
ytausch, line 84: Explanation: This only pushes to cf-graph so the permissions for regro are sufficient

.github/workflows/bot-feedstocks.yml
```diff
@@ -54,7 +52,7 @@ jobs:
 
           conda-forge-tick gather-all-feedstocks
         env:
-          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
+          REGRO_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
```
ytausch, line 55:
```suggestion
          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
```

.github/workflows/bot-events.yml
```diff
@@ -85,7 +83,7 @@ jobs:
             --event='${{ inputs.event }}' \
             --uid='${{ inputs.uid }}'
         env:
-          BOT_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
+          REGRO_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
```
ytausch, line 86:
```suggestion
          BOT_TOKEN : ${{ secrets.AUTOTICK_BOT_TOKEN }}
          REGRO_TOKEN: ${{ secrets.AUTOTICK_BOT_TOKEN }}
```

beckermr (changes requested):
I don't follow the need for this change. I thought the plan was to put testing feedstocks in the `regro` org directly and let the bot fork them, in which case we can use one token and things are a lot simpler.

ytausch:
> I don't follow the need for this change. I thought the plan was to put testing feedstocks in the `regro` org directly and let the bot fork them, in which case we can use one token and things are a lot simpler.

I know we talked about this already so sorry if I confused you.

The test environment needs three things:
1. A copy of cf-graph-countyfair which is not the real one.
2. A bot user account which is not the real bot user account.
3. A GitHub org containing the mock feedstocks. 

It is possible to put the mock cf-graph (1) in `regro` if the permissions are restricted correctly.

Mock feedstocks or forks thereof (2/3) should not be put in `regro` or any other existing org: **The test setup needs permission to create and delete arbitrary repositories for the owner of the feedstocks and fork.** The test setup should not be able to delete production resources such as cf-graph.

ytausch:
> It seems a lot easier to start with something simpler. Making and then deleting repos for running the tests is slow and error prone. We can post some notes on how to set things up and do most of it by hand.

It is already implemented. And it's surely not a bad thing that testing is as much automated as possible. Did not see that anything is slow or raises errors.

ytausch:
Converting to draft because a GitHub app could solve this issue in another way.

ytausch:
We decided to use a classic PAT for now, although GitHub discourages their usage, making this PR redundant. There are reasons though why we want to migrate the bot to use GitHub app authentication eventually.
Let's discuss the exact distribution of repositories separately! Happy to set up a call or use the next bot sync for that. I will reach out to you as soon as we have something to show.

ytausch:
I double-checked and found that GitHub apps indeed can't be used to fork repositories into a **user account**, but I could successfully fork into an **organization** if the app is installed in an organization. It requires the original repo to be public, but this should not be a concern for us. Nothing should speak against `cf-regro-autotick-bot` being an organization, right?

I agree that the increased rate limits are the main reason why we should consider using a GitHub app.

beckermr:
The bot is going to remain a machine user and not be changed to an org.

PR #3398, opened 2024-12-13
fix: refresh before testing label

beckermr:
#### Checklist:


beckermr:
pre-commit.ci autofix

PR #3417, opened 2024-12-16
Integration tests

### LINKED ISSUE #261: true integration testing

ytausch:
#### Description:

This PR adds integration tests to the autotickbot, interacting with a real GitHub test environment. Closes #261 (more details there). This is a "minimal breakthrough" PR, only adding one test for version updates (using the pydantic-feedstock as an example).

Implementation is finished.

In this demo PR in my fork the integration tests actually run (needs token setup): https://github.com/ytausch/cf-scripts/pull/2

I've added a `CODEOWNERS` file to automatically request reviews from me if anything related to the tests change. On this occasion, I've also added a selection of other files where I am the main contributor. Hope you're fine with this!

@pavelzw

#### Checklist:

- [x] Add more documentation for the integration tests
- [x] setup tokens (see comment below) - note the checks in this PR fail without that

- needs https://github.com/conda-forge/conda-forge-feedstock-ops/pull/34

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210327008468118


ytausch:
@beckermr, as discussed, feel free to have a look here!

The GitHub entities currently used as the test environment are
- https://github.com/conda-forge-bot-staging/ (for conda-forge)
- https://github.com/regro-staging (for the cf-graph)
- https://github.com/regro-cf-autotick-bot-staging (the bot user)

Using these entities provides clear separation from the production environment. We can arrange to give you the necessary permissions everywhere / to transfer ownership if you are fine with this setup.

As discussed, we use a GitHub classic token of regro-cf-autotick-bot-staging.

ytausch:
In this demo PR in my fork the integration tests actually run (needs token setup): https://github.com/ytausch/cf-scripts/pull/2

ytausch:
Merging main into this branch led to the bot no longer opening PRs in the integration tests.
I will investigate this (maybe this is a subtle change how the different steps of the bot have to be chained together?), but I think we can already start the review process herr (the tests have been working before the merge).

conda_forge_tick/deploy.py
```diff
@@ -133,7 +140,7 @@ def _deploy_batch(*, files_to_add, batch, n_added, max_per_batch=200):
 
         status = 1
         num_try = 0
-        while status != 0 and num_try < 20 and graph_ok:
+        while status != 0 and num_try < 100 and graph_ok:
```
ytausch, line 143: This looks like something went wrong in the merge

tests_integration/step_prepare.py
```diff
@@ -0,0 +1,75 @@
...
+    scenario = get_test_scenario(scenario_id)
+
+    logging.info("Preparing test scenario %d...", scenario_id)
+    logging.info("Scenario: %s", scenario)
+
+    run_all_prepare_functions(scenario)
+
+
+if __name__ == "__main__":
+    setup_logging(logging.INFO)
```
pavelzw, line 74: it could be nice to set this to debug when debug logging is enabled in the run: you can test this by checking `$RUNNER_DEBUG`: https://github.com/actions/toolkit/blob/5e8c25d1f5d92a53f2438de763f92d4809dd666f/packages/core/src/core.ts#L234

tests_integration/setup_repositories.py
```diff
@@ -0,0 +1,222 @@
...
+    write_github_output(GITHUB_OUTPUT_KEY_SCENARIO_IDS, _format_scenario_ids(ids))
+
+
+def main():
+    prepare_all_accounts()
+    write_scenario_ids()
+
+
+if __name__ == "__main__":
+    setup_logging(logging.INFO)
```
pavelzw, line 221: see other comment

tests_integration/step_validate.py
```diff
@@ -0,0 +1,36 @@
...
+            raise AttributeError("The test case must define a validate() function.")
+
+
+def main(scenario_id: int):
+    scenario = get_test_scenario(scenario_id)
+    run_all_validate_functions(scenario)
+
+
+if __name__ == "__main__":
+    setup_logging(logging.INFO)
```
pavelzw, line 35: see other comment

pavelzw (commented):
Thanks!

conda_forge_tick/update_upstream_versions.py
```diff
@@ -183,6 +203,26 @@ def get_latest_version_local(
...
+        f"no_proxy={os.environ.get('no_proxy', '')}",
+        "-e",
+        "SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt",
+        "-e",
+        "REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt",
+        "-v",
+        f"{os.environ['SSL_CERT_FILE']}:/etc/ssl/certs/ca-certificates.crt:ro",
+    ]
+
+
```
ytausch, line 225: will be removed if the PR linked above is merged

ytausch:
The integration tests now run again (see the PR in my fork).

@beckermr, I'd appreciate it if we could prioritize this PR. It would help to establish more trust in future code changes, and it's potentially also very work-intensive to fix the tests again after a significant amount of code changes have been committed to main.

ytausch:
Thanks for reviewing https://github.com/conda-forge/conda-forge-feedstock-ops/pull/34!

I removed the temporary workaround for its functionality in a72eb3753e92bc14d0498826ebf6f246fb3f341e.

How should we proceed with setting up the token?

tests_integration/lib/integration_test_helper.py
```diff
@@ -0,0 +1,166 @@
...
+        with tempfile.TemporaryDirectory() as tmpdir:
+            target_dir = Path(tmpdir) / full_feedstock_name
+            cli.clone_repo(matching_pr.head.repo.clone_url, target_dir)
+            cli.checkout_branch(target_dir, matching_pr.head.ref)
+
+            with open(target_dir / "recipe" / "meta.yaml") as f:
+                meta = f.read()
+
+        assert f'{{% set version = "{new_version}" %}}' in meta
+        assert f"sha256: {new_hash}" in meta
```
pavelzw, line 160: should we also account for recipe.yaml here?
we could for example use https://github.com/conda-forge/rattler-build-feedstock in the tests (in the future?)

pavelzw (approved):
great! apart from one comment this looks good to me :)

tests_integration/lib/integration_test_helper.py
ytausch: Would it be okay for you to do that in a follow-up PR? I purposefully tried to keep this PR minimal with only one test feedstock. In a follow-up, I will add a test for a recipe.yaml feedstock. rattler-build-feedstock sounds like a good example!
I think it makes sense to wait with changes to the integration test helpers until then, to avoid introducing "dead" test code.

pavelzw: yes, sounds good to me!

beckermr:
Ahhh this PR is on a fork so it cannot see the token we added. We'll have to test that the tests run using the fork I guess for now. I forgot about that. Sorry for the trouble!

ytausch:
Not your fault :)
This will also be a problem if the tests are run for PRs from forks.
I am currently looking for a way to solve this in a secure way and asking some GitHub experts at our company.
For your review, I'd suggest not worrying about this for now :D

beckermr:
We can solve this by enabling the merge queue. It works as follows:

- For PRs from forks, we skip the integration tests and run the other tests.
- Then when we press merge, the PR branch gets merged into main and the tests are run in the merge queue. The merge queue branch is on the upstream repo (and has passed code review), so it can see the secrets.
- If the tests pass in the merge queue, the PR is fully merge. If they fail, the PR is removed from the merge queue.

pavelzw:
Sounds good!

beckermr:
the merge queue should be functional now - for this PR we'll need to check the tests on the fork before merging

ytausch:
Awesome, this is also the solution I've wanted to propose now :D

Running the tests in the merge group *only* requires a little workaround for the workflow to be triggered correctly. I've implemented everything here in my latest commit: c9c45c4b302359b8e11107bfa1a3990392e8956e

Matt, we would need to add a new status check `run-test-scenario` to the list of required status checks, and then this should be all good!

.github/workflows/test-integration-dontrun.yml
```diff
@@ -0,0 +1,20 @@
+name: Integration Tests do not run on PRs
```
beckermr, line 1: This file is not working as intended. The required check has a different name according to github. Thus it doesn't show up in automerge and breaks the merge queue (which needs the required checks). Let's remove it and instead we can skip steps in the actual integration tests work if the PR is from a fork. The syntax 

```
if: github.event.pull_request.head.repo.full_name == github.repository  # then NOT on fork
```

from this [forum question](https://github.com/orgs/community/discussions/25217) should do the trick.

beckermr: You'll want this syntax actually

```
if: success() && github.event.pull_request.head.repo.full_name == github.repository
```

beckermr: @ytausch see my note here for another approach that allows PRs from branches to run the integration tests.

ytausch: I think I don't follow completely. Why does the required check have a different name? I think the status check should always be named after the ID (YAML key) of the job.

I have it from here: https://github.com/orgs/community/discussions/51120#discussioncomment-7589658

I see that your solution also works, but I want to avoid copy-pasting this if-condition 20 times if it's avoidable.

ytausch: Here, I tried out the approach that I've implemented so far: https://github.com/ytausch-terraform-tests/test-env/pull/11

Feel free to check out the CI logs to verify that the right workflows are executed. You are also welcome to submit a PR from your fork to try it out yourself!

ytausch: Ahh, I think you should not make `setup-repositories` a required status check for now, only `run-test-scenario`.

ytausch: It now works!

beckermr:
The approach here won't allow us to run integration tests on PRs from branches. Please fix that.

ytausch:
Does that work? eae1330
I know it looks a bit ugly that the "required" status check is shown twice: Once as skipped and once as successful. But I think this is better than copy-pasting if-conditions?

ytausch:
Oh, let me try another one:

> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success". For more information, see [Using conditions to control job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution).

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

This means we might be able to get rid of the dummy workflow completely!

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,321 @@
+name: Integration Tests
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
```
ytausch, line 8: do you think on: push is necessary?

Since we only have concurrency 1 for the integration tests, this might unnecessarily block resources

ytausch:
Tried the current setup here and it works! Think this is very elegant now.

https://github.com/ytausch-terraform-tests/test-env/pull/14

.github/workflows/test-integration.yml
beckermr: Yeah we can skip on push with the merge queue.

beckermr:
You need to add dummy steps that pass for prs from forks, but are skipped for prs from branches. Otherwise automerge is still broken.

ytausch:
Huh, why though? The status check of a skipped workflow is reported as "success". And in the PRs in my demo repo, exactly this worked.

Or do you mean something else than adding to the merge queue with "automerge"?

beckermr:
Ah i see. Sorry for my confusion and thank you!

ytausch:
No worries :) And thank you too!

conda_forge_tick/settings.py
```diff
@@ -0,0 +1,30 @@
...
+https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
+"""
+
+ENV_RANDOM_FRAC_TO_UPDATE = "CF_TICK_RANDOM_FRAC_TO_UPDATE"
+RANDOM_FRAC_TO_UPDATE = float(os.getenv(ENV_RANDOM_FRAC_TO_UPDATE, "0.1"))
+"""
+The fraction of feedstocks (randomly selected) to update in certain jobs. Default: 0.1.
+Currently used by update-upstream-versions and make-graph.
+In tests, you probably need to set this to 1.0 to update all feedstocks.
+"""
```
beckermr, line 30: This fraction should be separate for the versions and graph nodes.

beckermr (changes requested):
Looks good! A few changes are needed.

conda_forge_tick/settings.py
ytausch: Should I rename the env variables then? Currently, they have the same name.

beckermr: Yes rename them.

ytausch: Done in 7fe21f5ce47f029494f6d84ad933745b35173744

beckermr:
This one is waiting on @jaimergp but LGTM for my bits!

beckermr (approved):

jaimergp:
I'll take a look this week. It's indeed a lot, so I might do this in several passes. I hope that's not too frustrating @ytausch!

ytausch:
Take your time!

conda_forge_tick/settings.py
```diff
@@ -0,0 +1,39 @@
+import os
```
jaimergp, line 1: Good job with the name docstrings! A module docstring here would be useful to inform of the purpose of this module and the different design decisions you took:

- Each environment variable _name_ is _also_ given a Python name so it can be reused in other parts of the library.
- There's no validation of the passed value


Also I think you could add type hints to the variables. Most are str but I'm seeing bools and floats too.

conda_forge_tick/utils.py
```diff
@@ -1306,7 +1306,7 @@ def change_log_level(logger, new_level):
         logger.setLevel(saved_logger_level)
 
 
-def run_command_hiding_token(args: list[str], token: str) -> int:
+def run_command_hiding_token(args: list[str], token: str, **kwargs) -> int:
```
jaimergp, line 1309: If you want to do this, you need to do some validation to avoid "duplicate" kwarg errors:

```
>>> from subprocess import run
>>> def runn(*args, **kwargs):
...   return run(*args, text=True, **kwargs)
...
>>> runn(["echo", "True"])
True
CompletedProcess(args=['echo', 'True'], returncode=0)
>>> runn(["echo", "True"], text=False)
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in runn
TypeError: subprocess.run() got multiple values for keyword argument 'text'
```

Make sure kwargs does not include `text`, `stdout` or `stderr`. Or override the defaults with the ones in kwargs, but do handle them somehow.

tests/test_upstream_versions.py
```diff
@@ -1354,7 +1354,9 @@ def test_update_upstream_versions_run_parallel_custom_sources(
     ) == ("source a", "source b")
 
 
-@mock.patch("conda_forge_tick.update_upstream_versions.RANDOM_FRAC_TO_UPDATE", new=1.1)
+@mock.patch(
+    "conda_forge_tick.update_upstream_versions.FRAC_UPDATE_UPSTREAM_VERSIONS", new=1.1
+)
```
jaimergp, line 1359: Shouldn't these be patched in `settings`? I always struggle to remember how what the best practice is. It might be that the current change is the correct one.

conda_forge_tick/make_graph.py
```diff
@@ -231,7 +230,8 @@ def _build_graph_sequential(
     mark_not_archived=False,
 ) -> None:
     for name in names:
-        if RNG.random() >= RANDOM_FRAC_TO_UPDATE:
+        if RNG.random() >= FRAC_MAKE_GRAPH:
+            logger.debug(f"skipping {name} due to random fraction to update")
```
jaimergp, line 234:
```suggestion
            logger.debug("skipping %s due to random fraction to update", name)
```

conda_forge_tick/migration_runner.py
```diff
@@ -231,6 +231,7 @@ def run_migration_local(
 
     # it would be better if we don't re-instantiate ClonedFeedstockContext ourselves and let
     # FeedstockContext.reserve_clone_directory be the only way to create a ClonedFeedstockContext
+    # note: git_repo_owner is not set here because we cannot access it. Reading it possibly yields an incorrect value.
```
jaimergp, line 234: Can you elaborate why this is really a problem (or not) and what would be the ideal situation?

conda_forge_tick/os_utils.py
```diff
@@ -84,14 +85,16 @@ def clean_disk_space(ci_service: str = "github-actions") -> None:
...
+  sudo dpkg --remove -y -f firefox \
                           google-chrome-stable \
                           microsoft-edge-stable
   sudo apt-get autoremove -y >& /dev/null
   sudo apt-get autoclean -y >& /dev/null
   sudo docker image prune --all --force
   df -h
-""")
+"""
+                )
```
jaimergp, line 97: This is weird formatting. I much prefer the idiom:

```python
f.write(
    dedent(
        """
        whatever
        whatever
        """
    ).lstrip()
)
```

or `inspect.cleandoc()`.

Just a note in case this was intentional and not just formatting noise.

conda_forge_tick/deploy.py
```diff
@@ -21,6 +20,14 @@
 
 logger = logging.getLogger(__name__)
 
+"""
+Environment Variables:
+
+GITHUB_WORKFLOW (optional): The name of the workflow.
+RUN_URL (optional): The URL of the run.
+BOT_TOKEN (optional): The bot's GitHub token.
+"""
```
jaimergp, line 29: If this is a module docstring, it should be placed at the top of the module. If it's a comment, it should be a set of `#`-lines.

Also I think you can elaborate a bit on "Environment Variables". Maybe "Environment variables needed to XYZ"?

conda_forge_tick/deploy.py
```diff
@@ -148,22 +155,17 @@ def _deploy_batch(*, files_to_add, batch, n_added, max_per_batch=200):
...
-                    )
-                if status != 0:
-                    print(">>>>>>>>>>>> git push failed", flush=True)
-
+            status = run_command_hiding_token(
+                [
+                    "git",
+                    "push",
+                    f"https://{get_bot_token()}@github.com/{GRAPH_REPO}.git",
+                    "master",
```
jaimergp, line 163: Seeing `master` here I wonder whether we also need a `GRAPH_REPO_DEFAULT_BRANCH` setting.

conda_forge_tick/deploy.py
```diff
@@ -148,22 +155,17 @@ def _deploy_batch(*, files_to_add, batch, n_added, max_per_batch=200):
                     pass
 
                 print(">>>>>>>>>>>> git push try", flush=True)
-                with sensitive_env() as env:
```
jaimergp, line 151: Why is this removed?

conda_forge_tick/git_utils.py
```diff
@@ -193,7 +193,8 @@ def _run_git_command(
         """
         git_command = ["git"] + cmd
 
-        logger.debug(f"Running git command: {git_command}")
+        if not suppress_all_output:
+            logger.debug(f"Running git command: {git_command}")
```
jaimergp, line 197: I prefer sticking to the `%s` format in logging calls, but I see that there's a precedent already so I won't repeat this again. Up to your choice.

conda_forge_tick/lazy_json_backends.py
```diff
@@ -56,10 +57,7 @@
     "migrators",
 ]
 
-CF_TICK_GRAPH_GITHUB_BACKEND_REPO = "regro/cf-graph-countyfair"
-CF_TICK_GRAPH_GITHUB_BACKEND_BASE_URL = (
-    f"https://github.com/{CF_TICK_GRAPH_GITHUB_BACKEND_REPO}/raw/master"
-)
+CF_TICK_GRAPH_GITHUB_BACKEND_BASE_URL = f"https://github.com/{GRAPH_REPO}/raw/master"
```
jaimergp, line 60: Same comment about `master` here.

jaimergp (changes requested):
I am going to do this review in two parts:

1. Changes impacted by introducing a settings module to not hard-code `conda-forge` as the only organization.
2. The integration tests themselves.

This is for (1). 

In hindsight, I think these could have been a better split for the PR: remove the hardcoded `conda-forge` via `settings`, and once that had landed, start (2).

conda_forge_tick/os_utils.py
beckermr: This one looks like ruff-format running against the old code as opposed to a new change. I agree with @jaimergp but I am also ok if we defer this to another PR to go through and use dedent more widely.

conda_forge_tick/deploy.py
beckermr: Yes that'd be great. I have not switched over that repo to main for exactly this reason.

tests_integration/collect_test_scenarios.py
```diff
@@ -0,0 +1,93 @@
...
+    {
+        "feedstock1": ["aarch_migration", "version_update"],
+        "feedstock2": ["version_update"],
+    }
+
+    The return value of this function is deterministic (sorted by feedstock name and test case name).
+    """
+    test_cases = collections.defaultdict(list)
+
+    for test_case in DEFINITIONS_DIR.glob("*/*.py"):
```
jaimergp, line 25: This is the part where you say it's intentionally pseudo-random, right? I would add a comment here or `random.shuffle` explicitly in a prior step.

tests_integration/collect_test_scenarios.py
```diff
@@ -0,0 +1,93 @@
...
+        if test_case_name in SKIP_TEST_CASES:
+            continue
+        feedstock = test_case.parent.name
+        test_cases[feedstock].append(test_case_name)
+
+    return dict(
+        sorted(
+            (feedstock, sorted(test_cases))
+            for feedstock, test_cases in test_cases.items()
+        )
```
jaimergp, line 36: Oh, but then you do sort them?

tests_integration/collect_test_scenarios.py
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+
+def get_number_of_test_scenarios(integration_test_cases: dict[str, list[str]]) -> int:
+    return max(len(test_cases) for test_cases in integration_test_cases.values())
```
jaimergp, line 41: Isn't this supposed to be the sum?

tests_integration/collect_test_scenarios.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def get_all_test_scenario_ids(
+    integration_test_cases: dict[str, list[str]],
+) -> list[int]:
+    return list(range(get_number_of_test_scenarios(integration_test_cases)))
+
+
+def init_random():
+    random.seed(int(os.environ[ENV_GITHUB_RUN_ID]))
```
jaimergp, line 51: Is RUN_ID stable across re-runs of the same workflow? Might be a desirable property of this seed for debuggability.

tests_integration/collect_test_scenarios.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # make sure that each feedstock has exactly n_scenarios test cases
+    # We have to cut the additional test cases here to avoid that some test cases are not run.
+    test_cases_extended = {
+        feedstock: (
+            test_cases
+            * (n_scenarios // len(test_cases) + (n_scenarios % len(test_cases) > 0))
+        )[:n_scenarios]
+        for feedstock, test_cases in integration_test_cases.items()
+    }
```
jaimergp, line 79: I don't get this part of the need for it.

tests_integration/collect_test_scenarios.py
```diff
@@ -0,0 +1,93 @@
...
+
+    for test_cases in test_cases_extended.values():
+        random.shuffle(test_cases)
```
jaimergp, line 82: Aaand know you shuffle them. Hm. Ok, not sure why you sorted them before though.
Do add a comment that shuffle acts in-place.

tests_integration/setup_repositories.py
```diff
@@ -0,0 +1,222 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+IGNORE_FEEDSTOCK_NAMES = {
+    "__pycache__",
+}
+
+
+@dataclass(frozen=True)
+class GitHubAccountSetup:
```
jaimergp, line 45: I think it would be desirable to add a `__post_init__` method here with some validation (or maybe somewhere else) to check that the account or target_names passed are not sensitive. I want to avoid a scenario where someone erroneously passes `conda-forge` or `cf-autotick-bot` or whatever account with enough privileges and deletes stuff from production while running the tests. In other words, a disallowed list of protected accounts and resources.

tests_integration/setup_repositories.py
```diff
@@ -0,0 +1,222 @@
...
+    to_delete = existing_names - target_names
+    to_create = target_names - existing_names
+
+    LOGGER.info(
+        "Deleting the following repositories for %s: %s",
+        owner_name,
+        _or_empty_set(to_delete),
+    )
+    for name in to_delete:
+        owner.get_repo(name).delete()
```
jaimergp, line 137: My comment for validation and protection against accidental deletion of resources refers to lines like these. I don't care where we add the protection, but it needs to happen before this line is run.

tests_integration/step_prepare.py
```diff
@@ -0,0 +1,75 @@
...
+            )
+            pr.edit(state="closed")
+
+
+def reset_cf_graph():
+    IntegrationTestHelper.overwrite_github_repository(
+        GitHubAccount.REGRO_ORG,
+        "cf-graph-countyfair",
+        Path(__file__).parent / "resources" / "empty-graph",
+        branch="master",
```
jaimergp, line 47: Another instance of `master` to be parametrized in `settings.py`

tests_integration/step_validate.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def run_all_validate_functions(scenario: dict[str, str]):
+    test_helper = IntegrationTestHelper()
+    for test_module in get_test_case_modules(scenario):
+        try:
+            logging.info("Validating %s...", test_module.__name__)
+            test_module.validate(test_helper)
+        except AttributeError:
+            raise AttributeError("The test case must define a validate() function.")
```
jaimergp, line 26: Same as with `prepare`

tests_integration/mock_server_addon.py
```diff
@@ -0,0 +1,60 @@
+"""
```
jaimergp, line 1: Maybe add a shebang too?
```suggestion
#!/usr/bin/env mitmdump -s
"""
```

tests_integration/resources/empty-graph/README.md
```diff
@@ -0,0 +1,4 @@
+# cf-graph-countyfair (For Integration Tests)
+
+This repository is used in integration tests of [cf-scripts](https://github.com/regro/cf-scripts) to simulate the
```
jaimergp, line 3:
```suggestion
This repository is used in integration tests of [cf-scripts](https://github.com/regro/cf-scripts) to set up a clean initial state for the
```

Is that more accurate or did I get it wrong?

tests_integration/setup_repositories.py
jaimergp, line 1: The name `setup_repositories` does lead to think this script will create repositories. Maybe there's a better way of naming this?

- `setup_for_repositories`
- `clean_for_repositories`
- `ready_repositories`

?

jaimergp (changes requested):
Ok, first part of my 2nd review for the integration tests.

So far, what I've understood:

- The GHA workflow will ask some preparation scripts to collect all the test cases and return a collection of identifiers for each "scenario".
- Each feedstock has scenarios or test cases? The tests for each feedstock MUST run sequentially.
- Different feedstocks can be tested in parallel.

My main question so far is why we haven't used `pytest` for this, even if it would have required some custom plugins. I'm seeing a lot of "setup", "cleanup", "validate" that I think would have been easier to write and read in the context of our pytest-ready mental model.

I didn't go yet into the actual test cases though, so maybe I just need some more context.

tests_integration/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```text
+definitions/
+├── pydantic/
+│   ├── resources/
+│   │   ├── feedstock
+│   │   └── ... (entirely custom)
+│   ├── version_update.py
+│   ├── aarch_migration.py
+│   ├── some_other_test_case.py
```
jaimergp, line 31: Where do these names come from? Do they have to match something in the bot code?

jaimergp:
Ok so the more I look at it, the more I am convinced of the following:

- The integration tests can only be run on Github Actions because the orchestration of the tests is done in the workflow via the matrix of scenarios and several job steps. This leaves local debugging out of the picture afaik? That constitutes a blocker.
- Instead, why not pytest for everything?
	- The pseudo-randomized collection of test cases can be achieved with [pytest-random-order](https://github.com/pytest-dev/pytest-random-order) or similar solutions. There's a collection hook too in case we want to go ultra custom here.
	- Setup and pre-cleanup can be done with session- or module-wide fixtures.
	- The validate steps are the tests themselves.
	- The idea would be to have the test functions depend on "run-once" fixtures that handle the bot execution parts from pytest instead of GHA.
	- Then each test_something.py module could handle either a type of test or a type of feedstock, whatever you consider is best?

These comments only affect the architecture of the test framework, not the tests themselves. There's a lot of great work in here and I don't want you to feel discouraged just because I'm suggesting another way of orchestrating the tests. The settings refactor is a nice add and there's a lot of effort put into all the feedstock helpers! Thank you for that!

conda_forge_tick/settings.py
ytausch: 👍 , done in 318975ce96363bf6c6e6a163bc19165cbc63418a

conda_forge_tick/make_graph.py
ytausch: I agree this is the better style (and have applied your patch) - I think we should include the ruff rule that enforces it. Can do this in a separate PR!

conda_forge_tick/migration_runner.py
ytausch: I changed the comment here: 75303763683773508362b83d8ab017fe95eb6233 Does that clear things up for you?
We should probably refactor this in a follow-up PR.

conda_forge_tick/os_utils.py
ytausch: Yeah let's go through this in a separate PR.

conda_forge_tick/deploy.py
ytausch: I removed this comment in 7b24fa6527f655a4ff9fb61bcbc7e4ade93ca40e after I discovered it was not complete anyway.
I think we need a better way to document the environment variables, we should do this in a separate PR.

conda_forge_tick/deploy.py
ytausch: Since I don't see a use case for configuring this via an environment variable, I added this as a constant to settings.py here: 7535b95f743a6fa500adc1081303fbc4fb420148

conda_forge_tick/deploy.py
ytausch: Because it is totally sufficient to call `get_bot_token()` to access the bot token and we don't need the complexity of an extra level of indentation.

conda_forge_tick/lazy_json_backends.py
ytausch: fixed!

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,319 @@
...
+      - name: Run pip install
+        run: |
+          cd cf-scripts
+          pip install --no-deps --no-build-isolation -e .
+
+      - name: Set up Integration Test Repositories
+        id: setup_repositories
+        run: |
+          cd cf-scripts
+          python -m tests_integration.setup_repositories
```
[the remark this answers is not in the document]
ytausch: done in 0d849839dd1efd06c4fb54fb4d0b69649aa2bedd

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,319 @@
...
+          sudo wget -e use_proxy=yes -e http_proxy=127.0.0.1:8080 -O /usr/local/share/ca-certificates/mitmproxy.crt \
+            http://mitm.it/cert/pem
+          sudo update-ca-certificates
+        env:
+          SCENARIO_ID: ${{ matrix.scenario_id }}
+          PYTHONPATH: ${{ github.workspace }}/cf-scripts
+
+      - name: "[Test] Gather all Feedstocks"
+        run: |
+          cd cf-graph
```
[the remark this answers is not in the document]
ytausch: Your way is better - I just copied this over from the production workflows.
Fixed here: 026fca29012b9387368ac0da8bbbb908b2cc9085

Probably we also want to change the production workflows one time...

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,319 @@
...
+          # don't use proxy for git CLI operations (applies to all further steps)
+          git config --global http."https://github.com".proxy ""
+          conda-forge-tick --debug gather-all-feedstocks
+
+      - name: "[Test] Deploy to GitHub"
+        run: |
+          cd cf-graph
+          conda-forge-tick --debug deploy-to-github
+
+      - name: Reinstall Bot Code
```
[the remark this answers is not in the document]
ytausch: The individual steps interact with the cf-graph and conda-forge-pinning-feedstock repositories. For cf-graph, we explicitly want to test that bot steps communicate with each other via the cf-graph GitHub repository and not via some weird local state of the filesystem. Therefore, we discard cf-graph (and conda-forge-pinning-feedstock) in between steps.

For example, we don't want to forgive the bot if it forgets to push changes to GitHub and changes are only visible in the local clone.

Therefore, the reinstall clears up the runner and thus needs to "install" the bot again, i.e., clone the repositories again.

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,319 @@
...
+        run: |
+          source cf-scripts/tests_integration/clear_runner.sh
+          source cf-scripts/autotick-bot/install_bot_code.sh --no-clean-disk-space --no-pull-container
+
+      - name: "[Test] Make Graph (Node Attributes)"
+        run: |
+          cd cf-graph
+          conda-forge-tick --debug make-graph
+        env:
+          CF_TICK_FRAC_MAKE_GRAPH: 1.0
```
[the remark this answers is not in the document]
ytausch: fixed in cbd19ec4e914456c79dde9c2dc22a2816c719db7

tests_integration/collect_test_scenarios.py
ytausch: No, `Path.glob` is random. But we sort in the return below.

https://github.com/regro/cf-scripts/blob/cbd19ec4e914456c79dde9c2dc22a2816c719db7/tests_integration/collect_test_scenarios.py#L32-L37

The pseudorandomness is added in `get_test_scenario`.

tests_integration/collect_test_scenarios.py
ytausch: see above, I also clarified the comment here to avoid confusion: 23aa196ce57ad824d9c80d462bf7444f37c3ebdd

tests_integration/collect_test_scenarios.py
ytausch: Yes. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

tests_integration/collect_test_scenarios.py
ytausch: No. Per scenario, we can only run one test case per feedstock, but two test cases for different feedstocks can always run in parallel. So the number of test scenarios is the maximum number of test cases of a feedstock.

If a feedstock has less than this number of test cases, it will just repeat its test cases while we iterate the feedstock with the most test cases. See also my answer to your comment [here](https://github.com/regro/cf-scripts/pull/3417#discussion_r1956118310).

tests_integration/collect_test_scenarios.py
ytausch: comment added in 3bea225d8708dd724494a05c15d161587bd4559b

`collect_integration_test_cases` should not randomize anything, only `get_test_scenario`. The sorting is necessary because otherwise `glob` breaks the deterministic behavior of everything.

tests_integration/step_prepare.py
ytausch: done!

tests_integration/collect_test_scenarios.py
ytausch: See the docs here now (new): https://github.com/ytausch/cf-scripts/tree/integration-tests/tests_integration

tests_integration/setup_repositories.py
ytausch: There are already many safeguards in place to protect against this (ordered by priority):

- The integration test workflow does not use the production bot secret, i.e. does not have access to sensitive infra.

Somebody could try to import the integration test code from the production workflow though. But

- The `GitHubAccount` enum statically defines a list of GitHub accounts that are acceptable to use. Type checking (if respected) thus prevents using a different GitHub account in `GitHubAccountSetup` already. Modifying the `GitHubAccount` enum is possible of course, but so is also modifying an additional check. We will eventually introduce mypy to this repo, enforcing type checks (adding mypy is blocked by this PR itself, by the way. #3203 #3207).

What is left is that somebody could hack around with the production bot token locally on their computer. This should simply not be done. The integration test code is not the only place where this could be very harmful. If it's really not avoidable, the tests should happen at least in the test environment. Even better would be setting up something like the test environment for your local use as an individual developer.

I'd see the following extra potential for safeguards:

- Use GitHub Actions environments to prevent adding the production secret to the integration test job in the workflow file (because each GitHub Actions job can only have one environment, and we'd have `prod` and `integration-tests`).

WDYT?

ytausch: Maybe you have a different vulnerable scenario in mind?

tests_integration/setup_repositories.py
ytausch: see above

tests_integration/step_prepare.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def run_all_prepare_functions(scenario: dict[str, str]):
+    test_helper = IntegrationTestHelper()
+    for test_module in get_test_case_modules(scenario):
+        try:
+            logging.info("Preparing %s...", test_module.__name__)
+            test_module.prepare(test_helper)
+        except AttributeError:
+            raise AttributeError("The test case must define a prepare() function.")
```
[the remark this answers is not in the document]
ytausch: fixed in 41c461ec0382f6879695b34c887aede2ce03f0cf

tests_integration/resources/empty-graph/README.md
ytausch: That's good!

tests_integration/setup_repositories.py
ytausch: It creates repositories!

tests_integration/README.md
ytausch: clarified in ea48c90f7706ca6738648dfde813e3043fe5e34b

ytausch:
Thank you for reviewing this, @jaimergp!

> The integration tests can only be run on Github Actions because the orchestration of the tests is done in the workflow via the matrix of scenarios and several job steps. This leaves local debugging out of the picture afaik? That constitutes a blocker.

Allowing developers to debug the bot locally is still something we are/I am actively working on. While significant progress has been made, a large portion of bot tasks is still cumbersome to debug locally. A big part was the tight coupling of code to GitHub, which has been remedied by the `GitCli` and `GitPlatformBackend` abstractions I introduced.

Debugging the integration tests also means that you, by design, need access to a GitHub account setup like the integration test environment.

However, of course, we don't want to build the integration tests in a way that makes things harder than already. I thought of it like this:
- The matrix of scenarios does not block you from debugging locally. All you need is the run ID and the scenario ID from the failing test, and you can just execute that locally by setting the correct environment variable.
- You execute the individual steps one after another manually.

I had a setup like this while I developed the tests and did not find it annoying that I had to click through the individual steps. Of course, we might want to have one command that executes the whole bot's pipeline sequentially and locally, but this is far out of reach today I think.

---

> The pseudo-randomized collection of test cases can be achieved with [pytest-random-order](https://github.com/pytest-dev/pytest-random-order) or similar solutions. There's a collection hook too in case we want to go ultra custom here.

I've read the docs of `pytest-random-order` but I think that we would have to use the collection hook at best because our use case is very custom: We first need to run all prepare steps, and then all validate steps. I don't see how `pytest-random-order` helps with that.

> Setup and pre-cleanup can be done with session- or module-wide fixtures.
The validate steps are the tests themselves.
The idea would be to have the test functions depend on "run-once" fixtures that handle the bot execution parts from pytest instead of GHA.
Then each test_something.py module could handle either a type of test or a type of feedstock, whatever you consider is best?

I don't understand how you
- want to execute **all** fixtures/prepare functions for a test scenario (which consists of random test cases, each having different fixtures) before **all** pytest test cases (validate functions) of a scenario.

We also somehow need to assemble the FastAPI router for each test scenario from each test case.

To put it briefly, I'm not sure what exactly is left of pytest if we need to customize it everywhere to accomplish what we need. Currently, the test collection logic is very short with < 100 lines of code in `collect_test_scenarios.py` and I doubt a bit that it will get easier if we instead use a whole bunch of exotic pytest features to accomplish the same thing. 

Most value of the integration tests will come out of the fact that they break if the bot breaks, as an early warning signal, a *smoke test* to prevent faulty PRs from being merged. This is also why I saw being close to the GitHub workflow as an advantage: Each step in the integration test is called just like the production workflow would call it. Better local debugging would be cool, but a lot of other construction sites need to be cleaned up to make a different setup worth it I think.

---

This was a long wall of text and I think we likely need some more room for discussion. Happy to set up a meeting if that makes things easier!

beckermr:
> You execute the individual steps one after another manually.

This is what we want to avoid. With the pytest fixtures, we don't need to copy-paste from a GHA workflow and then have to cleanup resources later by hand. Instead, we can use fixtures that clean themselves up.

ytausch:
There is no need to "clean up" test resources. The test setup overwrites everything automatically.

jaimergp:
> want to execute all fixtures/prepare functions for a test scenario (which consists of random test cases, each having different fixtures) before all pytest test cases (validate functions) of a scenario.

I think I need to understand why we need to prepare for **all** test cases before we start executing them, and we can't just go one by one as needed.

> We also somehow need to assemble the FastAPI router for each test scenario from each test case.

That looks like the case for a fixture factory or just a helper function in the test body?

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,319 @@
+name: Integration Tests
+on:
```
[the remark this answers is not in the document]
jaimergp: Only asking because we do have it in `tests.yml`. Maybe it's redundant there? @beckermr

ytausch:
> I think I need to understand why we need to prepare for all test cases before we start executing them, and we can't just go one by one as needed.

For two reasons:
- Running a test scenario is slow because it means that we need to run the entire pipeline of the bot, one step after another. Parallelization is not possible because we need to exclusively use GitHub resources. So we want to group test cases of different feedstocks together for a very significant speedup.
- Testing the test cases individually would mean that the bot under test only sees conda-forges that only consist of one feedstock. This is not realistic and also would completely hide bugs where feedstocks interact with each other (which can occur because they are listed in shared data steuctures in cf-graph). Also, it is good practice to randomize test cases running together to catch bugs in their combination. This is what the current approach does.

> That looks like the case for a fixture factory or just a helper function in the test body?
The problem is that I don't see how the concept of "combining test cases randomly to run together" exists in pytest. So essentially the same as above.

jaimergp:
> Running a test scenario is slow

How slow are we talking about?

> a very significant speedup.

For clarity, I'll need some numbers attached to these statements. Not that I don't trust your word for it, but I want to know because maybe the compromise is acceptable under certain contexts.

> The problem is that I don't see how the concept of "combining test cases randomly to run together" within the constraints we need exists in pytest. So essentially the same as above.

A [collection plugin](https://docs.pytest.org/en/stable/reference/reference.html#collection-hooks) should help here, and hopefully fits better within the mental model of contributors that are already used to using pytest. @zklaus took a look at this type of hook for the pytest-conda-solvers plugin, and it fits in [a single module](https://github.com/zklaus/pytest-conda-solvers/blob/main/pytest_conda_solvers/plugin.py). Feel free to ask any questions if necessary.

The problem I see is that there's no local debugging workflow. It's not even documented. This a huge lift with a lot of custom code and I'm trying to make it easier down the road. I insist on pytest because of its huge community.

Your README is well written but it's missing a lot about the rationale of the design choices made in this PR. Choices that others might have done differently with the experience of maintaining this codebase. That's why I'm so "picky" with these questions and requests. As an example, the terminology used to describe all the moving pieces is insufficient to convey the structure of the tests: scenarios, definitions, test cases, integration tests... I still don't have a clear idea of how things are running.

In short, can you take a look at how this would look like with more pytest primitives and less custom code? No need to provide a working implementation, simply taking a look to see if we are in the realm of feasible or not.

beckermr:
I 100% agree with @jaimergp here. Even if the pytest customization is somewhat extensive, the benefits of using pytest for future devs out weigh the costs.

ytausch:
> How slow are we talking about?

8 Minutes. https://github.com/ytausch/cf-scripts/actions/runs/12989304494/job/36222194639

> For clarity, I'll need some numbers attached to these statements. Not that I don't trust your word for it, but I want to know because maybe the compromise is acceptable under certain contexts.

Okay, let's craft an example. Imagine we have 5 feedstocks with 2 test cases each. The current approach would run 2 test scenarios, which is roughly `2 * 8 = 16` minutes. Running everything sequentially roughly takes `2 * 8 * 5 = 80` minutes. In reality, we'd need even more test cases if we want to test every or most migrations of the bot. The sequential approach does not scale at all, in the current approach we could just distribute this across different feedstocks to keep the number of test scenarios low.

> The problem I see is that there's no local debugging workflow. It's not even documented. This a huge lift with a lot of custom code and I'm trying to make it easier down the road.

Documenting the local debugging better is something we can work on! Since this is dependent on whether we use pytest or not, I'd want to postpone this until we've decided.

> I still don't have a clear idea of how things are running.

Do you have a concrete point in the README you want to have improved?

ytausch:
> A [collection plugin](https://docs.pytest.org/en/stable/reference/reference.html#collection-hooks) should help here, and hopefully fits better within the mental model of contributors that are already used to using pytest. @zklaus took a look at this type of hook for the pytest-conda-solvers plugin, and it fits in [a single module](https://github.com/zklaus/pytest-conda-solvers/blob/main/pytest_conda_solvers/plugin.py). Feel free to ask any questions if necessary.

I'll have a look into this, thank you!

tests_integration/test_integration.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@pytest.fixture
+def mitmproxy(scenario: tuple[int, dict[str, TestCase]]):
+    scenario_id, _ = scenario
+    print(f"Starting mitmproxy for scenario {scenario_id}...")
+
+    proxy_process = subprocess.Popen(["./mock_proxy_start.sh"])
+    yield
+    proxy_process.kill()
```
jaimergp: I like https://pytest-xprocess.readthedocs.io/en/latest/ a lot for these cases. They handle cleanups and so on. Just mentioning it in case it helps. For example, I would guard the `.kill()` bit in a try / finally block.

jaimergp:
I know it's not ready yet but just wanted to say _thank you!_ for giving it a try with pytest. The code I'm seeing makes a lot of sense!

tests_integration/test_integration.py
ytausch: Thank you! This was indeed very helpful. Implemented in 56f08791e333f0dfd48dfbda4dde8f199089467f

tests_integration/test_integration.py
```diff
@@ -0,0 +1,297 @@
...
+
+    with tempfile.TemporaryDirectory() as tmpdir_s:
+        tmpdir = Path(tmpdir_s)
+
+        cf_graph_repo = settings().graph_github_backend_repo
+        subprocess.run(
+            [
+                "git",
+                "clone",
+                "--depth=5",
```
pavelzw, line 164: why 5?

pavelzw (approved):
lgtm, only minor comments from my side. thanks!

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: |
+          pytest -s -v \
+          --dist=no \
+          tests_integration
+        env:
+          BOT_TOKEN: ${{ secrets.GH_TOKEN_STAGING_BOT_USER }}
+          TEST_SETUP_TOKEN: ${{ secrets.GH_TOKEN_STAGING_BOT_USER }}
+
+      - name: Print Proxy Logs
+        run: cat /tmp/mitmproxy.log
```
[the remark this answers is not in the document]
pavelzw: just saw the mitmproxy script log level is only at 0 instead of 4 in CI. as long as this is kept, i guess this should be fine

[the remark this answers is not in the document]
ytausch: yep, this is the idea :)

tests_integration/test_integration.py
```diff
@@ -0,0 +1,297 @@
...
+CF_SCRIPTS_ROOT_DIR = TESTS_INTEGRATION_DIR.parent
+MITMPROXY_CONFDIR = TESTS_INTEGRATION_DIR / ".mitmproxy"
+MITMPROXY_CERT_BUNDLE_FILE = MITMPROXY_CONFDIR / "mitmproxy-cert-bundle.pem"
+
+
+@pytest.fixture(scope="module", autouse=True)
+def global_environment_setup():
+    """
+    Set up the global environment variables for the tests.
+    If we once migrate to pydantic-settings, this should be more fine-grained for each bot step.
```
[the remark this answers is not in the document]
ytausch: Ah yes^^

tests_integration/test_integration.py
ytausch: because this is the same setting that prod uses: https://github.com/regro/cf-scripts/blob/d3f591c2c7ac9f09c364c024d019cd7ff4f870d9/autotick-bot/install_bot_code.sh#L46

I'd guess this avoids downloading the entire graph history but still allows going back a few commits.

pavelzw: maybe add a reference to this file somewhere here?

tests_integration/mitmproxy_setup_wizard.sh
```diff
@@ -0,0 +1,68 @@
...
+mitmproxy_ca_crt_file="${mitmproxy_dir}/mitmproxy-ca.crt"
+
+if [ "${headless_mode}" = "true" ]; then
+    echo "Executing the headless mode trust script..."
+    bash "${headless_mode_trust_script}" "${mitmproxy_ca_crt_file}"
+else
+    echo "You now need to trust the mitmproxy-ca.crt certificate in your system's trust store."
+    echo "The exact process depends on your operating system."
+    echo "On MacOS, drag and drop the mitmproxy-ca.crt file into the Keychain Access app while having the 'Login' keychain selected."
+    echo "Then, double-click the certificate in the keychain and set ‘Always Trust‘ in the ‘Trust‘ section."
```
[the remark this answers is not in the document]
ytausch: done in 13d579f7feae90d2c9ad2911f2d7c93dffa17131, I used `/etc/debian_version` and `$OSTYPE` because this seems to be "the recommended way" (TM)

tests_integration/test_integration.py
ytausch: jup, done in 56e8afe39189e684f3fffd70803338f30902163d

ytausch:
@jaimergp Did you have a chance to have a look at this? I'll return to office on May 2nd :)

conda_forge_tick/auto_tick.py
```diff
@@ -960,6 +955,7 @@ def _is_migrator_done(_mg_start, good_prs, time_per, pr_limit, tried_prs):
 
 def _run_migrator(migrator, mctx, temp, time_per, git_backend: GitPlatformBackend):
     _mg_start = time.time()
+    home_directory = os.getcwd()
```
jaimergp, line 958: Why call this home_directory if it's not? Should it just be `working_dir` or similar?

tests_integration/.mitmproxy/README.md
```diff
@@ -0,0 +1,4 @@
+# mitmproxy confdir
+
+This directory is meant to be used as configuration directory for `mitmproxy`.
+Check the README in [tests_integration](../tests_integration) for more information.
```
jaimergp, line 4: There's a typo here.

tests_integration/collect_test_scenarios.py
```diff
@@ -0,0 +1,60 @@
...
+    def pop_test_scenario():
+        scenario: dict[str, TestCase] = {}
+        for feedstock in test_cases_extended:
+            scenario[feedstock] = test_cases_extended[feedstock].pop()
+        return scenario
+
+    for _ in range(scenario_id):
+        pop_test_scenario()
+
+    return pop_test_scenario()
```
jaimergp, line 60: Why pop here? :thinking:

tests_integration/lib/integration_test_helper.py
```diff
@@ -0,0 +1,166 @@
...
+from tests_integration.lib.shared import (
+    FEEDSTOCK_SUFFIX,
+    GitHubAccount,
+    get_github_token,
+)
+
+LOGGER = logging.getLogger(__name__)
+
+
+class IntegrationTestHelper:
```
jaimergp, line 23: Is there a real need for a class here? I don't see much shared state and it could just be a collection of public functions (plus their private counterparts). Or is it more about typing?

tests_integration/lib/shared.py
```diff
@@ -0,0 +1,123 @@
...
+        "https://api.anaconda.org/package/conda-forge/conda-forge-pinning",
+        "https://api.anaconda.org/download/conda-forge/conda-forge-pinning/*",
+        "https://binstar-cio-packages-prod.s3.amazonaws.com/*",
+    }
+
+    if any(not url.startswith("https://") for url in transparent_urls):
+        # this is to protect against mistakes and typos, adjust if it ever becomes too strict
+        raise ValueError("All URLs in transparent_urls must start with https://")
+
+    # noinspection HttpUrlsUsage
```
jaimergp, line 99: Can you clarify this comment?

tests_integration/lib/test_case.py
```diff
@@ -0,0 +1,29 @@
+from abc import ABC
+
+from fastapi import APIRouter
+
+from tests_integration.lib.integration_test_helper import IntegrationTestHelper
+
+
+class TestCase(ABC):
+    """
+    Abstract base class for a single test case in a scenario.
```
jaimergp, line 10: Can you elaborate here on this docstring what the expected lifetime of the testcase instance is? I think that we run `prepare()` then the bot does its thing, and then we run again with `validate()`, correct?

tests_integration/test_integration.py
jaimergp, line 1: Would be nice to have a top-level docstring here documenting the setup, or at least pointing to the README document for those that have not noticed its existence.

tests_integration/test_integration.py
```diff
@@ -0,0 +1,298 @@
...
+        )
+
+        os.chdir(cf_graph_dir)
+        yield
+
+    os.chdir(old_working_dir)
+
+
+@contextlib.contextmanager
+def mitmproxy_env():
```
jaimergp, line 219: This can be done with `monkeypatch.setenv` if you want, but maybe this is more practical.

tests_integration/test_integration.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+    Invoke the bot command with the given arguments.
+    """
+    from conda_forge_tick import cli
+
+    cli.main(args, standalone_mode=False)
+
+
+@pytest.mark.parametrize("use_containers", [False, True])
+def test_scenario(
```
jaimergp, line 245: This deserves a docstring with a brief overview and pointers to where to find more details

jaimergp (commented):
Took a quick look offline while in a plane, so apologies for the brevity here. I'll take a deeper look later this week. For now:

- Nice use use of parametrized fixtures to populate the test cases in pytest. However this introduces a few levels of indirections between the test function and the actual definition of the test case. It would be nice to have closer-to-code hints / comments to make this relation ship clear. Or at least a note that this is documented in the note. I feel that most folks would read the docstring of the test function first, so that should provide a good overview of the overall setup.
- I see a lot of modules with helper code. I'm not sure on what purpose this serves now that they don't need to be called separately and everything is orchestrated by pytest. Can you clarify why we need `collect_test_scenarios`, `step_prepare`, `step_validate`, `setup_repositories`, plus the `lib/` directory? I think the only module we need on its own is the MITM server so it can be run by XProcess, right? To better separate `test_*` from everything else, I'd rather aggregate everything under a single module or a single package so there's a clear separation.

conda_forge_tick/auto_tick.py
ytausch: This came from the original name `BOT_HOME_DIR`. I renamed it to `initial_working_dir` to make it less ambiguous.

tests_integration/.mitmproxy/README.md
ytausch: fixed in d2afa6f72e499a90a36236e1e5ebeb07e4087a55

tests_integration/collect_test_scenarios.py
ytausch: Thanks a lot for finding this! This was indeed incorrect (and unnecessarily complicated).

FIxed in 2630652c736317dc7e73c963f4fcac735b5e8904

tests_integration/lib/integration_test_helper.py
ytausch: We could convert everything into public functions, but this approach makes it very clear IMO that you are supposed to use the instance methods of the helper within a test case. Also, we keep the flexibility of adding some state that we might need in the future (e.g., the ID of a pull request that has been created).

tests_integration/lib/shared.py
ytausch: d9fce8959e6c13bc4e056bd061edac38a742c0e8

tests_integration/lib/test_case.py
ytausch: Yes, there even is only one global instance of each test case that might run multiple times (e61bb2d641b1d8a84dfc4fc733d4c63de1a1c400).

Ideally, we would create one instance *per test case execution*, but I think this adds too much complexity for now, as we don't store any state in a test case currently.

tests_integration/test_integration.py
ytausch: done in 7e65060948e2d200a3e882190ac0c8f1e0b4f9d0

tests_integration/test_integration.py
ytausch: refactoring to using `monkeypatch.setenv`: cd1007c22431d5b3b4e749f3a618b47ad60a7b9a

tests_integration/test_integration.py
ytausch: done in 5dd850d01f8e334af3b413e8981f49c9257cfdc2

ytausch:
Thanks a lot for having another look!

> Nice use use of parametrized fixtures to populate the test cases in pytest. However this introduces a few levels of indirections between the test function and the actual definition of the test case. It would be nice to have closer-to-code hints / comments to make this relationship clear. Or at least a note that this is documented in the README. I feel that most folks would read the docstring of the test function first, so that should provide a good overview of the overall setup.

Jup, good point about keeping developer experience in mind. Following your individual comments, I tried to clarify some things directly in the code and refer to the README. Let me know if you still miss it somewhere!

> I see a lot of modules with helper code. [...]

Good point, I think this grew over time because of the non-pytest approach that I followed previously. I refactored the module structure in d712b05ae04d3101dc22fe0aeddb3bae09c8f6e3. As I did it, the `tests_integration.lib` module centrally exposes the public API for writing test code. Still, maintainability is preserved by keeping the private sub-modules in the lib directory. I merged the very small sub-modules `step_prepare` and `step_validate` into `_run_test_cases`. Let me know what you think!

beckermr:
pre-commit.ci autofix

ytausch:
The tests currently fail, which is caused by the settings not being passed to the container job, which was previously not an issue because the container didn't pull `version_pr_info` from GitHub (see #4186).

A solution is blocked by https://github.com/conda-forge/conda-forge-feedstock-ops/pull/48.

tests_integration/__init__.py
jaimergp, line 1: We don't need this anymore, do we?

tests_integration/mock_proxy_start.sh
```diff
@@ -0,0 +1,25 @@
...
+else
+  flow_detail=0
+fi
+
+if [[ -z "$MITMPROXY_CONFDIR" ]]; then
+  echo "Set $MITMPROXY_CONFDIR to a directory containing a mitmproxy-ca.pem CA certificate to intercept HTTPS traffic."
+  exit 1
+fi
+
+# You might need to set PYTHONPATH to the root of cf-scripts
```
jaimergp, line 20: Can we enforce this with the appropriate working directory?

tests_integration/mock_server_addon.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env mitmdump -s
+"""
+Start this file with `mitmdump -s mock_server_addon.py`.
```
jaimergp, line 3: Can't we leverage `pytest-xprocess` for this kind of background server task?

tests_integration/test_integration.py
jaimergp: You can use [`monkeypatch.context()`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.context)

docs/assets/integration-tests/scenarios-definition-1-dark.svg
jaimergp, line 1: These SVGs look great but in the end they are just tables and I worry that future contributions will neglect them due to having to discover the .drawio file and so on. Even if it looks slightly worse, can't we just use markdown tables and a couple emojis here and there (as replacement for the background color)?

tests_integration/__init__.py
ytausch: We do since we import from this module:

https://github.com/regro/cf-scripts/blob/eb708b04298f52e8e92277d5512cfbf5bc65a144/tests_integration/test_integration.py#L19-L29

tests_integration/mock_proxy_start.sh
ytausch: The `PYTHONPATH` is needed because the `mock_server_addon` module needs to be able to import from `tests_integration`, and there is some weird stuff going on in mitmdump. I just tested again, and changing the working directory to the root of cf-scripts does not make it work without `PYTHONPATH`.

tests_integration/mock_server_addon.py
ytausch: We do (thanks to your feedback!):

https://github.com/ytausch/cf-scripts/blob/68f161b83eb2c9a1df626117b27aa065332f339d/tests_integration/test_integration.py#L125-L145

jaimergp: Ah I was confused by the need of a shell script. Is it just for convenience or does it actually have to be a shell script for the startup? I would have expected a call to mitmdump directly, since most of the setup there seems to be about setting certain env vars and flags, all possible from this interface.

jaimergp: I'd rather reduce the amount of moving pieces we need to maintain. I was thinking we could still have this block:

```
     class MitmproxyStarter(ProcessStarter): 
         args = ["./mock_proxy_start.sh"] 
         timeout = 60 
         popen_kwargs = {"cwd": TESTS_INTEGRATION_DIR} 
         env = os.environ | { 
             "SCENARIO_ID": str(scenario_id), 
             "PYTHONPATH": str(CF_SCRIPTS_ROOT_DIR.resolve()), 
         } 
  
         def startup_check(self): 
             return is_proxy_running(port=8080) 
  
     xprocess.ensure("mitmproxy", MitmproxyStarter) 
  
     yield 
  
     xprocess.getinfo("mitmproxy").terminate() 
```

as a standalone function that we can call from a Python script directly (under a `if __name__ == "__main__":` block)

But I won't hold this PR on this, no probs.

jaimergp (approved):

jaimergp:
I won't hold this PR any longer, sorry for the multiple rounds of reviews and extended time it took to get here! I'm very pleased with the result and this is a great improvement for the codebase! So thank you A LOT for the patience 🙏 Let's fix the conflicts and merge right away so we can take this out of our plates!

jaimergp:
cc @beckermr for visibility and eventual click-to-merge! I'm too afraid to click that button here 😬

conda_forge_tick/feedstock_parser.py
```diff
@@ -705,6 +709,8 @@ def load_feedstock_containerized(
         extra_container_args=[
             "-e",
             f"{ENV_CONDA_FORGE_ORG}={settings().conda_forge_org}",
+            "-e",
+            f"{ENV_GRAPH_GITHUB_BACKEND_REPO}={settings().graph_github_backend_repo}",
         ],
     )
```
ytausch, line 715: these additions are a hotfix until we have a solution for https://github.com/conda-forge/conda-forge-feedstock-ops/pull/49

tests_integration/mock_server_addon.py
ytausch: Okay, then let's keep this in mind if there is some rework in the future!

jaimergp:
All these conda-lock conflicts make me wonder whether an (additional?) non-main `push` event would be better suited in this repo.

ytausch:
@beckermr All set, this PR can be merged! Please see [here](https://github.com/ytausch/cf-scripts/pull/4) for validating that CI is indeed passing.

beckermr:
Do you want the commits squashed or kept as is?

This PR is big and IDK how much of the history is useful to you here.

ytausch:
Let's squash

beckermr:
I'll merge by hand once the tests on this PR pass.

PR #3525, opened 2025-01-06
feat: add version bumping for v1 recipes

wolfv:
Another version of version bumping for v1 recipes. This time it should handle all the cases :)

Needs: https://github.com/prefix-dev/rattler-build-conda-compat/pull/57

With the pixi toml attached below, you can try this with:

```
mkdir trash
cd trash && git clone https://github.com/conda-forge/jolt-physics-feedstock && cd ../
pixi run bump ./trash/jolt-physics-feedstock/ 5.2.0
cd  trash/jolt-physics-feedstock && git diff
```

Should show you an updated recipe with version & hash bumped to 5.2.0

I am using this `pixi.toml` locally: 

<details>
  <summary>pixi.toml</summary>

```toml
[project]
authors = ["Wolf Vollprecht <w.vollprecht@gmail.com>"]
channels = ["conda-forge"]
description = "Add a short description here"
name = "cf-scripts"
platforms = ["osx-arm64"]
version = "0.1.0"

[tasks]
pip = { cmd = "pip install -e ../../rattler-build-conda-compat --no-deps", outputs = [".pixi.toml"] }
bump = { cmd = "python ./conda_forge_tick/update_recipe/version.py", depends-on = ["pip"] }

[dependencies]
attrs = "*"
backoff = "*"
black = "*"
cachecontrol = "*"
cachetools = "*"
click = "*"
conda = "*"
conda-forge-feedstock-check-solvable = ">=0.4.0"
conda-forge-pinning = "*"
conda-libmamba-solver = "*"
conda-forge-metadata = ">=0.3.0"
conda-package-handling = "*"
conda-smithy = "*"
conda-forge-feedstock-ops = ">=0.9.0"
conda-build = ">=3.24.0"
curl = "*"
depfinder = "*"
distributed = "*"
doctr = "*"
feedparser = "*"
frozendict = "*"
git = "*"
"github3.py" = "*"
grayskull = ">=2.5"
jinja2 = "*"
lockfile = "*"
mamba = ">=0.23"
msgpack-python = "*"
networkx = "!=2.8.1"
numpy = "*"
packaging = "*"
psutil = "*"
pygithub = "*"
pymongo = "*"
pynamodb = "*"
python-dateutil = "*"
python-graphviz = "*"
python-rapidjson = "*"
requests = "*"
rever = "*"
"ruamel.yaml" = "*"
"ruamel.yaml.jinja2" = "*"
scipy = "*"
setuptools = "*"
setuptools_scm = ">=7"
stopit = "*"
streamz = "*"
tar = "*"
tqdm = "*"
wurlitzer = "*"
yaml = "*"
pandas = "*"
xonsh = "*"

[pypi-dependencies]
conda-forge-tick = { path = ".", editable = true }
rattler-build-conda-compat = { path = "../../rattler-build-conda-compat", editable = true }
```
</details>


wolfv:
I just added a first test that demonstrates that talking to the pypi API also works for the new recipe format.

I'll start cutting a release over on rattler-build-conda-compat and then we can move forward here. We can port some more tests over.

minrk:
This is the main blocker for me picking up v1 in some of the most expensive feedstocks on conda-forge, so I'm super excited about this one!

conda_forge_tick/migrators/version.py
```diff
@@ -197,21 +222,27 @@ def migrate(
...
-            raw_meta_yaml = fp.read()
-
-        updated_meta_yaml, errors = update_version(
-            raw_meta_yaml,
-            version,
-            hash_type=hash_type,
-        )
+        recipe_dir = Path(recipe_dir)
+        recipe_file = recipe_dir / "meta.yaml"
+        recipe_yaml = recipe_dir / "recipe.yaml"
```
h-vetinari, line 227: This naming is very confusing. Both are recipes, both are files, both are in yaml. I suggest either `meta_yaml` / `recipe_yaml` or something like `recipe_file_v0` / `recipe_file_v1`

conda_forge_tick/migrators/version.py
```diff
@@ -93,11 +93,36 @@ def filter(
...
+            {},
+            0,
+        )
+        print("Checking schema version: %d" % schema_version)
+        if schema_version == 0:
+            if (
+                "raw_meta_yaml" in attrs
+                and "{% set version" not in attrs["raw_meta_yaml"]
+            ):
+                print("ugh")
```
beckermr, line 109: Remove this.

conda_forge_tick/migrators/version.py
```diff
@@ -93,11 +93,36 @@ def filter(
...
+                return True
+        elif schema_version == 1:
+            print("Checking ... ")
```
beckermr, line 112: remove this

conda_forge_tick/migrators/version.py
```diff
@@ -93,11 +93,36 @@ def filter(
...
+                yaml = load_yaml(attrs["raw_meta_yaml"])
+                if "context" in yaml:
+                    if "version" not in yaml["context"]:
+                        return True
+        else:
+            raise NotImplementedError("Schema version not implemented!")
+        print("Still OK")
         conditional = super().filter(attrs)
-
+        print("Conditional: %r" % conditional)
```
beckermr, line 125: remove this

conda_forge_tick/update_recipe/version.py
```diff
@@ -152,20 +153,35 @@ def _try_pypi_api(url_tmpl: str, context: MutableMapping, hash_type: str, cmeta:
...
+        # for v1 recipe compatibility
+        if "outputs" in cmeta:
+            if package_name := output.get("package", {}).get("name", None):
+                orig_pypi_name_candidates.append(package_name)
+
     orig_pypi_name_candidates = sorted(
         {nc for nc in orig_pypi_name_candidates if nc is not None and len(nc) > 0},
         key=lambda x: len(x),
     )
-    logger.info("PyPI name candidates: %s", orig_pypi_name_candidates)
```
beckermr, line 168: please add back this statement

conda_forge_tick/update_recipe/version.py
```diff
@@ -557,20 +574,28 @@ def update_version_feedstock_dir(
...
     if updated_meta_yaml is not None:
-        with open(os.path.join(feedstock_dir, "recipe", "meta.yaml"), "w") as f:
+        with open(recipe_path, "w") as f:
             f.write(updated_meta_yaml)
 
     return updated_meta_yaml is not None, errors
 
 
-def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_type):
+def _update_version_feedstock_dir_containerized(
```
beckermr, line 596: This function does not use the recipe_version keywork AFAICT. Should it?

conda_forge_tick/update_recipe/version.py
```diff
@@ -625,7 +650,103 @@ def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_typ
     return data["updated"], data["errors"]
 
 
-def update_version(raw_meta_yaml, version, hash_type="sha256"):
+def update_version_v1(
```
beckermr, line 653: the prints here need to be debug logging statements

conda_forge_tick/url_transforms.py
```diff
@@ -132,3 +132,8 @@ def gen_transformed_urls(url):
         if new_url not in yielded:
             yield new_url
             yielded.add(new_url)
+
+
+if __name__ == "__main__":
+    url = "https://github.com/xtensor-stack/xtensor/archive/{{ version }}.tar.gz"
+    print(set(gen_transformed_urls(url)))
```
beckermr, line 139: Let's remove this debugging code.

conda_forge_tick/update_recipe/version.py
```diff
@@ -773,3 +897,30 @@ def update_version(raw_meta_yaml, version, hash_type="sha256"):
     else:
         logger.critical("Recipe did not change in version migration!")
         return None, errors
+
```
beckermr, line 900: Let's remove this debugging code.

conda_forge_tick/migrators/version.py
```diff
@@ -197,21 +222,27 @@ def migrate(
...
+        if recipe_file.exists():
+            raw_meta_yaml = recipe_file.read_text()
+            updated_meta_yaml, errors = update_version(
+                raw_meta_yaml,
+                version,
+                hash_type=hash_type,
+            )
+        elif recipe_yaml.exists():
+            recipe_file = recipe_yaml
+            updated_meta_yaml, errors = update_version_v1(
```
beckermr, line 237: This routine calls into jinja2 and so needs to be executed in a container. This is not happening here AFAICT.

beckermr (changes requested):
I left several comments that need addressing.

wolfv:
Thanks for the review @beckermr - I'll fix things up tomorrow morning. On a high-level, do things look alright to you?

beckermr:
Yes this looks good. Given how it works, we can switch the v0 recipes over to this algorithm as well and that will likely be more robust than what is in the bot now. 

I just noticed you missed some of old vs new jinja2 stuff in the url transforms module. The current test doesn't hit those code paths which is why nothing weird happened.

beckermr:
It looks like the relock command didn't work on this PR from a fork? Let me try again to be sure.

/relock-conda

conda_forge_tick/migrators/version.py
wolfv: This code is essentially equivalent to `update_version` which is called below, outside a container etc. (uses the same Jinja eval methods).

conda_forge_tick/update_recipe/version.py
```diff
@@ -557,20 +575,28 @@ def update_version_feedstock_dir(
         )
 
 
-def _update_version_feedstock_dir_local(feedstock_dir, version, hash_type):
-    with open(os.path.join(feedstock_dir, "recipe", "meta.yaml")) as f:
+def _update_version_feedstock_dir_local(
+    feedstock_dir, version, hash_type
+) -> (bool, set):
+    if os.path.join(feedstock_dir, "recipe", "recipe.yaml"):
+        return update_version_v1(feedstock_dir, version, hash_type)
```
minrk, line 582: `update_version_v1` returns the updated recipe (not a bool), it doesn't write it back to recipe.yaml. That's handled below in this funciton, but because of this `return`, the write-back never happens, so running `pixi run bump` results in no changes written to the file.

conda_forge_tick/migrators/version.py
h-vetinari: If you need to assign something to the same variable, then use `recipe_file` for that variable, and where you need to distinguish, use `recipe_file_v0` and `recipe_file_v1`. The current variable naming is not suitable

h-vetinari (dismissed):

conda_forge_tick/update_recipe/version.py
```diff
@@ -152,20 +153,36 @@ def _try_pypi_api(url_tmpl: str, context: MutableMapping, hash_type: str, cmeta:
         return None, None
```
ytausch, line 153: Not your fault, but can we get a docstring for `_try_pypi_api` now that we modified it? I personally don't understand the functionality from the function name.

wolfv:
One more issue over here: https://github.com/prefix-dev/rattler-build-conda-compat/pull/61 will release asap

conda_forge_tick/update_recipe/version.py
```diff
@@ -625,7 +649,95 @@ def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_typ
...
+
+    Parameters
+    ----------
+    feedstock_dir : str
+        The feedstock directory to update.
+    version : str
+        The new version of the recipe.
+    hash_type : str
+        The kind of hash used on the source.
+    """
```
ytausch, line 670: I'd appreciate an explanation of the return value in the docstring here.

conda_forge_tick/update_recipe/version.py
```diff
@@ -625,7 +649,95 @@ def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_typ
...
+                break
+
+        if not found_hash:
+            return None, {"could not find a hash for the source"}
+
+    return recipe, set()
+
+
+def update_version(raw_meta_yaml, version, hash_type="sha256") -> (str, set[str]):
     """Update the version in a recipe.
```
ytausch, line 746:
```suggestion
    """Update the version in a v0 recipe.
```

conda_forge_tick/update_recipe/version.py
wolfv: fixed, thanks!

conda_forge_tick/update_recipe/version.py
```diff
@@ -773,3 +885,32 @@ def update_version(raw_meta_yaml, version, hash_type="sha256"):
...
         logger.critical("Recipe did not change in version migration!")
         return None, errors
+
+
+if __name__ == "__main__":
+    # parse args and invoke update_version_feedstock_dir
+    import argparse
+
+    # set log level to debug
+    logging.basicConfig(level=logging.DEBUG)
```
ytausch, line 900: let's use `setup_logging` from `conda_forge_tick.utils` here

tests/test_migrators.py
```diff
@@ -466,35 +470,63 @@ def run_test_migration(
...
+        tmpdir_p.joinpath("meta.yaml").write_text(inp)
+        recipe_dir = str(tmpdir_p)
+    else:
+        tmpdir_p.joinpath(".ci_support").mkdir()
+        tmpdir_p.joinpath("recipe", "recipe.yaml").write_text(inp)
+        (tmpdir_p / ".ci_support" / "linux_64_.yaml").write_text(
+            "target_platform: linux-64"
+        )
 
+        recipe_dir = str(tmpdir_p / "recipe")
```
ytausch, line 489: Let's be consistent here.

```suggestion
    elif recipe_version == 1:
        tmpdir_p.joinpath(".ci_support").mkdir()
        tmpdir_p.joinpath("recipe", "recipe.yaml").write_text(inp)
        (tmpdir_p / ".ci_support" / "linux_64_.yaml").write_text(
            "target_platform: linux-64"
        )

        recipe_dir = str(tmpdir_p / "recipe")
    else:
        raise ValueError(f"Unsupported recipe version: {recipe_version}")
```

conda_forge_tick/migrators/version.py
h-vetinari: It's weird - your comment wasn't there when I wrote mine, and I got the notification afterwards too. Presumably you had written the comment earlier but only submitted the review later? Nothing important, was just wondering why it looks like I ignored your message 😅

ytausch:
After this is merged, is there anything left for v1 updates to be processed by the bot? Asking because this only touches the `auto-tick` step.

conda_forge_tick/migrators/version.py
h-vetinari: However, the naming here is still bad. Why should v0 get `file` and v1 get a completely different name? Please use consistent naming, like `recipe_file`, `recipe_file_v0` and `recipe_file_v1`.

Something like
```suggestion
        recipe_file = None  # will be determined based on version in use
        recipe_file_v0 = recipe_dir / "meta.yaml"
        recipe_file_v1 = recipe_dir / "recipe.yaml"
        if recipe_file_v0 .exists():
            recipe_file = recipe_file_v0
            [...]
       [...]
       recipe_file.write_text(updated_meta_yaml)
```

conda_forge_tick/update_recipe/version.py
wolfv: Done

conda_forge_tick/migrators/version.py
```diff
@@ -93,11 +93,32 @@ def filter(
...
+            {},
+            0,
+        )
+        print("Checking schema version: %d" % schema_version)
+        if schema_version == 0:
+            if (
+                "raw_meta_yaml" in attrs
+                and "{% set version" not in attrs["raw_meta_yaml"]
+            ):
+                return True
```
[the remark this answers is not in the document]
wolfv: I think part of the point is to make sure that we can access this attribute safely :) I changed the code slightly to make this work

conda_forge_tick/update_recipe/version.py
wolfv: Done

conda_forge_tick/update_recipe/version.py
wolfv: Good idea!

tests/test_migrators.py
wolfv: Alright! Applied suggestion.

wolfv:
@ytausch I believe that version updates should work once this migrator is enabled. Another high-value one would be the arch migrator: https://github.com/regro/cf-scripts/pull/3019

Unfortunately it was blocked because of mini-migrators ...

conda_forge_tick/update_recipe/version.py
wolfv: Nope! :)

conda_forge_tick/update_recipe/version.py
```diff
@@ -897,7 +925,10 @@ def update_version(raw_meta_yaml, version, hash_type="sha256") -> (str, set[str]
     import argparse
 
     # set log level to debug
-    logging.basicConfig(level=logging.DEBUG)
+    from conda_forge_tick.utils import setup_logging
+
+    setup_logging("INFO")
```
ytausch, line 930: didn't you want to set the log level to debug?

wolfv: I think info is fine, too. I left the string there so people can change it easily.

ytausch: ah, then you should adjust the comment above

wolfv: Done!

conda_forge_tick/update_recipe/version.py
```diff
@@ -625,9 +676,103 @@ def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_typ
...
+        The updated meta.yaml. Will be None if there is an error.
+    errors : set of str
+    """
+    # extract all the URL sources from a given recipe / feedstock directory
+    from rattler_build_conda_compat.loader import load_yaml
+    from rattler_build_conda_compat.recipe_sources import render_all_sources
+
+    feedstock_dir = Path(feedstock_dir)
+    recipe_path = feedstock_dir / "recipe" / "recipe.yaml"
+    recipe_yaml = load_yaml(recipe_path.read_text())
```
h-vetinari, line 705: Yet other variables being used here
```suggestion
    recipe_path = feedstock_dir / "recipe" / "recipe.yaml"
    recipe_text = recipe_path.read_text()
    recipe_yaml = load_yaml(recipe_text)
```

conda_forge_tick/update_recipe/version.py
```diff
@@ -625,9 +676,103 @@ def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_typ
...
+        # if there are no variants, then we need to add an empty one
+        variants = [{}]
+
+    rendered_sources = render_all_sources(
+        recipe_yaml, variants, override_version=version
+    )
+
+    # load recipe text
+    recipe_path = feedstock_dir / "recipe" / "recipe.yaml"
+    recipe = recipe_path.read_text()
```
h-vetinari, line 719: redundant with the above
```suggestion
```

conda_forge_tick/update_recipe/version.py
```diff
@@ -625,9 +676,103 @@ def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_typ
...
+
+    # update the version with a regex replace
+    for line in recipe.splitlines():
```
h-vetinari, line 722: suggest to rename `recipe` (very overloaded) to `recipe_text` (more explicit)

conda_forge_tick/update_recipe/version.py
```diff
@@ -557,15 +596,27 @@ def update_version_feedstock_dir(
...
-    updated_meta_yaml, errors = update_version(
-        raw_meta_yaml, version, hash_type=hash_type
-    )
+def _update_version_feedstock_dir_local(
+    feedstock_dir, version, hash_type
+) -> (bool, set):
+    feedstock_path = Path(feedstock_dir)
+
+    recipe = feedstock_path / "recipe" / "meta.yaml"
+    if recipe.exists():
```
h-vetinari, line 605: use the same pattern here
```suggestion
    recipe_path = None
    recipe_path_v0 = feedstock_path / "recipe" / "meta.yaml"
    if recipe_path_v0.exists():
        recipe_path = recipe_path_v0
```

conda_forge_tick/migrators/version.py
```diff
@@ -197,21 +221,34 @@ def migrate(
...
-        updated_meta_yaml, errors = update_version(
-            raw_meta_yaml,
-            version,
-            hash_type=hash_type,
-        )
+        recipe_dir = Path(recipe_dir)
+        recipe = None
+        recipe_v0 = recipe_dir / "meta.yaml"
+        recipe_v1 = recipe_dir / "recipe.yaml"
+        if recipe_v0.exists():
```
h-vetinari, line 228: Please use the same name as in other parts of the implementation. I think `recipe` is too ambiguous anyway, so it's preferable to match the already existing use of `recipe_path` here, rather than vice versa.
```suggestion
        recipe_path = None
        recipe_path_v0 = recipe_dir / "meta.yaml"
        recipe_path_v1 = recipe_dir / "recipe.yaml"
        if recipe_path_v0.exists():
```

h-vetinari (changes requested):
yet more `recipe_*` variable naming consistency issues.

conda_forge_tick/update_recipe/version.py
h-vetinari: might also want to rename `recipe_yaml` to `recipe_dict` or `recipe_info`, because after `load_yaml`, it's not actually a YAML in any reasonable sense of the word anymore.

wolfv: A (ruamel) YAML is also not exactly a dict so might be better to keep it as is.

conda_forge_tick/update_recipe/version.py
wolfv: Done

ytausch:
This might be helpful if the locking bot fails: https://github.com/conda-incubator/relock-conda/issues/245#issuecomment-2497999662

conda_forge_tick/update_recipe/version.py
beckermr: Why was this resolved? We need to remove this code. The little entry points like this rot inplace.

conda_forge_tick/migrators/version.py
```diff
@@ -93,11 +94,34 @@ def filter(
...
-            return True
 
-        conditional = super().filter(attrs)
+        schema_version = get_keys_default(
+            attrs,
+            ["meta_yaml", "schema_version"],
+            {},
+            0,
+        )
+        print("Checking schema version: %d" % schema_version)
```
beckermr, line 104:
```suggestion
```

conda_forge_tick/update_recipe/version.py
```diff
@@ -625,9 +676,103 @@ def _update_version_feedstock_dir_containerized(feedstock_dir, version, hash_typ
...
+    variants = [load_yaml(variant.read_text()) for variant in variants]
+    if not len(variants):
+        # if there are no variants, then we need to add an empty one
+        variants = [{}]
+
+    rendered_sources = render_all_sources(
+        recipe_yaml, variants, override_version=version
+    )
+
+    recipe_text = recipe_path.read_text()
```
h-vetinari, line 717: Please move that line up a bit to where `recipe_yaml` is calculated from it.

wolfv: feel free to just push to this PR :)

conda_forge_tick/migrators/version.py
wolfv: thx!

conda_forge_tick/update_recipe/version.py
beckermr: Yeah I'm happy to wait on this until all comments are addressed. As usual, the folks wanting things done have the onus to do them. I am not 100% happy with how the interactions and pushing in both directions (so me inlcuded) have happened with this PR and v0 -> v1 tasks in general. We need to find a better way to communicate and balance things.

h-vetinari: > feel free to just push to this PR :)

Alright, didn't want to assume, but fine by me. 👍 

> I am not 100% happy with how the interactions and pushing in both directions (so me inlcuded) have happened with this PR

Do you want to elaborate on that? For me it's been pretty much standard review - the work in itself is highly appreciated, and yet at the same time, we can ensure through review-response iterations that newly added code is consistent and otherwise of a high quality before merging.

conda_forge_tick/migrators/version.py
beckermr: Ah I see. The migrator is run in a container so that covers this case. However, because of the way this PR is done currently, the admin commands to update recipe versions won't work for v1. That will need to be fixed if someone wants to use that.

conda_forge_tick/update_recipe/version.py
beckermr: Sure! There has been some back and forth around skipping edge cases / excluding things just to get PRs done. I've also felt like i've been put in the position of blocking half-completed PRs. I'd rather not have to push back at each step and instead have things fully implemented up front. 

I also think I have not been the most responsive on PRs due to my frustration and wanting to avoid the interactions.

wolfv:
@beckermr I used `_get_new_url_tmpl_and_hash` which calls the function you mentioned so I hope that this works :)

wolfv:
It does indeed miss the code about the R versions though, so that will have to be added back in. Will try to tackle this tomorrow if @mgorny doesn't get to it before me

beckermr:
@wolfv That is exactly the problem. We need to ensure `get_transformed_urls` doesn't break on jinja2 with `${{` instead of `{{`.

beckermr:
The lockfile has rattler-build-conda-compat at 1.3.1 and so everything is already up to date.

beckermr:
pre-commit.ci autofix

beckermr:
pre-commit.ci autofix

wolfv:
@beckermr the tests have highlighted an issue with the current `extract_requirements` function. It didn't work well with `pin_subpackage` or `pin_compatible` expressions. These are turned into dictionaries when rendering, so that the result looks something like:

```yaml
- foo
- pin_subpackage:
    name: foobar
    pin:
      upper_bound: x.x
...
```

This ends up not being hashable, and then the extraction failed somewhere. 
For now, I fixed it by turning this dictionary into a normal looking spec of `pin_subpackage {name}`

Do you happen to know what the best thing to turn this into would be? I could also just make it a requirement on the name that comes out of this.

I also added the `split` filter to the jinja env unconditionally to help the v1 recipes. I hope that's OK.

beckermr:
Those need to be dictionaries for other parts of the code to work. So whatever requires them to be hashable needs to be changed. We can't merge this pr as it stands now.

wolfv:
> We can't merge this pr as it stands now.

I didn't ask for this, I just asked for guidance :)

beckermr:
Ah sorry. I meant to indicate this is blocking. I suggest if rattler is that strict about its inputs that the v1 version update code should make a copy of the data it sends in, adjust it, and then undo the changes on the way out.

wolfv:
Sorry, I think there is a misunderstanding. What I am fixing here is code that was already present for a long time in the graph generation. This is not related to the functionality of this PR, nor rattler-build. 

I think we didn't properly test recipes with `pin_subpackage` before, and now this started to highlight a problem in the graph generation. The failure came about in this line:

https://github.com/regro/cf-scripts/pull/3525/files#diff-80dce555ca3e04ee9e79590c4b418bad646f6ff177d8332104d68c78ce75ff3cR195

Does the code really expect this to be a dictionary in other places? Ie. 

```
run:
  - pin_subpackage:
      name: foo
      pin: { bla ..}
```

As I don't think conda-build would create specs in this way.

beckermr:
Ah thanks for asking a good question. The code parses the name only usually (https://github.com/wolfv/cf-scripts/blob/add-version-bump-v1/conda_forge_tick/utils.py#L58) but in some cases will output a dictionary (https://github.com/wolfv/cf-scripts/blob/add-version-bump-v1/conda_forge_tick/utils.py#L78) when dealing with pinnings. I was under the impression we were working with the pinnings parsing and we wanted to change it. However, we should not be using that at all, so I think the bug is somewhere else.

conda_forge_tick/feedstock_parser.py
beckermr, line 1: I don't the changes related to flattening dicts in this file are actually needed. I think there is a bug somewhere that is causing the parsing of v1 recipes to output the special pinning parsing options when it should not. In this function here: https://github.com/regro/cf-scripts/blob/main/conda_forge_tick/utils.py#L612 the dicts should be output only if `for_pinning` is set to true.

wolfv:
If I see correctly, then what ends up in the graph is just a simple package name: https://github.com/regro/cf-graph-countyfair/blob/8c12b88aa984748c8927302d87b4f898e32f72dd/node_attrs/6/b/e/9/0/zlib.json#L142-L145

(I believe the info comes from run_exports / pin_subpackage).

beckermr:
Yes. So there is a bug in the v1 parsing that is adding dicts in a case it should not. Sometimes we do output dicts, but only when for_pinning=True.

ytausch:
Hi, I am a bit out of the loop about what happened in this PR but please let me shamelessly advertise that #3417 adds a way to test version updates using the entire's bot pipeline, including actual PRs made to GitHub.

It's likely easy to secure this PR against subtle bugs if #3417 gets merged first.

wolfv:
Cool! I think the test migration that is ran in the tests of this PR are also quite good but the deeper the tests the better! 

Does your test also check if new versions are "found"? The tests in this PR only test the recipe editing essentially.

ytausch:
Yes, they start from an empty cf-graph and execute all the CLI entrypoints in sequence.
The search for new updates is mainly done by `update-upstream-versions` which is part of that :)

beckermr:
We'll want both. The integration tests handle items that happen when the bot runs live (git checkouts, etc.). However, it is faster to debug certain things with a more targeted test like is present in this PR. There is no need to couple the two.

conda_forge_tick/update_recipe/version.py
wolfv: I removed it now. I only removed similar code in another file last time - a mistake!

wolfv:
I just released `rattler-build-conda-compat` 1.3.2 which should fix the remaining errors. It will take a little more time to propagate through the CDNs etc. But I think then this should be ready :)

wolfv:
Nice, tests passed 🙏

conda_forge_tick/update_recipe/version.py
```diff
@@ -134,14 +135,40 @@ def _try_url_and_hash_it(url: str, hash_type: str):
 
 
 def _render_jinja2(tmpl, context):
-    return (
-        jinja2.sandbox.SandboxedEnvironment(undefined=jinja2.StrictUndefined)
-        .from_string(tmpl)
-        .render(**context)
-    )
+    env = jinja2.Environment(undefined=jinja2.StrictUndefined)
```
beckermr, line 138:
```suggestion
    env = jinja2.sandbox.SandboxedEnvironment(undefined=jinja2.StrictUndefined)
```

This needs to be sandboxed.

beckermr (approved):

wolfv:
Wow, thanks @beckermr! 

Do you know if anything else needs to happen?

beckermr:
Not sure. We'll find out if we get some version bumps!

wolfv:
I asked the bot for one here: https://github.com/conda-forge/conda-forge-webservices/actions/runs/12795837076/job/35674099360 

Do we need to update a docker image somewhere?

beckermr:
so to make that go live, you'll need to

1. bump conda-forge-tick-feedstock to the latest version (make sure to adjust the constraints in the recipe on rattler build conda compat when doing that)
2. update the webservices conda env

If you do 1, 2 will happen automatically.

wolfv:
OK, I bumped the conda-forge-tick-feedstock: https://github.com/conda-forge/conda-forge-tick-feedstock/pull/60

ytausch:
appears to work: https://github.com/conda-forge/subfinder-feedstock/pull/1/files

minrk:
absolutely wonderful, thanks folks!

PR #3562, opened 2025-01-14
[minor fix] fix GitHub Actions version comments

ytausch:
#### Description:

The comments behind the hash are not version pins, but solely intended as human-readable translation. This PR updates all external GitHub Actions to the latest compatible version and changes the comment to the full version string. Dependabot supports updating this.

We also now use `supercharge/mongodb-github-action` because it is actively maintained and functionally equivalent to the Action we used before.

#### Checklist:


ytausch:
@pavelzw could you review this, please?

pavelzw (approved):

PR #3564, opened 2025-01-14
feat: test and fix the dict unhashable error in isolation

wolfv:
I reopen a subset of #3525 to first make sure that the graph creation code works without issues. 

I added a little test fixture to create a "feedstock" easily which is necessary to properly and accurately render v1 feedstocks. 

This reproduces the issue we saw with un-hashable dicts in the `populate_feedstock_attributes` function. To recap, this is triggered by extracting dependencies with `pin_subpackage` or `pin_compatible`. 

I'll move on to fixing this now.


beckermr:
The bug is in the code in utils.py is my guess.

beckermr:
Specifically the function parse_recipe_yaml.

wolfv:
Hey @beckermr - this code fixes the issue at hand. The failure in CI seems unrelated (FAILED tests/test_url_exists.py::test_url_exists[http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-7.6p1.tar.gz-True] - AssertionError: assert False is True) - or at least idk how it would be related.

Let me know what you think? The main change is to normalize away the dictionaries from the v1 recipe parsing.

beckermr:
This is the wrong fix and is a patch over another bug IIUIC. From the failure, The function parse_recipe_yaml is returning dicts when for_pinning is False. We need to fix that and then the rest of these changes are not needed.

wolfv:
Thanks @beckermr now I finally understand what you mean :) I can try to fix it there indeed.

beckermr:
Thank you and sorry I was not clear enough!

wolfv:
Opened #3565 for the alternative approach.

PR #3580, opened 2025-01-17
NEW: Migrate NVIDIA redist feedstocks to use cf-nvidia-tools

carterbox:
#### Checklist:

### DIFF
--- a/conda_forge_tick/make_migrators.py
+++ b/conda_forge_tick/make_migrators.py
@@ -75,6 +75,7 @@ from conda_forge_tick.migrators import (
     XzLibLzmaDevelMigrator,
     make_from_lazy_json_data,
     skip_migrator_due_to_schema,
+    AddNVIDIATools,
 )
 from conda_forge_tick.migrators.arch import OSXArm
 from conda_forge_tick.migrators.migration_yaml import (
@@ -794,7 +795,11 @@ def initialize_migrators(
     gx: nx.DiGraph,
     dry_run: bool = False,
 ) -> MutableSequence[Migrator]:
-    migrators: List[Migrator] = []
+    migrators: List[Migrator] = [
+        AddNVIDIATools(
+            check_solvable=False,
+        ),
+    ]
 
     add_arch_migrate(migrators, gx)
 
--- a/conda_forge_tick/migrators/nvtools.py
+++ b/conda_forge_tick/migrators/nvtools.py
@@ -1,0 +1,245 @@
+import os.path
+import logging
+import copy
+from typing import Any
+
+from conda_forge_tick.contexts import ClonedFeedstockContext
+import conda_forge_tick.migrators
+
+from conda_forge_tick.migrators_types import AttrsTypedDict, MigrationUidTypedDict
+from conda_forge_tick.utils import (
+    yaml_safe_dump,
+    yaml_safe_load,
+    get_bot_run_url,
+)
+
+
+def _file_contains(filename, string):
+    with open(filename, "r") as f:
+        return string in f.read()
+
+
+def _insert_requirements_build(meta):
+    insert_before: bool = True
+    before: list[str] = []
+    after: list[str] = []
+    requirements_found = False
+    build_found = False
+    with open(meta, "r") as f:
+        for line in f:
+            if build_found:
+                insert_before = False
+            elif line.startswith("requirements"):
+                requirements_found = True
+            elif requirements_found and line.lstrip().startswith("build"):
+                build_found = True
+
+            if insert_before:
+                before.append(line)
+            else:
+                after.append(line)
+
+    if not (build_found or requirements_found):
+        return
+
+    with open(meta, "w") as f:
+        f.writelines(before + ["    - cf-nvidia-tools  # [linux]\n"] + after)
+
+
+def _insert_build_script(meta):
+    insert_before: bool = True
+    before: list[str] = []
+    after: list[str] = []
+    script_found = False
+    build_found = False
+    with open(meta, "r") as f:
+        for line in f:
+            if not insert_before:
+                pass
+            elif line.startswith("build"):
+                build_found = True
+            elif build_found and line.lstrip().startswith("script"):
+                script_found = True
+            elif build_found and script_found:
+                if line.lstrip().startswith("-"):
+                    # inside script
+                    pass
+                else:
+                    # empty script section?
+                    insert_before = False
+            if insert_before:
+                before.append(line)
+            else:
+                after.append(line)
+
+    if not (script_found or build_found):
+        return
+
+    with open(meta, "w") as f:
+        f.writelines(
+            before
+            + ['    - check-glibc "$PREFIX"/lib/*.so.* "$PREFIX"/bin/*  # [linux]\n']
+            + after
+        )
+
+
+class AddNVIDIATools(conda_forge_tick.migrators.Migrator):
+    """Add the cf-nvidia-tools package to NVIDIA redist feedstocks.
+
+    In order to ensure that NVIDIA's redistributed binaries (redists) are being packaged
+    correctly, NVIDIA has created a package containing a collection of tools to perform
+    common actions for NVIDIA recipes.
+
+    At this time, the package may be used to check Linux binaries for their minimum glibc
+    requirement in order to ensure that the correct metadata is being used in the conda
+    package.
+
+    This migrator will attempt to add this glibc check to all feedstocks which download any
+    artifacts from https://developer.download.nvidia.com. The check involves adding
+    something like:
+
+    ```bash
+    check-glibc "$PREFIX"/lib/*.so.* "$PREFIX"/bin/*
+    ```
+
+    to the build script after the package artifacts have been installed.
+
+    > [!NOTE]
+    > A human needs to verify that the glob expression is checking all of the correct
+    > artifacts!
+
+    More information about cf-nvidia-tools is available in the feedstock's
+    [README](https://github.com/conda-forge/cf-nvidia-tools-feedstock/tree/main/recipe).
+
+    Please ping carterbox for questions.
+    """
+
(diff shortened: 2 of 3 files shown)


beckermr (dismissed):
Why is check solvable turned to false?

carterbox:
@beckermr, I guessed that this parameter means that the bot will trying to update the nodes in a certain order, but this migrator can be applied in any order. Am I interpreting this parameter incorrectly?

beckermr:
check_solvable controls whether or not the bot ensures that the feedstock environments are solvable before it tries to issue a PR.

carterbox:
This migrator shouldn't be affecting the solvability of a recipe's environments because it only adds one package to the build environment. If an environment is unsolvable, I will want the PR opened anyways, so that a human can fix the feedstock.

carterbox:
I have tested my migration function locally, but not the filtering logic. Is there a sandbox for developing migrators, or a way to rate limit the migration?

carterbox:
@jakirkham, please review

carterbox:
Do not merge. Just actively looking for feedback at this point.

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+)
+
+
+def _file_contains(filename, string):
+    with open(filename) as f:
+        return string in f.read()
+
+
+def _insert_requirements_build(meta):
+    insert_before: bool = True
```
jakirkham, line 22: Think we can add this after compilers and like (assuming I'm understanding the intent of this parameter correctly)

```suggestion
    insert_before: bool = False
```

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+    before: list[str] = []
+    after: list[str] = []
+    requirements_found = False
+    build_found = False
+    with open(meta) as f:
+        for line in f:
+            if build_found:
+                insert_before = False
+            elif line.startswith("requirements"):
```
jakirkham, line 31:
```suggestion
            if line.startswith("requirements"):
```

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+            elif requirements_found and line.lstrip().startswith("build"):
+                build_found = True
+
+            if insert_before:
+                before.append(line)
+            else:
+                after.append(line)
+
+    if not (build_found or requirements_found):
+        return
```
jakirkham, line 42: So is the idea of this check that we are skipping metapackages? Or are there other cases this would be used for?

May help to have a comment above clarifying intent

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+
+    with open(meta, "w") as f:
+        f.writelines(before + ["    - cf-nvidia-tools  # [linux]\n"] + after)
+
+
+def _insert_build_script(meta):
```
jakirkham, line 48: What do you think about adding this to the recipe's `test/commands` section?

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # STEP 3: Remove os_version keys from conda-forge.yml
+        config = os.path.join(recipe_dir, "..", "conda-forge.yml")
+        with open(config) as f:
+            y = yaml_safe_load(f)
+        y_orig = copy.deepcopy(y)
+        y.pop("os_version", None)
+        if y_orig != y:
+            with open(config, "w") as f:
+                yaml_safe_dump(y, f)
```
jakirkham, line 206: So we are dropping the `os_version` specification of the Docker image to use?

Wonder if it makes to add/confirm `{{ stdlib("c") }}` is used and `c_stdlib_version` is set correctly

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+        body = f"{AddNVIDIATools.__doc__}\n\n"
+
+        if add_label_text:
+            body += (
+                "If this PR was opened in error or needs to be updated please add "
+                "the `bot-rerun` label to this PR. The bot will close this PR and "
+                "schedule another one. If you do not have permissions to add this "
+                "label, you can use the phrase "
+                "<code>@<space/>conda-forge-admin, please rerun bot</code> "
+                "in a PR comment to have the `conda-forge-admin` add it for you.\n\n"
```
jakirkham, line 230: Wonder if we should add some text on how to handle issues or ask for help in resolving

jakirkham (commented):
Thanks Daniel! 🙏

Appreciate you putting this together so quickly 😄

Had a couple questions below

conda_forge_tick/migrators/nvtools.py
carterbox: Yes. It's not necessary to force use an old container.

The tool fails if c_stdlib_version is not set, so indirectly this is already checked?

conda_forge_tick/migrators/nvtools.py
carterbox: Please look at line 113. The description directs users (us) to ping me for questions.

conda_forge_tick/migrators/nvtools.py
carterbox: This is not a function parameter. It is a variable which holds the state of "whether we are currently inserting lines into the variable named 'before'".

We're not going to do that because trying to locate the compiler packages and insert the new dependency after them is technically challenging and not necessary for a working migration. Adding the complexity of inserting after compiler packages instead of just inserting into requirements.build makes additional assumptions including 1. that the package has compiler packages listed in the top-level requirements.build section 2. the compiler packages are already sorted to the top of the requirements.build section

conda_forge_tick/migrators/nvtools.py
carterbox: This was intentional. Once you find the requirements.build section, you don't need to do any more searching for it because you already know where to split the text. Maybe I could refactor to not execute insert_before=False, but I'm not sure it's worth the effort?

conda_forge_tick/migrators/nvtools.py
carterbox: If we didn't find the requirements.build section, then we cannot add our new dependency to the file.

carterbox:
@jakirkham, I have refactored the item insertion logic. Hopefully it is now more easily groked.

beckermr:
@carterbox Let me know when you want review on this. I am not sure what the status is currently.

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,240 @@
+import copy
+import logging
+import os.path
+from typing import Any
+
+import conda_forge_tick.migrators
```
carterbox, line 6: This is probably the circular import.

carterbox:
- [ ] Add major version to cf-nvidia-tools pinning

- [ ] Update check-glibc calls with new globs

conda_forge_tick/make_migrators.py
```diff
@@ -797,7 +798,11 @@ def initialize_migrators(
     gx: nx.DiGraph,
     dry_run: bool = False,
 ) -> MutableSequence[Migrator]:
-    migrators: List[Migrator] = []
+    migrators: List[Migrator] = [
```
beckermr, line 801: We should append this migrator to the list after the noarch python migrator. There are some conventions in the code about where each kind of migrator goes in the list.

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,243 @@
...
+            If the 'bad' notice starts with the string then it is not
+            to be excluded. For example, rebuild migrations don't need
+            to worry about if the upstream can be fetched. Defaults to ``''``
+
+        Returns
+        -------
+        bool :
+            True if node is to be skipped
+        """
+        return (
```
beckermr, line 132: the filter method always needs calls to super and to check the schema version. Check some of the other migrators for examples.

beckermr (approved):

beckermr:
@jakirkham any more comments here?

jakirkham (commented):
Thanks Daniel! 🙏

Had some minor suggestions. Otherwise think this would be good to get in

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+        attrs : dict
+            The node attributes
+
+        Returns
+        -------
+        namedtuple or bool:
+            If namedtuple continue with PR, if False scrap local folder
+        """
+        # STEP 0: Bump the build number
+        self.set_build_number(os.path.join(recipe_dir, "meta.yaml"))
```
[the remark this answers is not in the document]
carterbox: https://github.com/regro/cf-scripts/pull/3580/commits/7116e83df0ae0b9bc3c569c72fc73a78ba080342

conda_forge_tick/migrators/nvtools.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # STEP 1: Add cf-nvidia-tools to build requirements
+        meta = os.path.join(recipe_dir, "meta.yaml")
```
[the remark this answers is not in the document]
carterbox: https://github.com/regro/cf-scripts/pull/3580/commits/7116e83df0ae0b9bc3c569c72fc73a78ba080342

carterbox:
Ready for merging. Thanks, all!

jakirkham (approved):

beckermr:
Ack. I missed a key bit here. The migrator classes are trick things and we missed a required init step in a subclass. This all needs cleanup in the bot, but for now, I am going to revert this PR since it is breaking things. I will revert twice so that we can work on a new PR and add testing to catch this next time.

PR #3764, opened 2025-02-25
add support for unusual version strings in GitHubReleases source

### LINKED ISSUE #3763: GitHubReleases version source does not accept unusual version strings like ISO timetamps
Hi,

the version string format of the popular [minio](https://github.com/minio/minio) package looks like this: `RELEASE.2025-02-18T16-25-55Z`.

Because the GitHubReleases update source [expects](https://github.com/regro/cf-scripts/blob/907eac7cb586de310105bfef6dd3a066b95afe09/conda_forge_tick/update_sources.py#L609) the string retrieved from the git tag associated with the latest GitHub release to be a valid Python version string, the autotick-bot does not find version updates for [minio-server-feedstock](https://github.com/conda-forge/minio-server-feedstock).

While debugging, I found two issues:

1. The git tag associated with the latest GitHub release must be a valid Python version string (as described above)
2. The autotick-bot extracts the version string by starting at the first numeral of the git tag associated with the latest GitHub release. See [here](https://github.com/regro/cf-scripts/blame/907eac7cb586de310105bfef6dd3a066b95afe09/conda_forge_tick/update_sources.py#L616-L617).

To solve this issue, I propose to solve 1.) by also allowing such version strings that are not valid Python version strings in the GithubReleases source. I've done that in this PR: #3764

2) can be solved by changing the recipe such that only the timestamp part `2025-02-18T16-25-55Z` is part of the version variable. Do you agree @beckermr?

The Github source has a similar issue but I suggest to not touch it because I don't see a need to do so.

Cc @pavelzw

ytausch:
#### Description:

Closes #3763.

#### Checklist:

### DIFF
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -9,9 +9,10 @@ import subprocess
 import typing
 import urllib.parse
 from pathlib import Path
-from typing import Iterator, List, Optional
+from typing import Iterator, List, Literal, Optional
 
 import feedparser
+import packaging.version
 import requests
 import yaml
 from conda.models.version import VersionOrder
@@ -598,7 +599,7 @@ class GithubReleases(AbstractSource):
         owner, repo = meta_yaml["url"].split("/")[3:5]
         return f"https://github.com/{owner}/{repo}/releases/latest"
 
-    def get_version(self, url: str) -> Optional[str]:
+    def get_version(self, url: str) -> Optional[str | Literal[False]]:
         r = requests.get(url)
         if not r.ok:
             return False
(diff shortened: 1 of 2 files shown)


jaimergp (approved):

conda_forge_tick/update_sources.py
```diff
@@ -598,16 +599,24 @@ def get_url(self, meta_yaml) -> Optional[str]:
...
-        if not len(latest) or latest == "latest" or parse_version(latest).is_prerelease:
-            return False
+        try:
+            if (
+                not len(latest)
+                or latest == "latest"
+                or parse_version(latest).is_prerelease
+            ):
+                return False
+        except packaging.version.InvalidVersion:
```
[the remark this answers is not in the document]
ytausch: Yeah I think we somehow have to support use cases like minio. Thank you!

conda_forge_tick/update_sources.py
```diff
@@ -598,16 +599,24 @@ def get_url(self, meta_yaml) -> Optional[str]:
...
             return False
         # "/releases/latest" redirects to "/releases/tag/<tag name>"
         url_components = r.url.split("/")
         latest = "/".join(url_components[url_components.index("releases") + 2 :])
         # If it is a pre-release don't give back the pre-release version
-        if not len(latest) or latest == "latest" or parse_version(latest).is_prerelease:
-            return False
+        try:
+            if (
+                not len(latest)
```
beckermr, line 612:
```suggestion
                len(latest) == 0
```

beckermr: This is clearer to python n00bs.

PR #3765, opened 2025-02-26
Enable `CondaForgeYAMLCleanup` migrator for v1 recipes

mgorny:
#### Description:

Enable `CondaForgeYAMLCleanup` migrator for v1 recipes.  Since it works on `conda-forge.yml` only, it does not require any changes.  So the most of the changes are related to extending the test.

#### Checklist:

### DIFF
--- a/conda_forge_tick/migrators/conda_forge_yaml_cleanup.py
+++ b/conda_forge_tick/migrators/conda_forge_yaml_cleanup.py
@@ -26,6 +26,9 @@ class CondaForgeYAMLCleanup(MiniMigrator):
 
     def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
         """remove recipes without a conda-forge.yml file that has the keys to remove or change"""
+        if super().filter(attrs):
+            return True
+
         cfy = attrs.get("conda-forge.yml", {})
         if any(key in cfy for key in (self.keys_to_remove + self.keys_to_change)):
             return False
--- a/conda_forge_tick/migrators/conda_forge_yaml_cleanup.py
+++ b/conda_forge_tick/migrators/conda_forge_yaml_cleanup.py
@@ -12,6 +12,7 @@ if typing.TYPE_CHECKING:
 
 
 class CondaForgeYAMLCleanup(MiniMigrator):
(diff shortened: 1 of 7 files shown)


mgorny:
@beckermr, the v0/v1 testing code is becoming horrible. Would you mind me if I tried converting all v0 tests to use the same tempdir approach as v1 — i.e. passed top-level tempdir rather than the `recipe` subdir?

I imagine it's going to be a lot of work, but it should make everything cleaner. Perhaps going in the other direction (i.e. making v1 tests use `recipe` subdir) would be less work, but I find the resulting backreferences kinda iffy.

ytausch:
If you do that, could you also migrate to `tmp_path`? `tmpdir` returns a `py.path.local` object, which really is outdated Python.

mgorny:
> If you do that, could you also migrate to `tmp_path`? `tmpdir` returns a `py.path.local` object, which really is outdated Python.

Sure can do. Does that mean it's fine for me to proceed, or should I wait for more ACKs?

beckermr:
Yes the testing code is fiddly in general, so improvements are welcome.

beckermr (changes requested):
You need to add 1 to the allowed schema version for the migrator as well IIRC.

mgorny:
Ah, sorry, removed it while debugging the tests, and forgot to put it back. Which means now I'm going to have to debug why the tests pass…

beckermr (approved):

beckermr:
Shall we merge this one @mgorny?

mgorny:
Yes, please.

mgorny:
That said, `tests/test_qt_qt_main.py` seems to be suffering from network problems now.

It would probably be also worth looking into mocking some network interactions in the test suite too.

beckermr:
I added network retries to that one to ensure it passes. We can up those maybe. I do not think that one can be mocked.

mgorny:
> I added network retries to that one to ensure it passes. We can up those maybe. I do not think that one can be mocked.

Right. I suppose we could try mocking the whole logic that grabs the hash, but perhaps that'd go too far.

beckermr:
I am so sorry! I squashed which will make for a bad conflict on your other PR. 

🤦

PR #3774, opened 2025-02-27
Always pass feedstock dir `Path` to migrator tests

mgorny:
**(work-in-progress posted to facilitate early feedback)**

#### Description:

Update the `run_test_migration()` function to always expect the top feedstock directory, passed as `Path`, for both schema versions. Previously, the function would expect feedstock directory for v1 recipes, and the recipe subdirectory for v0 recipes.  This improves consistency, as well as modernizes the code.

While at it, rewrite the relevant tests to use the modern `tmp_path` fixture rather than the deprecated `tmpdir` fixture.

#### Checklist:

Discussed in #3765.

### DIFF
--- a/conda-lock.yml
+++ b/conda-lock.yml
@@ -2110,39 +2110,39 @@ package:
   - category: main
     dependencies: {}
     hash:
-      md5: f825788b43bd8901934a6e671ce3a6e2
-      sha256: 0ce0bd2531873a81320b518d92f1e380aa5764b5bd69033e7b5330d36d52af96
+      md5: fbcc4a2a822659042189f814fbfed6fd
+      sha256: 121267b4c52f442d165396b8434716eab5ba339e18817990f3e0074322f08a4a
     manager: conda
     name: conda-forge-pinning
     optional: false
     platform: linux-64
     url:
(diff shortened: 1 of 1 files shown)


mgorny:
I've rebased it on #3765 to avoid conflicts when it's merged. It's mostly done, still need to update the API for minimigrator and YAML migrator tests.

beckermr:
I think I got the conflicts here. sorry for the squash!

mgorny:
> I think I got the conflicts here. sorry for the squash!

No problem, I treat the followup commits as dispensable.

mgorny:
> I have two questions.
> 
>     1. When do we need to call `str()` on `Path` objects? It is not always done in some spots versus others (e.g., when passing a path to the migrator methods) and I do not understand why.

I've tried to put it to cf-scripts' functions where `str` was passed previously, to avoid type mismatches. However, I've left `Path` objects where they work in stdlib, or where they were already passed.

>     2. The use of `/` in strings for joining paths like this `pathobj / "foo/bar"` appears like it would break using the test suite on windows. Is that true? If so, do we care?

That's the beauty of it — `pathlib` internally uses forward slashes, so everything works as expected.

```py
>>> from pathlib import PureWindowsPath
>>> str(PureWindowsPath(r"c:\some\directory") / "foo/bar")
'c:\\some\\directory\\foo\\bar'
```

beckermr (approved):

PR #3783, opened 2025-02-28
Some more progress towards recipe v1 minimigrators

mgorny:
#### Description:

- enable `Cos7Config` on v1 recipes
- move test data for `test_cross_compile` into files (but without actually updating the migrator, after seeing how much code I'd need to update)
- mark `DuplicateLinesCleanup` migrator as incompatible with v1 recipes (duplicate keys are an error there)

#### Checklist:


beckermr (changes requested):
When moving around recipes, please be careful that the tests of the recipe parser still find all of the examples. I'll look closer later.

beckermr:
This line here in particular is what I am concerned about: https://github.com/regro/cf-scripts/blob/main/tests/test_recipe_parser.py#L1612

mgorny:
> This line here in particular is what I am concerned about: https://github.com/regro/cf-scripts/blob/main/tests/test_recipe_parser.py#L1612

Ah, sorry, didn't expect that. Do you prefer that I move them back to top-level, or modify this test to look recursively?

beckermr:
We should modify the list to look recursively.

beckermr (approved):
Thank you!

mgorny:
Thank you!

PR #3802, opened 2025-03-03
fix: remove extra v1 testing code and enable replacement migrators

beckermr:
#### Description:

Enables replacement migrations for v1 and fixes v1 testing bugs.

#### Checklist:


tests/test_migrators_v1.py
```diff
@@ -132,6 +56,7 @@ def test_yaml_migration_rebuild(tmp_path):
         },
         tmp_path=tmp_path,
         is_v1=True,
```
beckermr, line 58:
```suggestion
```

PR #3808, opened 2025-03-04
Support recipe v1 in Build2HostMigrator

mgorny:
#### Description:

Add support for v1 recipes in `Build2HostMigrator`, and extend the relevant tests. I've also replaced the explicit schema version check with a `super()` call.

#### Checklist:

From the list on #3642

### DIFF
--- a/conda_forge_tick/migrators/cross_compile.py
+++ b/conda_forge_tick/migrators/cross_compile.py
@@ -240,13 +240,16 @@ class UpdateCMakeArgsMigrator(CrossCompilationMigratorBase):
 
 
 class Build2HostMigrator(MiniMigrator):
+    allowed_schema_versions = {0, 1}
     post_migration = False
 
     def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
+        if super().filter(attrs, not_bad_str_start):
+            return True
+
         build_reqs = attrs.get("requirements", {}).get("build", set())
         host_reqs = attrs.get("requirements", {}).get("host", set())
         run_reqs = attrs.get("requirements", {}).get("run", set())
-        skip_schema = skip_migrator_due_to_schema(attrs, self.allowed_schema_versions)
         if (
             len(attrs.get("outputs_names", [])) <= 1
             and "python" in build_reqs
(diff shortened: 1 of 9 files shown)


conda_forge_tick/migrators/cross_compile.py
```diff
@@ -261,15 +264,19 @@ def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
...
             return False
         else:
             return True
 
     def migrate(self, recipe_dir: str, attrs: "AttrsTypedDict", **kwargs: Any) -> None:
+        recipe_file = (
+            "recipe.yaml"
+            if attrs["meta_yaml"].get("schema_version", 0) == 1
+            else "meta.yaml"
+        )
```
beckermr, line 277: This will be more robust if we simply test for file existence.

tests/test_cross_compile.py
```diff
@@ -221,11 +226,14 @@ def test_cross_python_no_build(tmp_path):
 
 
 @flaky
-def test_build2host(tmp_path):
+@pytest.mark.parametrize("recipe_version", [0, 1])
+def test_build2host(recipe_version, tmp_path):
```
beckermr, line 230: I think we might want flaky inside the pytest parameterize as opposed to outside of it.

tests/test_cross_compile.py
```diff
@@ -234,15 +242,21 @@ def test_build2host(tmp_path):
...
         },
         tmp_path=tmp_path,
+        recipe_version=recipe_version,
     )
 
 
 @flaky
-def test_build2host_buildok(tmp_path):
+@pytest.mark.parametrize("recipe_version", [0, 1])
+def test_build2host_buildok(recipe_version, tmp_path):
```
beckermr, line 251: Ditto here.

tests/test_cross_compile.py
```diff
@@ -251,15 +265,21 @@ def test_build2host_buildok(tmp_path):
...
         },
         tmp_path=tmp_path,
+        recipe_version=recipe_version,
     )
 
 
 @flaky
-def test_build2host_bhskip(tmp_path):
+@pytest.mark.parametrize("recipe_version", [0, 1])
+def test_build2host_bhskip(recipe_version, tmp_path):
```
beckermr, line 274: ditto here

beckermr (changes requested):
A few things but otherwise looks great! Thank you!

isuruf (changes requested):
These migrators are for easy transition from conda-build 2 to 3. I think we should drop this mini-migrator.

mgorny:
Well, either way is fine with me. I've figured out it makes sense to keep it, in case people `conda-recipe-manager convert` an old recipe to v1. Technically, I suppose with v1 we could rewrite it to use `ruamel.yaml`, but I don't think there's a point in spending that much effort on it.

beckermr:
I agree it can be dropped. For that, we need to remove it from the various bits of the make_migrators file. Once we do that we can merge.

beckermr (approved):

PR #3845, opened 2025-03-09
Add a new migrator to combine `if` clauses in v1 recipes

### LINKED ISSUE #3833: New migrator: combine adjacent conditions in v1 recipes
As noted in #3832, I think it would make sense to have a migrator to combine adjacent conditions in v1 recipes into one.

For example, the following:

```yaml
      then: python
      then: cross-python_${{ target_platform }}
      then: cython
```

could be combined into:

```yaml
      then:
        - python
        - cross-python_${{ target_platform }}
        - cython
```

This would both be useful as post-migrator for other migrators (e.g. `CrossPythonMigrator`) that can produce such output, and to clean up recipes created via `conda-recipe-manager`.

If this sounds fine, I can try implementing my first minimigrator :-).

CC @wolfv

mgorny:
#### Description:

Create a new migrator that attempts to combine adjacent `if` clauses in v1 recipes.  The main use cases include running as a post-processor to other migrators such as `CrossPythonMigrator` that could yield multiple adjacent conditions, as well as cleaning up the result of running `conda-recipe-manager convert` that replaces inline selectors on adjacent lines with separate conditions.

The algorithm is slightly more complex than I wished it could be, but also simpler than I worried it would become.  As such, it has fair limitations: it can't combine non-adjacent conditions, or combine most of the sets featuring both `then` and `else` clauses.  Still, it should be able to save some manual work combining clauses, and bring the recipe closer to the desired state.

As a test case, I've included a converted and modified PyTorch recipe. It features some specific corner cases, including nested conditions, mixed `then`/`else` branches, inline and list arguments to them, and inversed order (`else` before `then`).

#### Checklist:

Fixes #3833


beckermr:
pre-commit.ci auotfix

mgorny:
Heh:

```
  Unable to find image 'mongo:latest' locally
  docker: Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit.
```

beckermr:
A retry will fix that one. We should remove the mongodb testing or move it to another job anyways.

conda_forge_tick/migrators/recipe_v1.py
```diff
@@ -0,0 +1,159 @@
...
+                    clause_order = [condition.thens, condition.elses]
+                    if condition.elses.printed:
+                        clause_order.reverse()
+                    for clause in clause_order:
+                        yield from flush_clause(clause, condition)
+                    conditions.pop()
+
+                # a top-level condition
+                if line.lstrip().startswith("- if:"):
+                    indent = line.split("-")[0] + "  "
```
beckermr, line 151: The indent is described as

```
    # base indent for the condition (i.e. up to "- if")
```

Thus I do not get why an extra space has been added.

mgorny: Ah. That was because, at least to my understanding, the following YAML is not valid:

```yaml
  - if: ...
  then: ...
```

In fact, FWICS you need at least two spaces, but I've figured out one is good enough to distinguish continuation from next item.

beckermr (changes requested):
I am concerned there are assumptions about 2-space versus n-space yaml indenting in the code. 

Also, and this comment doesn't need to be addressed but is just me thinking, I wonder if an approach of using a breadth-first recursive tree-traversal might allow us to more flexibly combine conditions.

conda_forge_tick/migrators/recipe_v1.py
beckermr: I am confused though. The indent to the start of `- if` is `line.split("-")[0]`:

```python
>>> line = "    - if: blah blah"
>>> print(len(line.split("-")[0]))
    4
```

beckermr: Also more fun:

```python
data = """\
run:
  - if: blah
    then: blah
# all of these fail
#   -   if: blah blah
#     then: blah
#   - if: blah blah
#       then: blah
# this is OK
  -   if: blah blah
      then: blah
"""

yaml.safe_load(data)
```

Apparently any level of indent to the mapping is allowed and it can be mixed between list entries. The only requirement is that it is always the same

mgorny:
> Also, and this comment doesn't need to be addressed but is just me thinking, I wonder if an approach of using a breadth-first recursive tree-traversal might allow us to more flexibly combine conditions.

I'm sorry but I'm, err, a "gut programmer". I don't have much clue about algorithms. I mostly let my cats run on the keyboard.

conda_forge_tick/migrators/recipe_v1.py
beckermr: I don't understand the code and I am concerned it assumes 2-space indenting. For example, if someone passes this

```
  -    if: 
       then:
```

then the indent for then `then` block is not `line.split("-")[0] + "  "`.

beckermr: Further the computed indent above doesn't match the comment in the code.

mgorny:
So, I'm not sure whether that's the thing you were thinking of, but my feverish brain came up with the following idea through the night:

1. Determine all possible "indents" at which conditions can be found.
2. For every "indent", find and replace all adjacent conditions.
3. Proceed with the next "indent" at updated recipe.

I suppose this would have the following advantages:

- The algorithm could be less convoluted and more reliable.
- Since we're going breadth first, combining outer conditions can create additional opportunities for combining deeper conditions as well.
- Technically, we could operate on "uninterrupted lists" rather than "adjacent conditions", and combine conditions with some elements between them. However, I'm not sure if this is desired.
- The logic would make it easier to extend the algorithm in the future, e.g. detect `else` opportunities like `not X` conditions, or combine+nest opportunities like `X and Y` conditions.

Is this what you were thinking of? Or at least close enough?

mgorny:
@beckermr, gentle ping.

beckermr:
Ah sorry!

Yes that is close but it will be clearer to me to write this as a recursive call like this that runs directly on the yaml itself

```python
def combined_conditions(node):
    "Breadth first recursive call to combined list conditions."""

    # recursion is breath first because we go through each element here 
    # before calling `combine_conditions` on any element in the node
    if isinstance(node, list):
        # 1. loop through list elements, gather the if conditions
        # 2. if elements share a compatible if condition
        # combine their if...then...else statements
        pass

    # now we descend down the tree
    if isinstance(node, dict):
        for k, v in node.items():
            if k != "if":  # don't combine if statements
                node[k] = combine_conditions(v)
    elif isinstance(node, list):
        for i in range(len(node)):
            node[i] = combine_conditions(node[i])

    return node
```

mgorny:
Ah. Actually, I initially tried working on the processed YAML, but `ruamel.yaml` kept changing the recipe style significantly. I don't think I have any code left of my initial attempts, though, and it's possible I didn't dig deep enough to find all the "off" switches.

beckermr:
The bot has a default ruamel.yaml style it uses. Let me send a link. 

https://github.com/regro/cf-scripts/blob/ff625d5b7e584045f8c3c9ce0cf2ccbba0d60828/conda_forge_tick/recipe_parser/_parser.py#L50

mgorny:
Thanks, this definitely helps! I am ashamed that I didn't grep that out while looking in other places where YAML is loaded/dumped. I will try rewriting it now.

mgorny:
@beckermr, seems to work quite nice. Open questions:

- Do we want to merge non-adjacent conditions? It produces correct results, but drops original grouping in recipes.
- Should we avoid combining conditions if there are comments between them? (usually indicating different dep groups)

beckermr:
I'm ok either way.

mgorny:
Well, if you're fine, then let's consider this ready for review and see how it works out. I can make further changes if people find the changes undesirable.

beckermr (changes requested):
One question. Otherwise LGTM!

tests/test_v1_yaml/version_pytorch_correct.yaml
```diff
@@ -0,0 +1,364 @@
...
+        - if: build_platform != target_platform
+          then:
+            - if: megabuild
+              then:
+                - python 3.12.*
+                - numpy  *
+            - if: not megabuild
+              then:
+                - python
+                - numpy
```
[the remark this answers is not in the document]
mgorny: That's literally what I want to do in a followup pull request. My idea is to do some easy heuristic to combine only "really safe" cases, like "X has no spaces, so `not X` is safe".

beckermr (approved):

mgorny:
Thanks!

Issue #3360, opened 2024-12-09
Migrator for `xorg-xextproto` & `xorg-xproto` to `xorg-xorgproto`


jakirkham:
The packages `xorg-xextproto` & `xorg-xproto` have been deprecated

Users are recommended to migrate to `xorg-xorgproto`

The linter now warns about this situation: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6688

Would it make sense to put together a migrator of some kind to manage this transition with the bot?

hmaarrfk:
There are a few other proto packages that can be caught by the migrator.
https://github.com/search?q=org%3Aconda-forge+xorg-+proto+language%3AYAML&type=code&l=YAML

once the core logic is in, then a cursory search through the first few pages might expand the ability to catch even more.

hmaarrfk:
* `xorg-xineramaproto`
* ` xorg-xproto`
* `xorg-xf86vidmodeproto`
* `xorg-xextproto`
* `xorg-glproto`
* `xorg-inputproto`
* `xorg-kbproto`

jakirkham:
Should we add linter hints for those as well?

hmaarrfk:
i mean we could, but its somewhat "low importance". Don't we already have a migrator for package renaming?

we could migrate all the proto packages that we find in: https://conda-forge.org/packages/

jakirkham:
Am not aware of an existing migrator. Hence the issue raised here to add one

At least currently see issues with clobbering when trying to install some packages from conda-forge. That would be nice to fix

hmaarrfk:
yeah i misremembered sorry. it was the whole piggyback minimigrator stuff and complications...

jakirkham:
So in general are we just want to migrate `xorg-*proto` to `xorg-xorgproto`?

hmaarrfk:
correct.

jakirkham:
Thanks Mark! 🙏

Updated the issue to reflect this

jakirkham:
> Should we add linter hints for those as well?

Extended the hints to these packages and anything else referencing `xorg-*proto` except `xorg-xorgproto`

xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6816

hmaarrfk:
see https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6822 for how to implement such a migrator.

jakirkham:
Think we would also need bot changes

[...]

PR #3851, opened 2025-03-10
Replace `XzLibLzmaDevelMigrator` with `MiniReplacement`

mgorny:
#### Description:

Introduce a `MiniReplacement` migrator that is a minimigrator based on `Replacement`, and replace the `XzLibLzmaDevelMigrator` with it. Extend to v1 recipes while at it.

#### Checklist:

Discussed in #3360

---------

@beckermr, could you do a quick initial review before I switch some more minimigrators over?


conda_forge_tick/migrators/replacement.py
```diff
@@ -147,3 +149,57 @@ def migrator_uid(self, attrs: "AttrsTypedDict") -> "MigrationUidTypedDict":
...
+        super().__init__()
+        self.old_pkg = old_pkg
+        self.new_pkg = new_pkg
+        self.pattern = re.compile(r"\s*-\s*(%s)(\s+|$)" % old_pkg)
+        self.packages = {old_pkg}
+
+    def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
+        requirements = attrs.get("requirements", {})
+        # TODO: do we want to limit to host requirements, like the split
+        # minimigrators did?
```
beckermr, line 182: We should make this an option and then turn it on as needed to match the current behavior. For ABI migrations, we only ever want to edit host.

beckermr (commented):
Looking good! One comment on the host-only thing which we need.

conda_forge_tick/migrators/replacement.py
```diff
@@ -147,3 +149,54 @@ def migrator_uid(self, attrs: "AttrsTypedDict") -> "MigrationUidTypedDict":
...
+    """
+
+    allowed_schema_versions = [0, 1]
+
+    def __init__(
+        self,
+        *,
+        old_pkg: "PackageName",
+        new_pkg: "PackageName",
+        requirement_types: tuple[str] = ("host",),
```
beckermr, line 172: Oh I am sorry I was mistaken. The old "xz" one did everything in the text file as long as it matched ` xz\n`. So I think actually what we want is none of the above and something much more general. We should inspect the other migrators to see what they look like too.

mgorny: Well, yes and no. It replaced everything but `filter()`-ed on `host` requirements. In that sense, the new minimigrator does the same thing. Ofc, I'm open to changing the behavior to something "saner".

beckermr: Right. My concern is us missing things like ignored run exports in `build/ignore_run_exports_from`, etc. The current migrator would catch that and we need to replicate that kind of behavior. I don't have an exhaustive list of where say `xz` might appear, which is why something more generic would be better.

beckermr: Just the word matching.

beckermr (approved):
LGTM! This Pr is great and thank you!

mgorny:
I do wonder if we really ought to test minimigrators combined with version migrators — we're seeing a lot of flakiness here, and I think version migrators per se have quite a lot coverage already.

beckermr:
The flakieness comes from the URL of the package we are testing with. We can move to a package with a less flaky URL or even pick a different test url in the recipe that doesn't correspond to the actual recipe at all.

h-vetinari:
@mgorny, this causes various workflows to break when they load the migrator file from the graph repo (current [location](https://github.com/regro/cf-graph-countyfair/blob/96f2cfc1c86ad7b921026fdfa6dd4fbffb40630f/migrators/a/6/0/0/1/xz_to_liblzma_devel.json)), including our main migrator bot and the status page update:
```
concurrent.futures.process._RemoteTraceback: 
"""
  File "/home/runner/micromamba/envs/cf-scripts/lib/python3.11/concurrent/futures/process.py", line 261, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/make_migrators.py", line 913, in _load
    return make_from_lazy_json_data(lzj.data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/migrators/core.py", line 158, in make_from_lazy_json_data
    kwargs["piggy_back_migrations"] = [
                                      ^
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/migrators/core.py", line 159, in <listcomp>
    make_from_lazy_json_data(mini_migrator)
  File "/home/runner/work/cf-scripts/cf-scripts/cf-scripts/conda_forge_tick/migrators/core.py", line 163, in make_from_lazy_json_data
    return cls(*data["args"], **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: MiniReplacement.__init__() missing 2 required keyword-only arguments: 'old_pkg' and 'new_pkg'
"""
```

mgorny:
I'm sorry about that.

@beckermr, could you revert it for now, please? And possibly advise on how to deal with that?

PR #3870, opened 2025-03-13
fix: find the last version tag as the max to fix a race condition

### LINKED ISSUE #3864: release tag calculation can fail & deadlock
I had been wondering why a pinning update didn't get reflected in the migrators. Much digging later, I found that the release workflow here is falling with `tag_name already_exists`:
```
github.GithubException.GithubException: Validation Failed: 422 {"message": "Validation Failed", "errors": [{"resource": "Release", "code": "already_exists", "field": "tag_name"}], "documentation_url": "https://docs.github.com/rest/releases/releases#create-a-release", "status": "422"}
```

That's because the tag logic uses the committer date
https://github.com/regro/cf-scripts/blob/3dd9ddd31cbd7d67d9808a526331ff2630db7f6e/.github/workflows/release.yml#L45-L46

and - for reasons I have not determined - there's an inversion in the timestamps for the most recent tags (not releases):

Image

A rough hypothesis for how this could have happened: the bot was already in the process of creating a tag for https://github.com/regro/cf-scripts/commit/ad7246b730fbf4cd1e49d6a823d6e09c7ffe0534 while @beckermr triggered a manual tag for https://github.com/regro/cf-scripts/commit/9cb55deefc8ad2e82f196410d791a650c7e0194b (which must have overtaken the ongoing bot process somehow), so when the bot looked up the next available tag, it used 2025.3.54, despite the fact that it was tagging a commit that had an earlier committerdate than the last tag.

Such cases should be caught, ideally before we create the tag, but at least once we create the next one.

beckermr:
#### Description:

Somehow the release workflow got out of order, as reported in #3864.

this PR fixes that by properly computing the maximum version tag instead of using the most recent tag.

#### Checklist:

Closes #3864

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,7 @@ jobs:
         id: version
         if: ${{ ! env.CI_SKIP }}
         run: |
-          echo "current version: "$(git tag --sort=committerdate | tail -1)
+          echo "recent version tags: "$(git tag --sort=committerdate | tail -4)
           NEXT=$(python autotick-bot/compute_next_version.py)
           echo "next version: ${NEXT}"
           echo "NEXT=${NEXT}" >> "$GITHUB_OUTPUT"
--- a/autotick-bot/compute_next_version.py
+++ b/autotick-bot/compute_next_version.py
@@ -38,7 +38,7 @@ else:
                 curr_version = _version
                 curr_version_line = line
     assert curr_version is not None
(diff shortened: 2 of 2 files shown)


beckermr:
pre-commit.ci autofix

beckermr:
cc @h-vetinari

beckermr:
Thanks for the code review and fixes @h-vetinari. I am surprised ruff did not catch that as a lint!

PR #3898, opened 2025-03-17
fix: move octave url to something more stable

beckermr:
#### Description:

As it says...

#### Checklist:


tests/test_minireplacement.py
```diff
@@ -46,7 +46,7 @@
         (
             "qtqtmain_octave_before_meta.yaml",
             ("qtqtmain_octave_after_meta.yaml", "qtqtmain_octave_xz_after_meta.yaml"),
-            "7.1.0",
+            "2025.3.25",
```
beckermr, line 49:
```suggestion
            "2025.3.75",
```

PR #3910, opened 2025-03-18
fix: again try again for nvidia

beckermr:
#### Description:

We're getting closer...

#### Checklist:

### DIFF
--- a/conda_forge_tick/make_migrators.py
+++ b/conda_forge_tick/make_migrators.py
@@ -854,15 +854,10 @@ def add_nvtools_migrator(
                     else:
                         src_list = [attrs["meta_yaml"]["source"]]
                     for src in src_list:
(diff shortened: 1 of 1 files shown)


beckermr:
pre-commit.ci autofix

carterbox:
Thanks, @beckermr. Is it easy to run make_migrators locally?

beckermr:
not so much and we need to add direct tests - the issue is that it takes 30 mins to run so we need to do something less intensive. we can do this in a followup PR.

jakirkham:
Thanks Matt! 🙏

Had spotted an issue and meant to reach out and see if we need to do more to fix it. Thank you for making the fixes

We would like to do more migrators in the near future. So if there are things we can do to make the overall process easier, please let us know 🙂

PR #3923, opened 2025-03-21
fix: remove side effects from feedstock loading

ytausch:
```
A function’s touch should leave no trace,
No hidden change, no state misplaced.
Let inputs stay as first they came,
Pure and true—untouched, the same.
```

#### Description:

This PR removes side effects from the `load_feedstock_local` and `populate_feedstock_attributes` functions.
Instead of modifying the `sub_graph` parameter, we create a new dictionary in the function body and return that one.

Rationale: As of now, `try_load_feedstock` **is flawed if run without containers**.

https://github.com/regro/cf-scripts/blob/160c08d1e846272182b5d27a4e640d1b2fce6137/conda_forge_tick/make_graph.py#L100-L115

This is because in no-container mode, the return value `data` of `load_feedstock` is identical to `attrs` which is cleared a few lines after. Hence, we always return a LazyJSON object which is empty besides the two values added by `_add_required_lazy_json_refs`.

In container mode, `load_feedstock` is already pure (free of side effects), hence the bug does not occur.

As far as I am aware, nothing depends on the side-effects (note that these for `load_feedstock` do not exist in container mode anyway!)

I also verified that this does not break any external code.

#### Checklist:

Cc @pavelzw

Blocks #3417

### DIFF
--- a/conda_forge_tick/feedstock_parser.py
+++ b/conda_forge_tick/feedstock_parser.py
@@ -234,16 +234,29 @@ def _clean_req_nones(reqs):
 
 def populate_feedstock_attributes(
     name: str,
-    sub_graph: typing.MutableMapping,
+    existing_node_attrs: typing.MutableMapping[str, typing.Any],
     meta_yaml: str | None = None,
     recipe_yaml: str | None = None,
     conda_forge_yaml: str | None = None,
     mark_not_archived: bool = False,
     feedstock_dir: str | Path | None = None,
-) -> typing.MutableMapping:
-    """Parse the various configuration information into something usable"""
+) -> dict[str, typing.Any]:
+    """
+    Parse the various configuration information into the node_attrs of a feedstock.
 
+    :param name: The name of the feedstock
+    :param existing_node_attrs: The existing node_attrs of the feedstock. Pass an empty dict if none.
+    :param meta_yaml: The meta.yaml file as a string
+    :param recipe_yaml: The recipe.yaml file as a string
+    :param conda_forge_yaml: The conda-forge.yaml file as a string
+    :param mark_not_archived: If True, forcibly mark the feedstock as not archived in the node attrs, even if it is archived.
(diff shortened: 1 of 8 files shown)


conda_forge_tick/feedstock_parser.py
```diff
@@ -538,6 +553,7 @@ def load_feedstock_local(
     sub_graph : MutableMapping
         The sub_graph, now updated with the feedstock metadata
     """
+    new_sub_graph = {key: value for key, value in sub_graph.items()}
```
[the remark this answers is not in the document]
ytausch: I didn't do this on purpose because I don't want to deep-copy nested LazyJSON objects that might already be referenced by the existing JSON file.

ytausch:
Ready for review!

tests/test_make_graph.py
```diff
@@ -0,0 +1,47 @@
...
+    with fake_lazy_json as loaded_lazy_json:
+        # FakeLazyJson is not an instance of LazyJson
+        # noinspection PyTypeChecker
+        data = try_load_feedstock(feedstock, loaded_lazy_json, mark_not_archived).data  # type: ignore
+
+    if mark_not_archived:
+        assert data["archived"] is False
+    else:
+        # apparently, this is how the function behaves
+        assert "archived" not in data
```
beckermr, line 29: So the feedstock node attrs are not stateless. The graph data will have an "archived" entry sometimes if the bot detects an archived feedstock on the fly. So the mark not archived keyword is there to reset existing feedstocks that are marked as archived. So this test function should set archived to true and ensure it gets reset.

beckermr (changes requested):

tests/test_make_graph.py
ytausch: Like this?

abf23f841342cf6b2158ce60005ce306f600bb9c

beckermr (approved):

PR #3932, opened 2025-03-24
feat: introduce global settings object

ytausch:
#### Description:

In #3417, I proposed to introduce a settings module to centralize the configuration of the bot. However, it had 2 flaws:
- It was not proposed in a separate PR.
- It loaded environment variables at import time, which complicates modifying them between CLI invocations.

This separate PR improves upon the old approach by using `pydantic-settings` to manage a global settings object. The values of the settings object are read from environment variables. As an advantage,

- environment variables can be modified, taking effect immediately on their next usage (because they aren't cached)
- the entire settings object can be overridden with another one, which allows type-safe settings modifications (e.g. for testing)

To keep the scope of this PR under control, not all configuration values are moved to the settings object; only the ones that are needed for the integration tests. I'd be happy though if this could become the new standard for adding new configuration values to stay flexible and have an overview of everything!

#### Checklist:

Cc @pavelzw

### DIFF
--- a/autotick-bot/install_bot_code.sh
+++ b/autotick-bot/install_bot_code.sh
@@ -1,4 +1,16 @@
 #!/bin/bash
+
+# Environment Variables:
+# - CF_FEEDSTOCK_OPS_CONTAINER_NAME: The name of the container image to use for the bot (optional, not used but left intact)
+# - CF_FEEDSTOCK_OPS_CONTAINER_TAG: The tag of the container image to use for the bot (optional).
+# - CF_TICK_GRAPH_GITHUB_BACKEND_REPO: The GitHub repository to clone cf-graph from. Default: regro/cf-graph-countyfair
+
+# Sets the following environment variables via GITHUB_ENV:
+# - CF_FEEDSTOCK_OPS_CONTAINER_NAME (see above)
+# - CF_FEEDSTOCK_OPS_CONTAINER_TAG (see above)
+
+set -euo pipefail
+
 git config --global user.name regro-cf-autotick-bot
 git config --global user.email 36490558+regro-cf-autotick-bot@users.noreply.github.com
 git config --global pull.rebase false
@@ -29,11 +41,14 @@ for arg in "$@"; do
   fi
 done
 if [[ "${clone_graph}" == "true" ]]; then
-  git clone --depth=5 https://github.com/regro/cf-graph-countyfair.git cf-graph
+  cf_graph_repo=${CF_TICK_GRAPH_GITHUB_BACKEND_REPO:-"regro/cf-graph-countyfair"}
+  cf_graph_remote="https://github.com/${cf_graph_repo}.git"
+  git clone --depth=5 "${cf_graph_remote}" cf-graph
 else
   echo "Skipping cloning of cf-graph"
 fi
 
+docker_name=${CF_FEEDSTOCK_OPS_CONTAINER_NAME:-"ghcr.io/regro/conda-forge-tick"}
 bot_tag=$(python -c "import conda_forge_tick; print(conda_forge_tick.__version__)")
 docker_tag=${CF_FEEDSTOCK_OPS_CONTAINER_TAG:-${bot_tag}}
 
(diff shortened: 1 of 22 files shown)


conda_forge_tick/git_utils.py
```diff
@@ -193,7 +193,8 @@ def _run_git_command(
         """
         git_command = ["git"] + cmd
 
-        logger.debug(f"Running git command: {git_command}")
+        if not suppress_all_output:
+            logger.debug("Running git command: %s", git_command)
```
ytausch, line 197: This is a bug I found while reading the code.

tests/conftest.py
```diff
@@ -108,3 +108,10 @@ def pytest_configure(config):
         "markers",
         "mongodb: mark tests that run with mongodb",
     )
+
+
+@pytest.fixture
+def temporary_environment():
+    old_env = os.environ.copy()
+    yield
+    os.environ = old_env
```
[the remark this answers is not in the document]
ytausch: good bot

autotick-bot/install_bot_code.sh
```diff
@@ -44,11 +59,12 @@ for arg in "$@"; do
...
 done
 if [[ "${pull_cont}" == "true" ]]; then
-  docker pull ghcr.io/regro/conda-forge-tick:${docker_tag}
+  docker pull "${docker_name}:${docker_tag}"
 fi
 
+# left intact if already set
 export CF_FEEDSTOCK_OPS_CONTAINER_TAG=${docker_tag}
-export CF_FEEDSTOCK_OPS_CONTAINER_NAME="ghcr.io/regro/conda-forge-tick"
+export CF_FEEDSTOCK_OPS_CONTAINER_NAME=${docker_name}
```
pavelzw, line 67: i would use quotes here

conda_forge_tick/settings.py
```diff
@@ -0,0 +1,106 @@
+from typing import Annotated
+
+from pydantic import Field
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+ENVIRONMENT_PREFIX = "CF_TICK_"
+"""
+All environment are expected to be prefixed with this.
```
pavelzw, line 8:
```suggestion
All environment variables are expected to be prefixed with this.
```

conda_forge_tick/settings.py
```diff
@@ -0,0 +1,106 @@
...
+"""
+
+ENV_CONDA_FORGE_ORG = ENVIRONMENT_PREFIX + "CONDA_FORGE_ORG"
```
pavelzw, line 11: not really pretty but i also can't think of another way. especially with the test this is fine imo

tests/test_upstream_versions.py
```diff
@@ -1356,11 +1357,26 @@ def test_update_upstream_versions_run_parallel_custom_sources(
...
+def version_update_frac_always():
+    new_settings = settings()
+
+    new_settings.frac_update_upstream_versions = True
+    use_settings(new_settings)
+
+    yield
+
+    # revert to default settings
+    use_settings(None)
```
pavelzw, line 1370: this means we cannot stack multiple settings override fixtures on top of each other. do you think we need this feature eventually?

pavelzw (approved):
thanks!

conda_forge_tick/settings.py
ytausch: yeah I stopped spending time on how to make this better :D

ytausch:
Ah, by the way to make reviewing easier: @beckermr you've already reviewed a large part of this in the integration tests PR. Via the commit history, you can see what has changed since then.

conda_forge_tick/settings.py
```diff
@@ -0,0 +1,106 @@
...
+def settings() -> BotSettings:
+    """
+    Get the current settings object.
+    """
+    if _USE_SETTINGS_OVERRIDE:
+        return _USE_SETTINGS_OVERRIDE.model_copy()  # prevent side-effects
+    return BotSettings()
+
+
+def use_settings(s: BotSettings | None) -> None:
```
beckermr, line 98: Should we make this a context manager?

beckermr (changes requested):
A few comments. Otherwise Lgtm! Looks fantastic and we should convert everything over as needed.

conda_forge_tick/deploy.py
```diff
@@ -148,22 +147,17 @@ def _deploy_batch(*, files_to_add, batch, n_added, max_per_batch=200):
...
-                                deploy_repo="regro/cf-graph-countyfair",
-                            ),
-                            "master",
-                        ],
-                        token=env.get("BOT_TOKEN", ""),
-                    )
-                if status != 0:
-                    print(">>>>>>>>>>>> git push failed", flush=True)
-
+            status = run_command_hiding_token(
```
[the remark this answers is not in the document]
ytausch: Oops, indeed! Thanks for pointing out! Fixed in 3ca076e2f65ff62af6008eedd4f6bfe6a9ad1347

conda_forge_tick/settings.py
ytausch: Sounds like a good idea! I've done this here: 507b9284f71396298f57d4ac7f5b9f02cf7f60f0

beckermr (approved):

PR #3935, opened 2025-03-24
Support combining trivial negations in v1 recipes

mgorny:
#### Description:

Extend the v1 recipe condition combiner to detect negations of trivial conditions (`x` vs. `not x`, provided that `x` does not use `and`, `or` or `if` operators).

#### Checklist:

### DIFF
--- a/conda_forge_tick/migrators/recipe_v1.py
+++ b/conda_forge_tick/migrators/recipe_v1.py
@@ -31,12 +31,20 @@ def is_negated_condition(a: str, b: str) -> bool:
     if not all(map(is_single_expression, (a, b))):
         return False
 
+    # X <-> not X
     a_not = a.startswith("not")
     b_not = b.startswith("not")
-    return (
+    if (
         a_not != b_not
         and a.removeprefix("not").lstrip() == b.removeprefix("not").lstrip()
-    )
+    ):
+        return True
+
+    # A == B <-> A != B
+    if a == b.replace("==", "!=") or a == b.replace("!=", "=="):
+        return True
+
+    return False
 
 
 def fold_branch(source: Any, dest: Any, branch: str, dest_branch: str) -> None:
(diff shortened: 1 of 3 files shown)


conda_forge_tick/migrators/recipe_v1.py
```diff
@@ -12,28 +12,60 @@
...
+    return None
 
 
-def fold_branch(source: Any, dest: Any, branch: str) -> None:
+def is_same_condition(a: str, b: str) -> bool:
+    return a == b
+
+
+def is_single_expression(condition: str) -> bool:
+    return not any(f" {x} " in condition for x in ("and", "or", "if"))
```
beckermr, line 26: Is syntax like `(...)and (...)` legal? This might fail there.

beckermr (changes requested):

conda_forge_tick/migrators/recipe_v1.py
mgorny: Yeah, I was wondering. Actually, overnight I got an idea that we could perhaps ask `jinja` to parse the expression — going to investigate that today.

mgorny: I have to leave now and finish later, but initial experiments with Jinja expression parser are very promising.

beckermr (changes requested):
The tests look good and reasonably complete. We need to look carefully at the jinja2 expression parser from a security perspective. If folks can use eval or other unsafe python syntax there it will be an issue.

mgorny:
We don't execute ("compile") the expressions, only lex and parse them into AST. I can double-check that no actual evaluation happens, if you want me to.

mgorny:
Checked a bunch of potential cases that I could think of: function calls are converted into `Call` nodes, item getters into `Getitem` nodes, attribute getters into `Getattr` nodes, filter uses into `Filter` nodes — and of course, variable refs into `Name` nodes. Nothing is evaluated, just parsed into an AST. More specific methods in Jinja (such as `compile_expression()`) afterwards put these ASTs into a template, and render it to evaluate the expression.

mgorny:
Thanks!

PR #3942, opened 2025-03-26
Support folding subconditions into superconditions in v1 recipes

mgorny:
#### Description:

Support combining subconditions into superconditions while folding `if` in v1 recipes, e.g.:

```yaml
- if: A
  then: FOO
- if: A and B
  then: BAR
```

will be combined into:

```yaml
- if: A
  then:
    - FOO
    - if: B
      then: BAR
```

#### Checklist:

### DIFF
--- a/conda_forge_tick/migrators/recipe_v1.py
+++ b/conda_forge_tick/migrators/recipe_v1.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import Any
 
 from jinja2 import Environment
-from jinja2.nodes import Compare, Node, Not
+from jinja2.nodes import And, Compare, Node, Not
 from jinja2.parser import Parser
 
 from conda_forge_tick.migrators.core import MiniMigrator
@@ -64,6 +64,10 @@ def is_negated_condition(a: Node, b: Node) -> bool:
     return False
 
 
+def is_sub_condition(sub_node: Node, super_node: Node) -> bool:
+    return isinstance(sub_node, And) and super_node in (sub_node.left, sub_node.right)
+
+
 def fold_branch(source: Any, dest: Any, branch: str, dest_branch: str) -> None:
     if branch not in source:
         return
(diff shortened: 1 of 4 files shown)


mgorny:
Actually, reverting the part combining adjacent `if A and B` with `if A and C` -- my code could have accidentally slurped parts of B and C.

beckermr:
> Actually, reverting the part combining adjacent `if A and B` with `if A and C` -- my code could have accidentally slurped parts of B and C.

Please add a test for this edge case if it is not already covered.

conda_forge_tick/migrators/recipe_v1.py
```diff
@@ -64,6 +64,22 @@ def is_negated_condition(a: Node, b: Node) -> bool:
     return False
 
 
+def is_sub_condition(sub_node: Node, super_node: Node) -> bool:
+    return isinstance(sub_node, And) and super_node in (sub_node.left, sub_node.right)
+
+
+def get_common_condition_part(sub_cond: str, super_cond: str) -> str | None:
```
beckermr, line 71: This function will fail if "(" appears in the conditions.

beckermr (changes requested):
To do this properly, we'll need to parse the expression formally I think.

conda_forge_tick/migrators/recipe_v1.py
mgorny: I know. However, it's a false negative rather than false positive — i.e. as the code is written now, it will not accidentally merge something that does not share a common part, but it may not merge more complex stuff that does have a common part. I'm a bit afraid that if I try to cover too many corner cases, I'll end up with false positives.

beckermr: Ok. Let's add some complex cases that could naively be merged to the test suite to ensure the code does not touch them.

mgorny:
I've added some unit tests for edge cases. Are there some other cases you'd like to see?

beckermr (approved):
Lgtm! Thank you!

mgorny:
Thanks. BTW after your comment, I'm actually thinking of trying to support the simplest case of extra parentheses.

PR #3960, opened 2025-03-31
fix: ensure we parse as v1 when searching for new versions via raw urls

beckermr:
#### Description:

This PR ensures we parse recipes properly as v0/v1 when getting new versions via raw urls.

#### Checklist:


beckermr:
I see the issue. I will make an issue to add a test, but am going to skip for now due to time.

PR #3966, opened 2025-04-01
Support handling extra parentheses when combining v1 recipes

mgorny:
#### Description:

Extend the v1 recipe logic to handle extra parentheses in super_cond, for example to combine "(a or b) and c" with "a or b".  To accomplish this, I have resorted to a regular expression, but I think in this case the gain is worth it -- especially that even without the new feature, it makes the code simpler.

#### Checklist:


beckermr:
pre-commit.ci autofix

beckermr (approved):

PR #3897, opened 2025-03-17
refactor: centralize migrator filtering to migrators

beckermr:
#### Description:

This PR centralizes all filtering of the graph done by the migrators to the migrators themselves. This logic was spread across multiple modules and that caused the code to be hard to maintain and unintuitive for folks adding new code. Now every migrator can compute the set of nodes it works on and the set of nodes it has completed without external code.

To do before merging:

- [x] test pass
- [x] check that migrators serialize and unserialize properly
- [x] check that migrator graphs are correct
- [x] check that status page still works

#### Checklist:

- [x] Pydantic model updated or no update needed

#### Cross-refs, links to issues, etc:

closes #3888

PR #3986, opened 2025-04-03
Revert "refactor: centralize migrator filtering to migrators"

beckermr:
Reverts regro/cf-scripts#3897


beckermr:
Reverting this since the memory usage is killing the bot.

PR #4206, opened 2025-05-16
style(ruff): enforce logging style

ytausch:
#### Description:

Thorughout the code, we use different logging styles: `logger.log(f"a: {a}")` (f-string), and `logger.log("a: %s", a)`. In a previous discussion with Jaime, we agreed upon using the latter style because it allows deferred evaluation of arguments.

This PR introduces the ruff rulesets `LOG` and `G` to enforce a consistent style. 

#### Checklist:

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -166,12 +166,14 @@ def _prepare_feedstock_repository(
         backend.fork(context.git_repo_owner, context.git_repo_name)
     except RepositoryNotFoundError:
         logger.warning(
-            f"Could not fork {context.git_repo_owner}/{context.git_repo_name}: Not Found"
+            "Could not fork %s/%s: Not Found",
+            context.git_repo_owner,
+            context.git_repo_name,
         )
 
         error_message = f"{context.feedstock_name}: Git repository not found."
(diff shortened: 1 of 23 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -791,7 +796,7 @@ def _run_migrator_on_feedstock_branch(
 
             if is_github_api_limit_reached():
                 logger.warning("GitHub API error", exc_info=e)
-                break_loop = True
+                break_loop
```
beckermr, line 799: This change looks wrong.

beckermr (changes requested):

conda_forge_tick/auto_tick.py
ytausch: Jup, I auto-generated most fixes with GitHub Copilot agent, then opened the draft PR, and now manually reviewed everything in 4120b8c. Should be fixed now!

beckermr (approved):

beckermr:
The test failure is for too many requests against github raw URLs. We can merge if that doesn't pass.

ytausch:
Seems like I don't have permission for that :)

PR #4236, opened 2025-05-23
feat: add workflow dispatch for integration tests and rename

beckermr:
#### Checklist:


beckermr:
I've renamed the workflow to keep it easier to track in the github UI. I also added a dispatch event so we can run them at will. Works like a charm! great work @ytausch!

ytausch (approved):

PR #4507, opened 2025-07-23
use `pytest-retry` for tests

h-vetinari:
Test failures like
```
=========================== short test summary info ============================
FAILED tests/test_update_deps.py::test_make_grayskull_recipe_github_url - requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/esheldon/ngmix/git/refs/tags
= 1 failed, 548 passed, 2 skipped, 5 xfailed, 1 xpassed, 55 warnings in 318.38s (0:05:18) =
```
are incredibly annoying, because they provide no useful information whatsoever, but they _do_ stop our migrator infrastructure in its tracks (e.g. picking up new versions of the pinning), c.f. #4466.

Before going as far as allowing failures in the relock (#4470), just retry the tests.

Manually relocking was actually surprisingly hard. After some digging around, I found the instructions
```
conda-lock -f environment.yml --lockfile conda-lock.yml
```
but that creates a huge diff with what's checked in. The missing magic sauce turned out to be a reformatting [function](https://github.com/conda-incubator/relock-conda/blob/5db5f2de7d6dcb8add7f434a84af37abf1012c2e/relock.py#L37-L52) called in the relock-conda action.


.gitattributes
```diff
@@ -1,2 +1,4 @@
+* text=auto
+
```
ytausch, line 2: Could you explain why you added this?

conda_forge_tick/contexts.py
ytausch, line 1: Why is a diff generated for this file? Are the line endings different maybe?

ytausch (commented):
[pytest-retry](https://github.com/str0zzapreti/pytest-retry) seems to be far less popular than flaky. Why don't we stick with flaky but increase the number of runs / the delay between flaky test runs?

h-vetinari:
> [pytest-retry](https://github.com/str0zzapreti/pytest-retry) seems to be far less popular than flaky. Why don't we stick with flaky but increase the number of runs / the delay between flaky test runs?

The main reason is that flaky needs annotations and handmade delay functions, whereas all that can simply be handled by `--retries X --retry-delay Y`.

I don't want to manually annotate all tests as we discover each possible failure, but rather just retry failures by default.

beckermr:
@h-vetinari it appears the merge failed which means something failed despite the retries. Do you want to increase the number?

h-vetinari:
The last few [test runs](https://github.com/regro/cf-scripts/actions/workflows/tests.yml) all passed, but that could just be dumb luck. I tried to look for signs of retried tests, but couldn't find any. 🤷

beckermr:
Yeah I noticed the retry output is missing. We'll see!

h-vetinari:
We got a retry report ([CI run](https://github.com/regro/cf-scripts/actions/runs/16485253655/job/46608632649)), it looks like one test failed despite all retries
```
[deleted same error for prior attempts]

 	test_container_tasks_provide_source_code_local failed after 6 attempts!
	  File "/home/runner/work/cf-scripts/cf-scripts/conda_forge_tick/provide_source_code.py", line 180, in provide_source_code_local
	    yield provide(md)
	          ^^^^^^^^^^^
	RuntimeError: SHA256 mismatch for 0.80.tar.gz: obtained '951cf0babc2ac800372f21d98d563b060f84209206db5810c2f1c4e6b418c7a7' != expected 'f8670395d8b9c0ed02f3feeed839b1328185fd6f22a07e09349e7cecb50ac421'
	
	During handling of the above exception, another exception occurred:
	
	  File "/home/runner/micromamba/envs/cf-scripts/lib/python3.11/site-packages/_pytest/runner.py", line 342, in from_call
	    result: Optional[TResult] = func()
	                                ^^^^^^
	RuntimeError: conda build src exception: SHA256 mismatch for 0.80.tar.gz: obtained '951cf0babc2ac800372f21d98d563b060f84209206db5810c2f1c4e6b418c7a7' != expected 'f8670395d8b9c0ed02f3feeed839b1328185fd6f22a07e09349e7cecb50ac421'

=========================== end of test retry report ===========================
```

But at least that indicates that the mechanism is working. :)

h-vetinari:
Just noticed a [failure](https://github.com/regro/cf-scripts/actions/runs/16544366991/job/46789935265) on main, though it happened during _collection_ (i.e. `error` rather than `fail`), which don't get retried it seems:

```
==================================== ERRORS ====================================
___________________ ERROR collecting tests/test_staticlib.py ___________________
tests/test_staticlib.py:81: in <module>
    get_latest_static_lib("llvm 15.*", "osx-64")
conda_forge_tick/migrators/staticlib.py:102: in get_latest_static_lib
    rd = _read_repodata(platform_arch)
conda_forge_tick/migrators/staticlib.py:78: in _read_repodata
    raise RuntimeError(f"Download of repodata for {platform_arch} failed!")
E   RuntimeError: Download of repodata for osx-64 failed!
```

The code in question already does manual retries with exponential backoff, so I guess we were just very unlucky there (and the repodata unavailable for >100sec)
https://github.com/regro/cf-scripts/blob/8237c3d5ae8c6e47c52565de1b872ab6d525219a/conda_forge_tick/migrators/staticlib.py#L62-L80

Issue #4517, opened 2025-07-24
why does a changed hash on a random feedstock fail the test suite here?


h-vetinari:
I was looking into the failures observed despite #4507, which look like
```
 	test_container_tasks_provide_source_code_local failed after 6 attempts!
	Traceback (most recent call last):
	  File "/home/runner/work/cf-scripts/cf-scripts/conda_forge_tick/provide_source_code.py", line 180, in provide_source_code_local
	    yield provide(md)
	          ^^^^^^^^^^^
	RuntimeError: SHA256 mismatch for 0.80.tar.gz: obtained '951cf0babc2ac800372f21d98d563b060f84209206db5810c2f1c4e6b418c7a7' != expected 'f8670395d8b9c0ed02f3feeed839b1328185fd6f22a07e09349e7cecb50ac421'
```
and are coming from 
https://github.com/regro/cf-scripts/blob/b0f22814b412b3a250d819a9294f12400981238c/tests/test_container_tasks.py#L786

Indeed, that recipe currently [has](https://github.com/conda-forge/mattspy-feedstock/blob/4ab3a8ce9287bf77524c86e135e0732f66a0ae09/recipe/meta.yaml#L8)
```yaml
source:
  url: https://github.com/beckermr/mattspy/archive/refs/tags/{{ version }}.tar.gz
  sha256: f8670395d8b9c0ed02f3feeed839b1328185fd6f22a07e09349e7cecb50ac421
```
but the hash of that tag has changed
```
>curl -L https://github.com/beckermr/mattspy/archive/refs/tags/0.80.tar.gz | sha256sum
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 20.7M    0 20.7M    0     0  2305k      0 --:--:--  0:00:09 --:--:-- 2638k
951cf0babc2ac800372f21d98d563b060f84209206db5810c2f1c4e6b418c7a7 *-
```

I can open a rerender there to update the hash, but my first question is why this would fail CI here in the first place? 🤔 

CC @beckermr

h-vetinari:
> We can switch to a more stable url.

But that would have to happen on https://github.com/conda-forge/mattspy-feedstock, right?

beckermr:
No. Change the test to use smithy's recipe.

h-vetinari:
[...]

PR #4519, opened 2025-07-25
switch external feedstock in test_container_tasks_provide_source_code_local

h-vetinari:
Fix a failing test due to #4517


h-vetinari:
OK, even if this is not a perfect fix for #4517 (I didn't understand what you wanted exactly @beckermr), I'm merging this to unblock the migrators.

PR #4524, opened 2025-07-25
fix: use heuristic to prefer correct versions for feedstocks

beckermr:
#### Description:

For fenics-basix, the ordering of the outputs from rattler-build was causing us to grab the wrong version. It is really hard for the bot know which output has the preferred version. This PR adds some heuristics based on feedstock names and how context variables are set. That should help.

#### Checklist:

### DIFF
--- a/conda_forge_tick/utils.py
+++ b/conda_forge_tick/utils.py
@@ -536,6 +536,39 @@ def _parse_recipes(
     RecipeTypedDict
         A dict conforming to conda-build's rendered output
     """
+    # some heuristics to find the "top-level" name and version
+    # we use these in the following order
+    # 1. Does the output name match the "name" field in the context block
+    #    if present?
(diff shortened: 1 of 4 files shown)


conda_forge_tick/utils.py
```diff
@@ -536,6 +536,38 @@ def _parse_recipes(
...
     """
+    # some heuristics to find the "top-level" name and version
+    # we use these in the following order
+    # 1. Does the output name match the "name" field in the context block
+    #    if present?
+    # 2. Does the output version match the "version" field in the
+    #    context block if present?
+    # 3. Does the output name match the feedstock-name in the extra
+    #    field if present?
+    if len(validated_recipes) > 0:
```
minrk, line 547: Should this be only when there's more than one?

```suggestion
    if len(validated_recipes) > 1:
```

PR #4587, opened 2025-08-15
REF: Scrape NVIDIA domain for redist package versions

carterbox:
#### Checklist:

### DIFF
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -9,13 +9,15 @@ import typing
 import urllib.parse
 from pathlib import Path
 from typing import Iterator, List, Literal, Optional
+import json
 
 import feedparser
 import orjson
 import requests
+import urllib.request
 import yaml
 from conda.models.version import VersionOrder
-from packaging.version import InvalidVersion
+from packaging.version import InvalidVersion, Version
 from packaging.version import parse as parse_version
 
 from conda_forge_tick.utils import parse_meta_yaml, parse_recipe_yaml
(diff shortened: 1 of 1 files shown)


conda_forge_tick/update_sources.py
```diff
@@ -651,80 +653,125 @@ class NVIDIA(AbstractSource):
...
+        logger.debug("Searching %s for redistrib_X.Y.Z.json", actual_url)
+        response = requests.get(actual_url)
+        html_content = response.text
+        # Search for links to redistrib_*.json in the response and strip the versions from there
+        # re doesn't support repeating patterns, so we assume there are three version numbers
+        redistrib_pattern = re.compile(
+            pattern=r"redistrib_[0-9]+\.[0-9]+\.[0-9]+\.json<"
+        )
+        result = re.findall(redistrib_pattern, html_content)
+        stripped_results = [x[10:-6] for x in result]
```
leofang, line 760: Can we 
- strip filename extension `.json`
- split the string based on `_`
- get the version string from the second element 

?

conda_forge_tick/update_sources.py
```diff
@@ -651,80 +653,125 @@ class NVIDIA(AbstractSource):
...
+        stripped_results.sort(key=Version, reverse=True)
```
leofang, line 761: 👍

conda_forge_tick/update_sources.py
```diff
@@ -651,80 +653,125 @@ class NVIDIA(AbstractSource):
...
+            ]
+        elif "feedstock-name" in meta_yaml["meta_yaml"]["extra"]:
+            logger.debug("Found extra/feedstock-name.")
+            nvidia_redist_json_name = meta_yaml["meta_yaml"]["extra"]["feedstock-name"]
+        else:
+            logger.debug("Found the feedstock's name.")
+            nvidia_redist_json_name = meta_yaml["feedstock_name"]
+        logger.debug(
+            "The compute-subdir for NVIDIA URL is %s.", nvidia_redist_json_name
+        )
```
leofang, line 725: This assumes the `extra` section may contain the needed fields. I am not sure if this is true today. I assume it is.

conda_forge_tick/update_sources.py
```diff
@@ -651,80 +653,125 @@ class NVIDIA(AbstractSource):
...
-        return url  # = next version, same as in BaseRawURL
+        """Get latest version of a package by scraping nvidia.com JSONs.
+
+        url should be an address of the following form: "https://developer.download.
+        nvidia.com/compute/{nvidia_compute_subdir}/redist#{nvidia_redist_json_name}".
+        where we expect 'https://developer.download.nvidia.com/compute/
+        {{ nvidia_compute_subdir }}/redist/' to be a list of links which includes
+        multiple 'redistrib_X.Y.Z.json'. The latest of which will be the nominal latest
+        version. AND the string after # is the slug used to access the part of the JSON
+        blob which is relevant to the package.
```
leofang, line 740: I don't understand what this means.

conda_forge_tick/update_sources.py
```diff
@@ -651,80 +653,125 @@ class NVIDIA(AbstractSource):
...
+        """
+        actual_url, slug = url.split("#")
+        logger.debug("Searching %s for redistrib_X.Y.Z.json", actual_url)
+        response = requests.get(actual_url)
+        html_content = response.text
+        # Search for links to redistrib_*.json in the response and strip the versions from there
+        # re doesn't support repeating patterns, so we assume there are three version numbers
+        redistrib_pattern = re.compile(
+            pattern=r"redistrib_[0-9]+\.[0-9]+\.[0-9]+\.json<"
+        )
```
leofang, line 758: Note to self: This was already the assumption of the old impl

leofang (commented):
Looks nice! Thanks!

conda_forge_tick/update_sources.py
carterbox: The purpose of this logic is to allow controlling the redist url from inside the feedstock instead of maintaining that information here in the tick-bot implementation. Right now, none of the feedstocks will have `extra/redist-json-name` or `extra/compute-subir`; they are not expected to have them. These extra are only needed if the feedstock name is not correlated with the URL of the NVIDIA redist; e.g. when the URL of the NVIDIA redist is irregular.

conda_forge_tick/update_sources.py
carterbox: Please look at line 676 of this MR. Repeated here for convenience:

```json
{
    "release_date": "2024-12-03",
    "release_label": "0.8.1",  # this is the same version used in the filename
    "release_product": "nvjpeg2000",
    "libnvjpeg_2k": {  # this is the key which corresponds to redist-json-name
        "name": "NVIDIA nvJPEG 2000",
        "license": "nvJPEG 2K",
        "license_path": "libnvjpeg_2k/LICENSE.txt",
        "version": "0.8.1.40",
        ...
}
```

Note that we need two pieces of information to get the latest version of a package (1) The URL of a JSON blob AND (2) the key (slug) inside the JSON blob which contains the latest version of the package. This is because *one* JSON blob may contain formation about *multiple* packages e.g. sometimes `nvidia_compute_subdir != nvidia_redist_json_name`.

conda_forge_tick/update_sources.py
carterbox:
```suggestion
        stripped_results = [x.removesuffix(".json").split("_")[1] for x in result]
```

leofang:
pre-commit.ci fix

leofang:
oh there's a merge conflict, so pre-commit failed

carterbox:
@leofang I have resolved the merge conflict and run pre-commit locally.

carterbox:
Thanks @beckermr! I agree with your assessment that this updater is awkwardly forced into the API of the existing one. However I think the new recipe keys (or similar) would still be necessary however because conda-forge feedstock names do not 1:1 correspond to NVIDIA product names.

Are you referring to the existing VersionFromFeed class?

beckermr:
Yes that class could be used if things were more unified and names matched etc.

leofang:
> It may be useful on the Nvidia end to understand if a simpler, more generic solution like an XML feed of the latest released versions located at canonical URL might be easier to maintain all around.

@kmittman is working on something on our end (NV) that might be able to address this in the future.

kmittman:
Sorry I was OOTO on vacation. Yes, we are planning to provide RSS/Atom/JSON feeds as part of the release process, in the near future.

The redist JSON manifests are based on the concept of each product release label being unique and monotonically increasing. Each product may contain 1 or more components; each may have a its own version scheme and build numbers (not necessarily unique between releases). Components may have multiple variants and may contain sub-packages (e.g. dev/devel and runtime)

To put it another way:
- define `PREFIX=https://developer.download.nvidia.com/compute`
- `${PREFIX}/${product}/redist/redistrib_${label}.json` refers to 
- `${PREFIX}/${product}/redist/${component}/${platform}/${filename}` 
where filename is either
- `${component}-${os}-${arch}-${version}-archive.tar.xz` (or `.zip`)
- `${component}-${os}-${arch}-${version}_${variant}-archive.tar.xz` (or `.zip`)

I'm not sure if the issue is mapping these `${product}` paths to conda-forge feedstocks (or vice-versa) ?

Or perhaps its more nuanced, like determining the next release label? It's liable to break in the future, currently for [shell-parse.sh](https://github.com/NVIDIA/build-system-archive-import-examples/blob/main/shell-parse.sh) `--latest` scrape the index.html page, like such

```
$ curl -s https://developer.download.nvidia.com/compute/cudnn/redist/ | sed -e 's|>|\n|g' -e 's|<|\n|g' | grep "^redistrib_" | sort -Vr | head -n1
```

PR #4664, opened 2025-08-30
Add support for ignoring odd minor versions

### LINKED ISSUE #1152: autotick-bot: add support for ignoring odd minor version numbers
The GNOME project consistently uses odd minor version numbers in their releases to indicate 'development' packages that should only be used for testing. Even minor version numbers represent 'stable' releases.

Currently the bot opens PRs for all releases, development and stable, which has led to many development releases being merged in: though this hasn't led to any serious problems so far, it should be avoided, as there is no API stability guarantee between such releases.

It would be great if an option could be added somewhere in the recipe or feedstock config to do this.

danyeaw:
#### Description:

 Adds support for ignoring odd minor version updates for odd/even version schemes.

I think we want to ignore them both for checking and storing upstream version changes and for auto tick of the version, but please let me know if there is a different architecture we want.

#### Checklist:

Closes #1152.

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -64,11 +64,11 @@ from conda_forge_tick.utils import (
     fold_log_lines,
     frozen_to_json_friendly,
     get_bot_run_url,
-    get_keys_default,
     get_migrator_name,
     load_existing_graph,
     sanitize_string,
 )
+from conda_forge_tick.version_filters import filter_version
 
 from .migrators_types import MigrationUidTypedDict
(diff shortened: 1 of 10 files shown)


beckermr (changes requested):
This isn't due to your work, but the code that checks various ignored or even/odd versions is duplicated. Can we go ahead and combine the logic to one function in one spot?

beckermr (approved):

beckermr:
Thank you for the great PR!

danyeaw:
Thanks @beckermr for the fixups!

PR #4752, opened 2025-09-19
chore: enforce mypy for all files starting with a letter from a-e

ytausch:
#### Description:

This will ultimately be the way we enable mypy for all files in this repo: Step by step. This PR enables mypy for all top-level modules and tests starting with a letter from a to e.

#### Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,3 +39,8 @@ repos:
       - id: ruff-check
         args: [ --fix ]
       - id: ruff-format
+
+  - repo: https://github.com/conda-incubator/pre-commit-mirrors-mypy
+    rev: 1.18.1
+    hooks:
+      - id: mypy-conda
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,3 +44,9 @@ repos:
     rev: 1.18.1
     hooks:
       - id: mypy-conda
+        exclude: |
+          (?x)^(
+            ^tests\/.*|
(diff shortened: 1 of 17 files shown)


.pre-commit-config.yaml
```diff
@@ -39,3 +39,13 @@ repos:
       - id: ruff-check
         args: [ --fix ]
       - id: ruff-format
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
```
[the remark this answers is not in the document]
beckermr: Using the official mirror seems fine. I don't know why the conda-based on existed in the first place, not that it didn't have a good reason.

conda_forge_tick/deploy.py
```diff
@@ -93,10 +94,11 @@ def _pull_changes(batch):
     return n_added
 
 
-def _deploy_batch(*, files_to_add, batch, n_added, max_per_batch=200):
+def _deploy_batch(*, files_to_add: Iterable[str], batch, n_added, max_per_batch=200):
```
beckermr, line 97:
```suggestion
def _deploy_batch(*, files_to_add: set[str] | list[str], batch, n_added, max_per_batch=200):
```

I think this will let us remove the cast below which would convert a set to a list when not needed.

beckermr: Python sets have a pop method: https://docs.python.org/3/library/stdtypes.html#frozenset.pop

beckermr (approved):

PR #4755, opened 2025-09-19
pref: push versions as we go so we can update everything in one go

beckermr:
#### Description:

This PR will ensure we never miss looking for a version for a feedstock while also ensuring the version updates get queued as we go.

#### Checklist:


conda_forge_tick/update_upstream_versions.py
```diff
@@ -459,9 +459,13 @@ def _update_upstream_versions_process_pool(
             # writing out file
             lazyjson = LazyJson(f"versions/{node}.json")
             with lazyjson as version_attrs:
+                changed = version_attrs.data == version_data
```
beckermr, line 462:
```suggestion
                changed = version_attrs.data != version_data
```

Issue #4749, opened 2025-09-18
Use exponential back-off for bot retries


h-vetinari:
I continue to strain against the way our bot infrastructure handles retries for large migrations. In the case of the R 4.5 migration, we're currently stuck on r-curl, even though that feedstock was migrated [manually](https://github.com/conda-forge/r-curl-feedstock/pull/50) a week ago.

<details>

<img width="1627" height="917" alt="Image" src="https://github.com/user-attachments/assets/8a836fe7-6b5a-4028-b327-101c450ece82" />

</details>

In this case, not even #4652 would have helped, because the bot then failed on a missing dependency that was due to a random CI flake (which I restarted as soon as I saw the failure). The last [attempt](https://github.com/regro/cf-scripts/actions/runs/17653960149) of the bot is a week ago, and the bot hasn't retried since then.

What seems to have happened is that the bot exhausted its retries, and now we're in neverneverland territory of "maybe one day we'll retry the feedstock again". For large migrations, I find it an unacceptable outcome if the bot cannot process a feedstock (already fixed manually!) within 1-2 days. This is not a personal criticism towards anyone, just that we as engineers and maintainers need to be able to do better.

The bot already uses `max_solver_attempts:`, but using a simple +1 scheme AFAICT
https://github.com/regro/cf-scripts/blob/a084180505ecc5c404c2edcfcd4d5983ae92383c/conda_forge_tick/auto_tick.py#L740-L746

What we could do is that we take the timestamp of the last bot attempt (which should be part of the node metadata AFAIU), e.g. in
https://github.com/regro/cf-scripts/blob/a084180505ecc5c404c2edcfcd4d5983ae92383c/conda_forge_tick/auto_tick.py#L136-L142

and then apply an exponential back-off before "spending" a retry (rather than burning our retry budget as quickly as possible):
```python
BASELINE = 4h  # or 8h, or 1 day, or whatever
INTERVALS = [BASELINE * 2 ** x for x in range(10)]
if time.time() > last_bot_attempt_ts + INTERVALS[retries_so_far]:
    # do the bot attempt, thus incrementing retry counter, else skip
```
That way we could also use a higher default retry count, because retries would happen much more rarely per unit of time.

h-vetinari:
What are the steps necessary for adding a new field to the graph nodes? Presumably additive changes are not too disruptive?

h-vetinari:
[...]

PR #4774, opened 2025-09-20
feat: record timestamp of last PR attempt and use exp backoff

beckermr:
#### Description:

Per #4749, this PR adds the timestamp of the last solver attempt to the bot metadata and uses exp backoff to retry PRs. I have updated the pydantic model and added hooks into all of the schema handling. Some of the schema handling sections are likely not needed since this data was never recorded in the old bot model. However, it is easier to treat things uniformly. 

We should actually remove the schema migration handling code at some point and bake in the new schemas, but not today.

#### Checklist:


beckermr:
I can annotate the PR diff with comments on what each bit does it that would be helpful @h-vetinari.

beckermr (commented):
I annoted the diff @h-vetinari. Feel free to comment with questions!

beckermr:
One minor bug here is that leaf nodes will almost always come last in the random sort. We can add an increment to the number of descendants to force them to come before failed nodes not being retried.

PR #4781, opened 2025-09-21
feat: remove max_solver_attempts and refactor to parameterize migrator shares directly

beckermr:
#### Description:

After we moved to exponential backoff, we no longer need this parameter for the migrators. It was used to determine the migrator's share of time. So now we parameterize this directly.

#### Checklist:

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -733,7 +733,7 @@ def run(
     return migrate_return_value, pr_lazy_json
 
 
-def _compute_time_per_migrator(migrators):
+def _compute_time_per_migrator(migrators, max_attempts_for_share=3):
     # we weight each migrator by the number of available nodes to migrate with a
     # a penalty for attempts and accounting for the pr_limit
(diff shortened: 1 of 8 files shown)


conda_forge_tick/auto_tick.py
```diff
@@ -733,7 +733,7 @@ def run(
     return migrate_return_value, pr_lazy_json
 
 
-def _compute_time_per_migrator(migrators):
+def _compute_time_per_migrator(migrators, max_attempts_for_share=3):
```
h-vetinari, line 736: Maybe I'm not understanding the mechanics entirely, but this just seems to make `3` to be the unoverrideable default; the `_for_share` suffix is confusing to me, because we're still comparing this to a number that can increase globally?
https://github.com/regro/cf-scripts/blob/8da10e3528b8f340fe8811c5560208ceeae0e211/conda_forge_tick/auto_tick.py#L756-L761

If we're hardcoding a value for the number of attempts here, I think it should be something like 10, not 3.

Issue #4805, opened 2025-09-23
Version updates for git sources?


jaimergp:
Comes from [#general > XDR / TI-RPC multi-platform libraries on conda-forge? @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/XDR.20.2F.20TI-RPC.20multi-platform.20libraries.20on.20conda-forge.3F/near/540972278) and https://github.com/conda-forge/libtirpc-feedstock/pull/5#discussion_r2371746497.

I wonder if we can add a provider for pure git sources that uses `git ls-remote --tags "REPOSITORY_URL"` to list the remote tags and see if there's any updates.

PR #4811, opened 2025-09-24
Implement version updates from git repo tags

jaimergp:
#### Description:

Adds support for version updates in this type of recipe:

```yaml
# meta.yaml
source:
  git_url: git://linux-nfs.org/~steved/libtirpc
  git_rev: libtirpc-{{ "-".join(version.split(".")) }}

# recipe.yaml
source:
  git: git://linux-nfs.org/~steved/libtirpc
  tag: libtirpc-${{ version | split(".") | join("-") }}
```

#### Checklist:

- Closes https://github.com/regro/cf-scripts/issues/4805


conda_forge_tick/update_sources.py
```diff
@@ -590,6 +590,35 @@ def get_url(self, meta_yaml) -> Optional[str]:
         return super().get_url(meta_yaml)
 
 
+class GitTags(AbstractSource):
+    name = "GitTags"
+
+    def get_url(self, meta_yaml):
+        return meta_yaml.get("git_url")
```
beckermr, line 597: This key will not be available at the top level in the node attributes. The name of the variable is confusing here (not your fault) and the code paths are very confusing for parsing (not your fault). 

The url and the hash are added as special cases [here](https://github.com/regro/cf-scripts/blob/main/conda_forge_tick/feedstock_parser.py#L530).

Instead, you'll need to look directly in the meta_yaml key at the source section.

tests/test_upstream_versions.py
```diff
@@ -1943,6 +1945,46 @@ def test_latest_version_cratesio(tmpdir):
...
+        inp = fd.read()
+
+    pmy = LazyJson(os.path.join(str(tmpdir), "cf-scripts-test.json"))
+    with pmy as _pmy:
+        yml = (
+            parse_recipe_yaml(inp, use_container=False)
+            if yaml_path == YAML_PATH_V1
+            else parse_meta_yaml(inp, use_container=False)
+        )
+        _pmy.update(yml["source"])
```
beckermr, line 1972: The tests is passing when it should not because this line is adding more keys to the node attributes than should be there. We should only add the url and the hash keys here.

conda_forge_tick/update_sources.py
jaimergp: Added in [c4266d5](https://github.com/regro/cf-scripts/pull/4811/commits/c4266d569a563a355a3e6d6e4637609d6e8a221e), thanks!

tests/test_upstream_versions.py
beckermr: I am not sure. Let's see what happens!

conda_forge_tick/update_sources.py
```diff
@@ -590,6 +590,39 @@ def get_url(self, meta_yaml) -> Optional[str]:
         return super().get_url(meta_yaml)
 
 
+class GitTags(AbstractSource):
+    name = "GitTags"
+
+    def get_url(self, meta_yaml):
+        return (
+            (meta_yaml.get("meta_yaml") or meta_yaml.get("recipe_yaml", {}))
```
beckermr, line 598: We confusingly always call it "meta_yaml". https://github.com/regro/cf-scripts/blob/main/conda_forge_tick/feedstock_parser.py#L462

conda_forge_tick/update_sources.py
```diff
@@ -590,6 +590,39 @@ def get_url(self, meta_yaml) -> Optional[str]:
...
+                (i for i, c in enumerate(tag) if c.isdigit()), len("refs/tags/")
+            )
+            version = tag[first_digit:].replace("-", ".")
+            try:
+                parse_version(version)
+            except InvalidVersion:
+                continue
+            else:
+                versions.append(version)
+        return max(versions, key=VersionOrder)
```
beckermr, line 623: What happens if we don't find any versions here?

conda_forge_tick/update_sources.py
beckermr: The test code is setting these keys incorrectly for v1 I think.

beckermr (approved):

beckermr:
Thanks for the pr! I should make some issues around the parts of this that were hard. What I saw were


- the naming of meta_yaml in the variable name of the methods
- the naming of meta_yaml in the node attributes
- unclear expectations around error handling
- testing code with unclear patterns

Anything else to add here?

jaimergp:
I missed having type hints in the signatures (that'd clarify what `meta_yaml` actually is), and docstrings in the relevant base classes and their methods.

I was also unsure how much logging we should have for debugging.

Issue #4913, opened 2025-10-14
Tighten the `GitHub` source's version check based on prefix?


leofang:
These two lines:
https://github.com/regro/cf-scripts/blob/e82552ebf06e3348b048b4c706c13f7fec74f604/conda_forge_tick/update_sources.py#L715-L716
are likely the root cause of https://github.com/conda-forge/cccl-python-feedstock/issues/9. It failed because in the NVIDIA/cccl repo there are two different naming conventions for git tags
- `vX.Y.Z`: This is for C++ releases
- `python-X.Y.Z`: This is for Python releases

but the current prefix logic would match both conventions, causing the bot to misidentify `3.0.3` as the next release of `0.3.0` (the expected version is `0.3.1`).

I propose to fix the second line to 
```python
self.ver_prefix_remove = [self.version_prefix]
```  
so that we only use the found prefix to traverse the RSS feed (`release.atom`). Please let me know if this makes sense or if there is a better approach, thanks! 🙂

beckermr:
People change tag conventions on the same repo and so this would break those cases. Is there another way? Could we add a set of prefixes to ignore that is configured per feedstock in the conda-forge.yml?

PR #4918, opened 2025-10-16
feat: add field allowed_tag_globs to version_updates schema

oleksandr-pavlyk:
The field, if specified, gives regular expression for tags which are considered. Tags not matching the regular expression are ignored.

This feature is needed for cccl-python-feedstock, where the regexp would be set to 'python-*'. See gh-4913 for context and discussion.

#### Checklist:

gh-4913

### DIFF
--- a/conda_forge_tick/cf_tick_schema.json
+++ b/conda_forge_tick/cf_tick_schema.json
@@ -113,7 +113,7 @@
             }
           ],
           "default": null,
-          "description": "For version sources that parse repo/vcs tags (e.g., `gittags`, `github`, `githubreleases`), the list of glob patterns that are allowed. This field can be used to filter the set of tags to only those relevant for the feedstock.",
+          "description": "For version sources that parse repo/vcs tags (e.g., `gittags`, `github`, `githubreleases`), the list of glob patterns that define which tags are allowed. This field can be used to filter the set of tags to only those relevant for the feedstock.",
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
A few comments:

- Let's change "whitelist" to "allowlist".
- Let's do the check inside the `is_ignored_version` and similar functions. This will ensure it gets applied uniformly to all version sources.

beckermr:
The field is a regex that is tested against the tags in the feed. I think that is fine, though this does open the bot up to regex abuse (i.e. DOS from expensive to compute regex functions). 

We should switch to glob-style matching. I think this will cover most cases and is less subject to abuse.

beckermr:
OK. We do need to ensure that all version sources and other times versions are ignored use the same logic everywhere. 

So let's do the following. In `is_version_ignored`, let's branch on whether or not the first character of the version is a digit. If it is, we apply the function as we do now. If it is not (e.g., someone passes `release-5.6.0`), then we check against the allowed globs for prefixes. Finally, we can add an additional check with `is_version_ignored` in the implementation of each version source as needed. I can push these changes to this PR if you'd like.

beckermr:
OK @oleksandr-pavlyk - I think this PR will do the trick. I centralized the logic (in a slightly different way than I had suggested before) and made sure all version sources that parse repo tags used the logic.

PR #4979, opened 2025-10-31
chore: Enable mypy for file prefix "m"

ytausch:


conda_forge_tick/migrators/noarch_python_min.py
```diff
@@ -469,6 +468,8 @@ def migrate(self, recipe_dir, attrs, **kwargs):
         for line in attrs.get("raw_meta_yaml", "").splitlines():
             if "{{ python_min }}" in line:
                 muid = super().migrate(recipe_dir, attrs)
+                if muid is False:
+                    raise ValueError("migrate returned False")
```
beckermr: We should not raise here. Instead we should condition the next statements on if muid is a dict and return the value.

beckermr:
The error strings will likely appear on the status page. So if they really are only due to some actual bug in the bot, then a string 

"The autotick bot encountered a fatal error (<description of error>). Please raise an issue at regro/cf-scripts."

would be good since it would help users know what to do.

PR #5044, opened 2025-11-12
Add default expansion for potentially unbound variable

zklaus:
#### Description:

When the build script sets `set -u`, accessing unbound variables exits the script.
The current `[[ ${CROSSCOMPILING_EMULATOR} == "" ]]` test can run into this issue as seen in https://github.com/conda-forge/brotli-feedstock/pull/45 for the osx-arm64 build.
While `set -u` is rare and the combination with other circumstances even rarer, we should use default expansion for this variable just as we do for the other.

#### Checklist:

### DIFF
--- a/conda-lock.yml
+++ b/conda-lock.yml
@@ -2289,36 +2289,36 @@ package:
   category: main
   optional: false
 - name: conda-forge-pinning
-  version: 2025.11.12.00.33.58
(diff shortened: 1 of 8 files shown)


h-vetinari:
Thanks for the PR! :)

Looks like you need to update test expectations, for example in `tests/test_cross_compile.py::test_make_check`.

Issue #5150, opened 2025-12-04
PR shown as "passing" in migration, should be "conflicts"


Carreau:
https://github.com/conda-forge/memray-feedstock/pull/56 has been likely conflict for 3weeks, but is still in it's json file in cf-graph-countyfair `$ cat pr_json/4/1/4/9/3/2986639532.json` (https://github.com/regro/cf-graph-countyfair/blob/master/pr_json/4/1/4/9/3/2986639532.json) still mergeable "clean". 

This file was not updated in a month, my guess by bad luck as only some PRs get updated from update_pr ? That randomly update only 50% of them ?

Or any other ideas ? Maybe keep track of when the file was last update and always update the ~20 oldest updated ?

Carreau:
And I did post a comment on the issue, so I'm stupid, because now it's modified; but I'm sure it's not the only one.

PR #5151, opened 2025-12-05
Don't trust last-modified for PR older then 7 days

Carreau:
See #5150, GitHub "Last-Modified" seem to means that the PR itself has not received any modifications; but it _may_ be conflicting now, The etag seem better but is also a weak etag.

Therefore, if Last-Mofified is older than 7 days, skip it, refresh the PR and replace the Last_modified with the date, this prevent PRs which really have identical outputs because not modified to get force-refreshed only once every weeks.

This only affects PRs where the status is passing, and merging is clean as we don't really care for others (I think).

304 do count on API calls, so maybe we should just care less, and make unconditional calls ?

I also added the ability to locally test on a single feedstock (it works), and modified the --offline to also not try to close PRs. Otherwise it fais as my GH TOKEN is read-only; well, I was debugging with a RW token yesterday and eded up commenting on a feedstock I did not meant to...

#### Checklist:

#5150

### DIFF
--- a/conda-lock.yml
+++ b/conda-lock.yml
@@ -974,7 +974,7 @@ package:
   category: main
   optional: false
 - name: botocore
-  version: 1.42.5
+  version: 1.42.6
   manager: conda
   platform: linux-64
   dependencies:
@@ -982,14 +982,14 @@ package:
     python: '>=3.10'
     python-dateutil: '>=2.1,<3.0.0'
     urllib3: '>=1.25.4,!=2.2.0,<3'
-  url: https://conda.anaconda.org/conda-forge/noarch/botocore-1.42.5-pyhd8ed1ab_0.conda
+  url: https://conda.anaconda.org/conda-forge/noarch/botocore-1.42.6-pyhd8ed1ab_0.conda
   hash:
-    md5: 94b994963a98c8a0b64b85c791ae6836
-    sha256: abfa7c8159bbfc4ef59cb6ab9bda733d6fdc573a69b06eff86b58623a8b57fb0
+    md5: d185c49ce23a9f3d5070335d02262950
+    sha256: 1e97839e16d2338feb12bbc2655808bed703039f604b1829e5d215c5e5394048
   category: main
   optional: false
 - name: botocore
-  version: 1.42.5
+  version: 1.42.6
   manager: conda
   platform: osx-64
   dependencies:
@@ -997,14 +997,14 @@ package:
     python: '>=3.10'
     python-dateutil: '>=2.1,<3.0.0'
     urllib3: '>=1.25.4,!=2.2.0,<3'
-  url: https://conda.anaconda.org/conda-forge/noarch/botocore-1.42.5-pyhd8ed1ab_0.conda
+  url: https://conda.anaconda.org/conda-forge/noarch/botocore-1.42.6-pyhd8ed1ab_0.conda
   hash:
-    md5: 94b994963a98c8a0b64b85c791ae6836
-    sha256: abfa7c8159bbfc4ef59cb6ab9bda733d6fdc573a69b06eff86b58623a8b57fb0
+    md5: d185c49ce23a9f3d5070335d02262950
+    sha256: 1e97839e16d2338feb12bbc2655808bed703039f604b1829e5d215c5e5394048
   category: main
   optional: false
 - name: botocore
-  version: 1.42.5
+  version: 1.42.6
   manager: conda
   platform: osx-arm64
   dependencies:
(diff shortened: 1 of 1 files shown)


Carreau:
```shell
cf-graph-countyfair $ git diff -G'mergeable_state' --name-only | xargs -n1 jq -r '.html_url'
... 
```

To see affects PRs

conda_forge_tick/git_utils.py
```diff
@@ -1662,7 +1671,27 @@ def refresh_pr(
...
+            if pr_json.get("mergeable_state") == "clean" and "Last-Modified" in pr_json:
+                try:
+                    # Parse Last-Modified (Date header) to check if cached data is >7 days old
+                    last_modified = utils.parsedate_to_datetime(
+                        pr_json["Last-Modified"]
+                    )
+                    now = datetime.now(last_modified.tzinfo)
+                    age_days = (now - last_modified).total_seconds() / 86400
+
+                    if age_days > 7:
```
beckermr: The seven day threshold needs to be added to the bot settings module.

conda_forge_tick/update_prs.py
```diff
@@ -159,16 +202,60 @@ def update_pr_combined(
...
+    job : int, optional
+        The job number (1-indexed) for parallel processing. Default is 1.
+    n_jobs : int, optional
+        The total number of jobs for parallel processing. Default is 1.
+    feedstock : str | None, optional
+        If provided, only update PRs for this specific feedstock.
+        Must end with "-feedstock" suffix (validated in main()).
+    offline : bool, optional
+        If True, skip operations that write to GitHub (close_out_labels, close_out_dirty_prs).
+        Default is False.
```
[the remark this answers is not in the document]
Carreau: I can just remove it for now, it was just useful to test locally for now if that's easier.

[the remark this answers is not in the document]
beckermr: That's fine by me.

Carreau:
While I think about writing it down, it looks like this also both:

 - shuffle node ids, 
 - Keep 1/2 of the selected ids with a random 50% threshhold, 
 
Are doing the same.

I think if we shuffle then keeping only 1/2 is about the same and avoid many RNG generation and skip some work.

conda_forge_tick/config_schema.py
```diff
@@ -246,6 +246,13 @@ class BotConfig(BaseModel):
         description="Automatically remake untouched bot PRs with conflicts.",
     )
 
+    pr_refresh_age_days: Optional[float] = Field(
+        default=7.0,
+        description="Number of days after which to refresh PR cache for 'clean' PRs "
+        "to detect potential conflicts. Works around GitHub API bug #5150 where "
+        "Last-Modified caching can hide merge conflicts. Set to 0 to always refresh.",
+    )
```
beckermr: This is in the wrong file. Internal settings for the bot go in this file: https://github.com/regro/cf-scripts/blob/main/conda_forge_tick/settings.py

Carreau: Yep, that make a lot more sens.

conda_forge_tick/git_utils.py
```diff
@@ -1657,12 +1662,46 @@ def refresh_pr(
...
+            # See https://github.com/regro/cf-scripts/issues/5150
+            # we could only do so for "clean" PR, but as the 304 anyway consume API
+            # might as well do it.
+            # This also ensure all the json will soon get the 'last_fetched' key
+            # and logic can be simplified.
+            force_refresh = False
+            if "last_fetched" not in pr_json:
+                # TODO: Dec 2025, not all pr_json have 'last_fetch',
+                # this will need to be simplified in a few weeks once the bot has
+                # updated every PRs.
```
beckermr:
```suggestion
```

We should leave this block in permanently.

beckermr (changes requested):
A few more items.

conda_forge_tick/update_prs.py
```diff
@@ -63,11 +63,45 @@ def _combined_update_function(
...
+        Filtered node IDs matching the feedstock.
+    """
+    if not feedstock_filter:
+        return node_ids
+
+    return [node_id for node_id in node_ids if node_id == feedstock_filter]
+
+
+def _update_pr(
+    update_function, dry_run, gx, job, n_jobs, feedstock_filter=None, offline=False
```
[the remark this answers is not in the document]
Carreau:
```suggestion
    update_function, dry_run, gx, job, n_jobs, feedstock_filter=None
```

Yes oversignt from my part.

Carreau:
Thanks !

Carreau:
ok, None has no attribute tzinfo:

see #5193

PR #5193, opened 2025-12-11
None has not attribute tzinfo

Carreau:
Should fix some errors in #5190 

#### Checklist:


Carreau:
@beckermr

h-vetinari (commented):
I feel we're creating multiple paths here where the logic really only needs to know whether "`last_fetched`  is useful" or not.

I'd suggest to merge the "cannot be used" cases by using something along the lines of
```python
last_fetched: str | None = pr_json.get("last_fetched", None)
```

Also, the `try:` block should really be limited to where the parsing can go wrong, not stretch for 10s of lines

PR #5194, opened 2025-12-11
feat: simpler last fetched code

beckermr:
#### Description:

Per comments from @h-vetinari we can simplify a lot here.

#### Checklist:

### DIFF
--- a/conda_forge_tick/git_utils.py
+++ b/conda_forge_tick/git_utils.py
@@ -1679,7 +1679,7 @@ def refresh_pr(
             # This also ensure all the json will soon get the 'last_fetched' key
             # and logic can be simplified.
             last_fetched: str | None = pr_json.get("last_fetched", None)
-            if last_fetched
+            if last_fetched:
                 last_fetched_date: datetime = dateutil.parser.isoparse(last_fetched)
             else:
                 last_fetched_date = datetime.now() - timedelta(


conda_forge_tick/git_utils.py
```diff
@@ -1678,32 +1678,22 @@ def refresh_pr(
             # might as well do it.
             # This also ensure all the json will soon get the 'last_fetched' key
             # and logic can be simplified.
-            force_refresh = False
+            last_fetched: str | None = pr_json.get("last_fetched", None)
+            if isinstance(last_fetched, str):
```
beckermr:
```suggestion
            if last_fetched
```

conda_forge_tick/git_utils.py
```diff
@@ -1678,32 +1678,22 @@ def refresh_pr(
             # might as well do it.
             # This also ensure all the json will soon get the 'last_fetched' key
             # and logic can be simplified.
-            force_refresh = False
+            last_fetched: str | None = pr_json.get("last_fetched", None)
+            if last_fetched:
+                last_fetched_date: datetime = dateutil.parser.isoparse(last_fetched)
```
h-vetinari: this can still fail if (for whatever reason) `pr_json["last_fetched"]` is present but not parseable.

conda_forge_tick/git_utils.py
```diff
@@ -1678,32 +1678,22 @@ def refresh_pr(
...
+            else:
+                last_fetched_date = datetime.now() - timedelta(
+                    days=pr_refresh_age_days + 1
+                )
```
h-vetinari: I get that this unifies the `if age_days > pr_refresh_age_days:`, but I find it circuitous to construct just the right value that then _fails_ the condition.

h-vetinari (commented):
_de gustibus non est disputandum_ as the Romans used to say. Which is to say, I prefer my version 😋

PR #5195, opened 2025-12-11
simplify logic in git_utils

h-vetinari:
Follow-up to #5193


beckermr:
Something is wrong with the pydantic model

beckermr:
I found it. https://github.com/regro/cf-scripts/pull/5197

My bug from an earlier PR.

PR #5610, opened 2026-02-21
make sure parse_recipe_yaml is called with cbc_path

### LINKED ISSUE #5607: make_migrator doesn't load variables from conda_build_config.yaml
I'm trying to debug why several migrators aren't getting created.

At least:

- mumps_mpi
- impi_devel
- mpich

are not getting version bumps from the bot

When testing with the cf-scripts repo:

```
migrators = []
create_migration_yaml_creator(migrators, gx, pin_to_debug="impi_devel")
```

tries to produce:

```
impi_devel:
    package name: impi-devel
    feedstock name: intel_repack
    possible pin spec: {'package_name': 'mkl-dpcpp'}
    possible pin spec: {'package_name': 'mkl'}
    possible pin spec: {'package_name': 'dal'}
    possible pin spec: {'package_name': 'impi_rt', 'min_pin': 'x.x.x', 'max_pin': 'x.x.x'}
    migrator:
        curr version: 2025.10.0
        curr pin: 2021.16.0
        pin_spec: x
        pinnings: ['impi_devel']
```

but the version number is wrong. The current version is actually `2021.17.0`. I guess this is because the output version doesn't match the recipe version, which is not handled.

```python
create_migration_yaml_creator(migrators, gx, pin_to_debug="mpich")
```

fails with:

```
CRITICAL:error parsing recipe.yaml:

Error:   × Failed to parse recipe

Error: 
  × failed to parse match spec: unable to parse version spec: .*
    ╭─[../../../../../private/var/folders/nd/nw_2h27d0636jdhk85jq_zmm0000gn/T/.tmpUUt4GH/recipe.yaml:44:15]
 43 │           then:
 44 │             - libfabric-devel ${{ libfabric }}.*
    ·               ─────────────────┬────────────────
    ·                                ╰── error parsing `libfabric-devel .*` as a match spec
 45 │             - libhwloc
    ╰────

```

because it doesn't load variables from the recipe's `conda_build_config.yaml` when rendering. Should we be migrating this to `variants.yaml` in the recipe, or should the bot be taking conda_build_config into account?

I can't currently see why mumps_mpi is wrong right now because I did the migration by hand after it wasn't issued for some months by the bot.

minrk:
#### Description:

it's frequently required to parse recipes (v1 at least) with cbc config, either from the recipe or .ci_support.

I only found one ci_support record in feedstock_attrs, so use that. I'm not sure if there's a better value to use or a better way to find the key? I could only find "the first one that starts with ci_support".

Fixes, for example,

```
create_migration_yaml_creator(migrators, gx, pin_to_debug="mpich")
```

#### Checklist:

closes #5607

### DIFF
--- a/conda_forge_tick/make_migrators.py
+++ b/conda_forge_tick/make_migrators.py
@@ -9,6 +9,8 @@ import secrets
 import sys
 import time
 from concurrent.futures import as_completed
+from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import (
     Any,
     List,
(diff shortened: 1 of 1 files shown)


beckermr (approved):

minrk:
This appeared to fix the mpich bot PRs: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8269

PR #5901, opened 2026-04-12
fix: ensure we have a separate log-level bit in commands

beckermr:
#### Description:

The CLI had an override of `--debug` in a subcommand. This PR cleans that up by adding a `--log-level` option.

#### Checklist:


conda_forge_tick/cli.py
```diff
@@ -75,8 +78,9 @@ def main(
     dry_run: bool,
     online: bool,
     no_containers: bool,
+    log_level,
```
beckermr:
```suggestion
    log_level: str,
```

PR #5928, opened 2026-04-21
fix: properly extract tags that include forward slashes

salim-b:
Fixes extracting tags from the `github` source (atom feed containing URLs of the form `https://github.com/<owner>/<repo>/releases/tag/<tag>` for all releases of a repository incl. side-packages).

Before, when a tag contained a forward slash like e.g. `sass_api/17.5.0`, everything up to and including the last slash was stripped, wrongly extracting `17.5.0` as the tag instead of the full `sass_api/17.5.0` tag name. As a consequence, the [`bot.version_updates.allowed_tag_globs`](https://conda-forge.org/docs/maintainer/conda_forge_yml/#bot) config setting wouldn't work as intended.

#### Checklist:

### DIFF
--- a/conda_forge_tick/update_sources.py
+++ b/conda_forge_tick/update_sources.py
@@ -211,10 +211,15 @@ class VersionFromFeed(AbstractSource, abc.ABC):
         vers = []
 
         for entry in data["entries"]:
-            ver = urllib.parse.unquote(entry["link"]).split("/")[-1]
+            ver = urllib.parse.unquote(entry["link"].split("/")[-1])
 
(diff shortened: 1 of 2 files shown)


beckermr (changes requested):
Thanks for the PR! Please add a test!

salim-b:
> Thanks for the PR! Please add a test!

Done :)

beckermr (approved):
The changes and tests look good. This could break existing globs on other feedstocks, but that is a risk we'll have to take I think.

PR #5935, opened 2026-04-22
fix: better PR instructions

beckermr:
Hopefully this will reduce confusion and mistakes.

#### Description:

As it says on the tin...

#### Checklist:


beckermr:
pre-commit.ci autofix

jjrob:
Updated instructions look good. Thanks for considering all of those cases!

PR #6217, opened 2026-06-10
Convert preformatted error strings into structured payloads for status report

jaimergp:
#### Description:

This PR introduces new fields that supersede their value-is-a-string counterparts with structured payloads. This way we can avoid pre-formatting the error strings and let the frontend handle it. Namely:

- In `conda-forge-bot-data/status/version_status.v2.json`, `errors` points now to dictionaries containing details like `kind`, `url`, `base_branch` and `messages` (one or more plain strings).
- In `conda-forge-bot-data/status/migration_json/*.json`, `_feedstock_status.[].pre_pr_migrator_status` is now the same type of dictionary, plus an optional `attempts` float.

#### Checklist:

- https://github.com/conda-forge/conda-forge-bot/issues/6218

### DIFF
--- a/conda_forge_tick/auto_tick.py
+++ b/conda_forge_tick/auto_tick.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import gc
 import glob
 import logging
@@ -6,7 +8,7 @@ import textwrap
 import time
 import traceback
 import typing
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 from typing import Any, Literal, cast
 from urllib.error import URLError
 from uuid import uuid4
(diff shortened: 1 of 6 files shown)


beckermr (commented):
A few comments:

- we should include html escaping in the standardize string function and use that everywhere 
- we're going to duplicate a lot of data this way. I'd rather use a dynamic schema replacing the old field and then removing that handling code once we migrate the data already written.

jaimergp:
Let me know if the last commit addressed your concerns, @beckermr.

beckermr:
@jaimergp shall we drop the v0/v1 version status json blobs in this PR as well? We should keep the v2 ones ofc.

beckermr (commented):
Alright, I see what happened.

I was not clear and caused confusion. I am sorry. When I said not duplicating data, I meant in the bot node+pr+version data, not in the status blobs. We still need to write the older v2 status blob until the website is converted to use v3.

Sorry!

jaimergp:
> We still need to write the older v2 status blob until the website is converted to use v3.

I was planning to submit a PR to the website before merging here so it handles both strings and dictionaries (even if it's in a crude way with raw dumps for now). What do you think?

beckermr:
People who have not refreshed their browser caches may see an outage this way?

That is a small time window so seems fine.

jaimergp:
I would assume that new Docusaurus builds force use updated JS assets (their paths are hashed), so there shouldn't be much of an issue with stale caches. I also [asked](https://chatgpt.com/s/t_6a297b443fa08191bf229861a49abf86) ChatGPT and seems to be true. So yea, short window and theoretically no stale caches = not much of a risk.

beckermr:
I've had consistent problems with stale caches in my browser, but that could be me. Anyhow, happy to proceed this way!

jaimergp:
See https://github.com/conda-forge/conda-forge.github.io/pull/2854

PR #6251, opened 2026-06-18
Add a migrator for moving from cross to native `linux_aarch64` builds

### LINKED ISSUE #2852: A migrator (or a hint) for moving from cross-compilation to native GHA builds
### Your question:

Now that we have native `linux_aarch64`, <del>`linux_ppc64le`</del>, `osx_arm64` runners, I suppose it makes sense to start moving feedstocks away from cross-compiling and to native builds on these platforms. Especially that people seem to be still unaware that they don't need to add platform support via cross anymore.

What I'm thinking of is creating a migrator that would try switching feedstocks from cross compilation to native builds for the platforms that support it. For example, from:

```yaml
provider:
  linux_64: github_actions
build_platform:
  linux_aarch64: linux_64
```

to:

```yaml
provider:
  linux_64: github_actions
  linux_aarch64: github_actions
```

Alternatively, if migrator is too aggressive, we could have conda-smithy emit a hint that native builds could be used. However, I think migrator is better since it saves users from doing this manually.

mgorny:
#### Description:

Add a migrator that detects feedstocks using `linux_64` GHA runners to cross-compile to `linux_aarch64`, and attempt switching them to native `linux_aarch64` runners.  This only covers GHA, as Azure does not feature native AArch64 runners.  For now, we're covering only `linux_aarch64`, since `osx_arm64` needs concurrent build limit raised first.

#### Checklist:

Fixes conda-forge/conda-forge.github.io#2852

### DIFF
--- a/conda_forge_tick/migrators/cross_to_native.py
+++ b/conda_forge_tick/migrators/cross_to_native.py
@@ -1,0 +1,104 @@
+from pathlib import Path
+from typing import Any
+
+from conda_forge_tick.contexts import ClonedFeedstockContext, FeedstockContext
+from conda_forge_tick.migrators.core import Migrator
+from conda_forge_tick.migrators_types import (
+    AttrsTypedDict,
+    CondaForgeYamlContents,
+    MigrationUidTypedDict,
+)
+from conda_forge_tick.utils import (
+    yaml_safe_dump,
+    yaml_safe_load,
+)
+
+
+class CrossToNativeMigrator(Migrator):
+    name = "Cross-to-native Migrator"
+    rerender = True
+    migrator_version = 1
+
+    # platforms to migrate
+    _platforms = {"linux_aarch64"}
+    # build platforms where default = github_actions
+    _gha_platforms = {"linux_64"}
+
+    def _migrate_platform(self, platform: str, cfyaml: CondaForgeYamlContents) -> bool:
+        build_platform = cfyaml.get("build_platform", {}).get(platform)
+        if build_platform is None:
+            return False
+        provider = cfyaml.get("provider", {}).get(build_platform, "default")
+        return provider == "github_actions" or (
+            provider == "default" and build_platform in self._gha_platforms
+        )
+
+    def filter_not_in_migration(
+        self, attrs: AttrsTypedDict, not_bad_str_start: str = ""
+    ) -> bool:
+        if super().filter_not_in_migration(attrs, not_bad_str_start):
+            return True
+
+        # TODO: check if there are any relevant cross-targets
+        cfyaml = attrs.get("conda-forge.yml", {})
+        for platform in self._platforms:
+            if self._migrate_platform(platform, cfyaml):
+                return False
+
+        return True
+
+    def migrate(
+        self, recipe_dir: str, attrs: AttrsTypedDict, **kwargs: Any
+    ) -> MigrationUidTypedDict:
+        # Only v0 recipes are supported, the handful of v1 recipes is not worth the complexity.
+        recipe_file = next(
+            filter(
+                lambda x: x.exists(),
+                (Path(recipe_dir) / "recipe.yaml", Path(recipe_dir) / "meta.yaml"),
+            )
+        )
+        self.set_build_number(recipe_file)
+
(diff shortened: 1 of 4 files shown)


conda_forge_tick/migrators/cross_to_native.py
```diff
@@ -0,0 +1,104 @@
...
+    def migrate(
+        self, recipe_dir: str, attrs: AttrsTypedDict, **kwargs: Any
+    ) -> MigrationUidTypedDict:
+        # Only v0 recipes are supported, the handful of v1 recipes is not worth the complexity.
+        recipe_file = next(
+            filter(
+                lambda x: x.exists(),
+                (Path(recipe_dir) / "recipe.yaml", Path(recipe_dir) / "meta.yaml"),
+            )
+        )
```
pb01ka: I think `next()` without a default would raise `StopIteration` when neither file exists (**an edge case**), which would surface as a confusing error. May be, providing a default (and raising a descriptive `FileNotFoundError`) would be good?

May be absence of both files (i.e., `recipe.yaml` and `meta.yaml`) in the `recipe_dir` is not something that would happen practically. However imagine if by chance someone deleted both the files by mistake and then calls the bot. Just a corner case, not sure how practical would it be to handle this. I leave the decision upto you. :-).

```python
recipe_file = next(
    filter(
        lambda x: x.exists(),
        (Path(recipe_dir) / "recipe.yaml", Path(recipe_dir) / "meta.yaml"),
    ),
    None,
)
if recipe_file is None:
    raise FileNotFoundError(f"No recipe file found in {recipe_dir}")
```

I also noticed the current pattern is used elsewhere in the codebase as well. I have also noticed the usage of `os.path.exists` over `Path.exists` (the latter is preferred way and we should be using `Path.exists` everywhere else).

pb01ka (approved):
Other than the minor comment (addressing it is fully optional), this PR looks excellent. I would give a +1 for merging it.

conda_forge_tick/migrators/cross_to_native.py
mgorny: I don't think there's a use case where a migrator could work without a recipe file. They will pretty much all explode.

pb01ka: Sure. Please feel free to merge. Nice work. 🚀 💯

conda_forge_tick/migrators/cross_to_native.py
```diff
@@ -0,0 +1,104 @@
...
+        with open(cfyaml_path) as fp:
+            cfyaml = yaml_safe_load(fp)
+
+        for platform in self._platforms:
+            if not self._migrate_platform(platform, cfyaml):
+                continue
+
+            build_platform = cfyaml["build_platform"].pop(platform)
+            provider = cfyaml.setdefault("provider", {}).get(build_platform, "default")
+            cfyaml["provider"][platform] = provider
```
isuruf: I'd rather use "default" here regardless of the value for build_platform

isuruf (changes requested):

isuruf (commented):
I think we'd want to use a minimigrator instead of a full migrator sending PRs.

mgorny:
That was deliberate, since the change could be introducing test failures due to native testing. But then, I guess random test regressions do happen anyway, and there's no reason to be building new packages over this.

mgorny:
> I think we'd want to use a minimigrator instead of a full migrator sending PRs.

Done.

mgorny:
Wait, sorry, I forgot to update the integration!

tests/test_cross_to_native.py
```diff
@@ -0,0 +1,160 @@
...
+    expected_providers = ""
+    if provider is not None:
+        expected_providers += f"  linux_64: {provider}\n"
+
+    assert (
+        cfyaml.read_text()
+        == f"""\
+build_platform:
+  linux_ppc64le: linux_64
+  osx_arm64: osx_64
```
isuruf, line 76: Can move `osx_arm64` to native too?

mgorny: Sure. Actually, I should check which of our CIs have which native runners.

PR #6253, opened 2026-06-19
add excluded_dependencies for win-arm64 migrator

h-vetinari:
For the win-arm64 [migrator](https://conda-forge.org/status/migration/?name=supportwindowsarm64platform), a lot of R packages are included, because the bot sees that `ci-setup` depends on `r-languageserver`. Filter out that package so we can get rid of a bunch of dependencies in the graph that we don't need to bootstrap the build tools.

### DIFF
--- a/conda_forge_tick/migrators/arch.py
+++ b/conda_forge_tick/migrators/arch.py
@@ -39,7 +39,8 @@ def _filter_excluded_deps(graph, excluded_dependencies):
     """
     nodes_to_remove = set(excluded_dependencies)
     for excluded_dep in excluded_dependencies:
-        nodes_to_remove |= set(nx.descendants(graph, excluded_dep))
+        if excluded_dep in graph.nodes:
+            nodes_to_remove |= set(nx.descendants(graph, excluded_dep))
     for node in nodes_to_remove:
         pluck(graph, node)
     # post-plucking cleanup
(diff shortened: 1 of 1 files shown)


h-vetinari:
This should help with the win-arm [migration](https://conda-forge.org/status/migration/?name=supportwindowsarm64platform), in the sense that it should remove a lot of packages that we don't need for bootstrapping the build tools. This issue was called out specifically by @isuruf in his [blog post](https://conda-forge.org/blog/2026/02/09/win-arm64/):
> Looking at the graph, there are nodes that are clearly unneeded for the packages given in the list. Help is welcome to debug the bot on trimming down the graph here so we can focus on the packages actually needed for getting the build tools out.

In terms of the diff here, I don't understand why `nx.descendants` fails for `r-languageserver`, but it's obviously and definitely part of the graph... Either way, I think the guard I added makes sense, so I don't consider this particularly hacky.

h-vetinari:
Gentle ping @beckermr @isuruf. I'd like to try this to see if it cuts down the useless deps in the migrator.

isuruf (approved):

h-vetinari:
Good news: this substantially reduced the amount of spurious dependencies, see [status page](https://conda-forge.org/status/migration/?name=supportwindowsarm64platform) 🥳

Issue #6272, opened 2026-06-25
Bot failure when v1 recipe does not support `linux_64` (`pyobjc-framework-quartz`)


mgorny:
`pyobjc-framework-quartz` has the following `skip:` condition:

```yaml
  skip: (not osx) or match(python, ">=2,<3") or python_impl == 'pypy' or (arm64 and match(python, "<3.9"))
```

This causes the following failure:

```
CRITICAL conda_forge_tick.utils:utils.py:547 error parsing recipe.yaml:
[]

 Processing recipe at path: /tmp/.tmpi9gxcq/recipe.yaml

 ╭─ Finding outputs from recipe
 │ Found 1 variants
 │
 │ Build variant: pyobjc-framework-quartz-12.2-h81b34b9_0 (skipped)
 │
 │ ╭─────────────────┬────────────╮
 │ │ Variant         ┆ Version    │
 │ ╞═════════════════╪════════════╡
 │ │ build_platform  ┆ "linux-64" │
 │ │ target_platform ┆ "linux-64" │
 │ ╰─────────────────┴────────────╯
 │
 ╰─────────────────── (took 0 seconds)
 Skipping pyobjc-framework-quartz 12.2 - skip conditions evaluated to true
```

mgorny:
#6273 is adding an XFAIL-ing test.

mgorny:
I was trying to debug this further but it seems that `conda-forge-tikc update-upstream-versions` insist on using the remote data, and I can't figure out how to make it trigger just the right bot command for parsing the feedstock.

mgorny:
Okay, I've figured out I just need to call:

```python
from conda_forge_tick.feedstock_parser import load_feedstock

load_feedstock("pyobjc-framework-quartz", {})
```

And apparently the issue is that we're not passing `--target-platform` when `--variant-config` is passed, except that rattler-build doesn't support getting it from the variant config:

```
 │ ⚠ warning Setting 'target_platform' in a variant config file is not supported and will be ignored. Please use the '--target-platform' command-line flag to spe
 │ ⚠ warning cify the target platform.
```

I have a trivial fix, but will have to figure out a proper test case for this.

PR #6273, opened 2026-06-25
Add an XFAIL-ing test case for `pyobjc-framework-quartz`

mgorny:
#### Description:

Add a `pyobjc-framework-quartz` per the failures in https://github.com/conda-forge/conda-forge-bot/issues/6272.  It is failing due to the render happening with `linux-64`.  The test also fails due to the `match(python, ...)` skip but that might be just a harness problem.

#### Checklist:

#6272

### DIFF
--- a/tests/test_v1_yaml/version_pyobjc.yaml
+++ b/tests/test_v1_yaml/version_pyobjc.yaml
@@ -1,0 +1,70 @@
+schema_version: 1
+
+context:
+  name: pyobjc-framework-Quartz
+  version: "12.2"
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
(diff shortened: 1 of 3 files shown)


mgorny:
From my quick debugging, the following calls are happening:

```
['rattler-build', 'build', '--render-only']
['rattler-build', 'build', '--render-only', '--variant-config', '/tmp/pytest-of-mgorny/pytest-33/test_version_up_v1_pyobjc_12_20/.ci_support/linux_64_.yaml']
```

Where the YAML file is:

```yaml
target_platform:
- linux-64
```

mgorny:
Okay, this is wrong. It's failing entirely because of the test harness. I'm going to start with a fix, and then try to figure out how to test it properly.

PR #6319, opened 2026-07-16
Make win-arm64 and osx-arm64 migrators native

jaimergp:
#### Description:

Once smithy 2026.7 is out, GHA will be the default CI provider for Windows x64 and arm64. This means we don't need to cross-compile. I want to avoid the migrator creating cross-compiling PRs when that happens. Is this the way to prevent that from happening? Or does it need a new migrator?

We can do the same with `osx-arm64`, so I also added it here.

#### Checklist:

- https://github.com/conda-forge/conda-forge.github.io/issues/1940


conda_forge_tick/migrators/arch.py
```diff
@@ -434,10 +434,10 @@ class WinArm64(_CrossCompileRebuild):
     """A Migrator that adds win-arm64 builds to feedstocks."""
 
     allowed_schema_versions = {0, 1}
-    migrator_version = 1
-    build_platform = {"win_arm64": "win_64"}
+    migrator_version = 2
+    build_platform = {"win_arm64": "win_arm64"}
     pkg_list_filename = "win_arm64.txt"
-    arches = {"win_arm64": "win_64"}
+    arches = {"win_arm64": "default"}
```
jaimergp, line 521: The values in this dict are not being used at all. In the Linux ARM/PPC migrator this is a dictionary for providers. Is this correct?

jaimergp: I adjusted the code so `arches` is the source for `provider`. The iteration over `arches` is now done on `build_platform` instead.

conda_forge_tick/migrators/arch.py
```diff
@@ -393,10 +395,10 @@ class OSXArm(_CrossCompileRebuild):
     """A Migrator that adds osx-arm64 builds to feedstocks."""
 
     allowed_schema_versions = {0, 1}
-    migrator_version = 1
-    build_platform = {"osx_arm64": "osx_64"}
+    migrator_version = 2
+    build_platform = {"osx_arm64": "osx_arm64"}
```
jaimergp: I think we can also clear this one because what matters is that `arches` defines the non-default provider.

conda_forge_tick/migrators/arch.py
```diff
@@ -393,10 +395,10 @@ class OSXArm(_CrossCompileRebuild):
     """A Migrator that adds osx-arm64 builds to feedstocks."""
 
     allowed_schema_versions = {0, 1}
-    migrator_version = 1
-    build_platform = {"osx_arm64": "osx_64"}
+    migrator_version = 2
```
jaimergp: Do we need to bump this? Will this restart the migrator?

conda_forge_tick/migrators/arch.py
h-vetinari: yes please

conda_forge_tick/migrators/arch.py
h-vetinari: I don't think we want to bump this, because yes, new migrator_version cause everything in scope to be rebuilt; we don't want to rebuild all feedstocks that added osx-arm64 builds.

h-vetinari:
I know we already have [several feedstocks](https://github.com/search?q=org%3Aconda-forge+%2Fwin_arm64%3A+default%2F+path%3Aconda-forge.yml&type=code) on `provider: win_arm64: default`. I don't necessarily understand how that works, given that a lot of build tools are not done yet. Just wanted to check what "water mark" we need to achieve on this before going ahead here, given PRs like e.g. https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8887

edit: just saw this [comment](https://github.com/conda-forge/conda-forge.github.io/issues/1940#issuecomment-5203451955) (which had been lost in my notifications) which explains the situation

jaimergp:
So on the native track we are missing `make` because of missing Mingw GCC. On the cross-compiled track we are missing `cross-python`, so any Python package fails.

Edit: Isuru pointed out that we can use `m2-make` instead of `make`, in those cases.

jaimergp:
Added a change so the native case only applies to packages having `python` in their `host`

h-vetinari:
> Added a change so the native case only applies to packages having `python` in their `host`

I think we can revert that now. IMO we should build everything natively, now that all the build tools up until the ci setup are in place.

jaimergp:
> IMO we should build everything natively, now that all the build tools up until the ci setup are in place.

That reminds me, we need to revert the bootstrapping phase changes in smithy and ci-setup IIRC.

h-vetinari:
I don't know which changes you're referring to; I found https://github.com/conda-forge/conda-smithy/pull/2622 (where something can be fixed), but for the CI setup, the related changes (like https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/422 etc.) appear valid even after going native; IOW, I'm not sure what you had in mind that needs reverting here

conda_forge_tick/migrators/arch.py
```diff
@@ -177,7 +177,7 @@ class _ArchesConfiguredMixin:
     itself rather than only in the bot's PR records.
     """
 
-    arches: dict = {}
+    arches: dict[str, str] = {}
```
h-vetinari, line 180: it might be worthwhile to consider renaming this to `providers` (perhaps in a separate PR), because `arches` is kinda misleading here IMO

h-vetinari:
> That reminds me, we need to revert the bootstrapping phase changes in smithy and ci-setup IIRC.

The smithy changes were fixed/reverted in https://github.com/conda-forge/conda-smithy/pull/2680 AFAIU; not sure what's you had remembered w.r.t. the ci-setup

conda_forge_tick/migrators/arch.py
h-vetinari: Code here uses both currently
https://github.com/conda-forge/conda-forge-bot/blob/adf6f6feefad683a6ac96fd93a2055224e46220e/conda_forge_tick/migrators/arch.py#L298-L302, so out of scope for now

PR #6381, opened 2026-07-28
Add R feedstock v0 -> v1 recipe migration: `GenericV0ToV1Migrator` + `RV0ToV1Migrator`

pb01ka:
### Description

This is step 2 of the [R feedstock v0 -> v1 migration plan](https://gist.github.com/pb01ka/4d915b2196470ca7c95643c31212b52b), split into two layers, both planned in this PR:

- **`GenericV0ToV1Migrator`** (implemented): a `MiniMigrator` class that only knows the `conda-recipe-manager` (`crm`) conversion mechanics. It has no R-specific logic.
- **`RV0ToV1Migrator`** (in progress, to be pushed in next commit of this PR): an R-specific subclass layered on top of the base class, adding the R-specific pre/post-processing and registered for R feedstocks. See TODOs below for what's still outstanding.

#### Why split into a base class + subclass

Keeping the `crm`-mechanics layer free of R-specifics means:

- It's independently testable against `crm`'s actual behavior (raised exceptions vs. warnings vs. errors), without needing R recipe fixtures.
- It's reusable if we migrate other package clusters through the same v0 -> v1 mechanism later.
- `RV0ToV1Migrator` can extend `IGNORED_WARNINGS` and add its own pre/post-processing without touching the safe-to-auto-convert decision itself.

### TODOs - remaining work in this PR

- [ ] Implement `RV0ToV1Migrator`: layer R-specific pre/post-processing on top of `GenericV0ToV1Migrator`: `{{ compiler('c') }}` / `{{ stdlib('c') }}` / `native`/`posix`/`m2w64` token cleanup (comparable to what `RUCRTCleanup` already does for the ucrt migration), and multi-output handling for CRAN "recommended" vs. non-recommended packages.
- [ ] Re-run the feasibility sample against a larger set of R feedstocks once `RV0ToV1Migrator` exists, to catch any remaining edge cases.
- [ ] Multi-output recipes (e.g. `r-base` itself) currently crash crm's `render_to_v1_recipe_format()` with a raw `AttributeError` in its dependency-upgrade logic; `RV0ToV1Migrator` catches this and skips the recipe rather than converting it. Add a local workaround so multi-output CRAN "recommended" vs. non-recommended recipes actually convert instead of always being skipped.

### Cross-refs, links to issues, etc:

https://gist.github.com/pb01ka/4d915b2196470ca7c95643c31212b52b

### DIFF
--- a/conda_forge_tick/migrators/r_v0_to_v1.py
+++ b/conda_forge_tick/migrators/r_v0_to_v1.py
@@ -1,0 +1,41 @@
+import typing
+
+from conda_forge_tick.migrators.v0_to_v1 import GenericV0ToV1Migrator
+
+if typing.TYPE_CHECKING:
+    from ..migrators_types import AttrsTypedDict
+
+
+def _is_r_feedstock(attrs: "AttrsTypedDict") -> bool:
+    return (
+        attrs.get("feedstock_name", "").startswith("r-")
+        or attrs.get("name", "").startswith("r-")
+    ) and (
+        "r-base" in attrs.get("raw_meta_yaml", "")
+        or "r-base" in attrs.get("requirements", {}).get("run", set())
+    )
+
+
+class RV0ToV1Migrator(GenericV0ToV1Migrator):
+    """R-specific v0 -> v1 recipe migrator.
+
+    R recipes near-universally template their compiler and Windows/ucrt
+    cross-compilation requirements with ``{{ compiler(...) }}`` and the
+    ``native``/``posix``/``m2w64`` substitution variables. crm already
+    converts these to correct v1 ``if``/``then`` selector syntax; it just
+    can't upgrade the (non-existent) version constraint on a templated
+    dependency, and warns about that on every single one. Ignoring that
+    warning here is safe: there was no version constraint on these
+    dependencies in the v0 recipe for crm to preserve in the first place,
+    so the produced recipe.yaml is unaffected.
+    """
+
+    IGNORED_WARNINGS = GenericV0ToV1Migrator.IGNORED_WARNINGS + (
+        "Recipe upgrades cannot currently upgrade ambiguous version "
+        "constraints on dependencies that use variables:",
+    )
+
+    def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
+        if super().filter(attrs, not_bad_str_start):
+            return True
+        return not _is_r_feedstock(attrs)
--- a/conda_forge_tick/migrators/r_v0_to_v1.py
+++ b/conda_forge_tick/migrators/r_v0_to_v1.py
@@ -53,6 +53,25 @@ class RV0ToV1Migrator(GenericV0ToV1Migrator):
             return self._SPDX_OVERRIDES[lic]
         return super()._to_spdx(lic)
 
+    def _build_script_review_reasons(self, build_sh: str) -> list[str]:
+        # In real feedstocks carrying this pattern (e.g. r-pca3d, r-msqc),
+        # this install_name_tool fix is nested inside `if target_platform ==
+        # osx-64` - but only reachable from the branch that already excludes
+        # osx-64 (the one osx-arm64 falls into), making it dead code that
+        # predates osx-arm64 support entirely.
+        if (
+            "install_name_tool -change" in build_sh
+            and "R.framework/Versions" in build_sh
+        ):
+            return [
+                "build.sh's `install_name_tool` dylib-path fixes for osx-64 are "
+                "commonly nested inside the branch that already excludes "
+                "osx-64 (the one osx-arm64 falls into) - making them dead code "
+                "that's likely never run. Verify whether that's the case here, "
+                "and if so, whether osx-arm64 needs the missing fix."
+            ]
+        return []
+
     def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
         if super().filter(attrs, not_bad_str_start):
             return True
--- a/conda_forge_tick/migrators/r_v0_to_v1.py
+++ b/conda_forge_tick/migrators/r_v0_to_v1.py
@@ -7,6 +7,10 @@ if typing.TYPE_CHECKING:
 
 
 def _is_r_feedstock(attrs: "AttrsTypedDict") -> bool:
+    # R feedstocks are all maintained by the same conda-forge/r team, e.g.
+    # https://github.com/conda-forge/r-essentials-feedstock/blob/main/recipe/meta.yaml#L69-L71
+    if "conda-forge/r" in attrs.get("extra", {}).get("recipe-maintainers", []):
+        return True
     return (
         attrs.get("feedstock_name", "").startswith("r-")
         or attrs.get("name", "").startswith("r-")
--- a/conda_forge_tick/migrators/r_v0_to_v1.py
+++ b/conda_forge_tick/migrators/r_v0_to_v1.py
@@ -39,6 +39,20 @@ class RV0ToV1Migrator(GenericV0ToV1Migrator):
         "constraints on dependencies that use variables:",
     )
 
+    # R-specific SPDX corrections, layered on top of the shared
+    # migrators/license.py mapping via _to_spdx() below - kept here, not in
+    # that shared table, since this is a CRAN-specific convention.
+    _SPDX_OVERRIDES: dict[str, str] = {
+        # CRAN's "Unlimited" license, commonly pre-wrapped as the SPDX
+        # custom-license reference "LicenseRef-Unlimited".
+        "LicenseRef-Unlimited": "Unlimited",
+    }
+
+    def _to_spdx(self, lic: str) -> str:
+        if lic in self._SPDX_OVERRIDES:
+            return self._SPDX_OVERRIDES[lic]
+        return super()._to_spdx(lic)
+
     def filter(self, attrs: "AttrsTypedDict", not_bad_str_start: str = "") -> bool:
         if super().filter(attrs, not_bad_str_start):
             return True
--- a/conda_forge_tick/migrators/v0_to_v1.py
+++ b/conda_forge_tick/migrators/v0_to_v1.py
@@ -39,6 +39,80 @@ _QUOTED_STRING_RE = re.compile(r"(['\"])((?:(?!\1).)*)\1")
 _JINJA_TOKEN_RE = re.compile(r"\$\{\{|\}\}")
 _V0_SELECTOR_RE = re.compile(r"#\s*\[")
 
+# A block-mapping `key: "..."` (or `key: '...'`) whose value opens a quote -
+# see _join_folded_quoted_scalars.
+_SCALAR_KEY_RE = re.compile(r"^([ \t]*[\w.-]+:[ \t]*)([\"'])")
+
+
+def _find_scalar_close(text: str, quote: str) -> int:
+    r"""Given a quote character, find where it closes in
+    ``text`` (-1 if it doesn't), respecting ``\"``
+    and ``''`` escapes.
+    """
+    i, n = 0, len(text)
+    while i < n:
+        ch = text[i]
+        if quote == '"' and ch == "\\":
+            i += 2
+            continue
+        if ch == quote:
+            if quote == "'" and i + 1 < n and text[i + 1] == "'":
+                i += 2
+                continue
+            return i + 1
+        i += 1
+    return -1
+
+
+def _join_folded_quoted_scalars(raw_meta_yaml: str) -> str:
+    """Join a quoted scalar value that YAML folds across multiple physical
+    lines back onto a single line.
+
+    YAML folds a line break inside a quoted scalar into a single space, so
+    this is a semantic no-op - but crm's recipe reader parses selectors line
+    by line, and a continuation line shaped like ``'word' more: words``
+    (quote-led, with a colon later in the line) gets misread as its own
+    key/value pair instead of plain scalar content, crashing with a raw
+    ``ParsingException`` instead of a clean warning (real example:
+    r-stringi's ``about/summary``, a long description wrapped across two
+    lines). Physically joining the fold before crm ever sees it removes the
+    ambiguous line shape entirely.
+    """
+    lines = raw_meta_yaml.split("\n")
+    out: list[str] = []
+    i = 0
+    while i < len(lines):
+        line = lines[i]
+        match = _SCALAR_KEY_RE.match(line)
+        if not match:
+            out.append(line)
+            i += 1
+            continue
+        quote = match.group(2)
+        if _find_scalar_close(line[match.end() :], quote) != -1:
+            out.append(line)
+            i += 1
+            continue
+
+        chunk = [line]
+        j = i + 1
+        closed = False
+        while j < len(lines):
+            chunk.append(lines[j])
+            closed = _find_scalar_close(lines[j].strip(), quote) != -1
+            j += 1
+            if closed:
+                break
+        if not closed:
+            # Genuinely unclosed/malformed - leave as-is and let crm raise
+            # its own error rather than silently swallowing lines.
+            out.append(line)
+            i += 1
+            continue
+        out.append(chunk[0] + "".join(" " + c.strip() for c in chunk[1:]))
+        i = j
+    return "\n".join(out)
+
 
 def _normalize_legacy_license(raw_meta_yaml: str, to_spdx) -> str:
     """Rewrite a recognized legacy license string to its SPDX identifier.
@@ -224,10 +298,11 @@ class GenericV0ToV1Migrator(MiniMigrator):
         not safe to ship unattended, along with the list of messages that
         explain why (empty if the conversion is clean).
         """
-        # Fix legacy license strings crm's own matcher won't recognize, and
-        # hide `environ["X"]` calls before crm's duplicate-key merge can
-        # mangle them - see the two functions' docstrings above.
+        # Join folded multi-line quoted scalars before crm's line-based reader can misparse a continuation line.
+        raw_meta_yaml = _join_folded_quoted_scalars(raw_meta_yaml)
+        # Fix legacy license strings crm's own matcher won't recognize.
         raw_meta_yaml = _normalize_legacy_license(raw_meta_yaml, self._to_spdx)
+        # Hide `environ["X"]` calls before crm's duplicate-key merge can mangle them.
         raw_meta_yaml = _hide_environ_calls(raw_meta_yaml)
 
         # Mirrors what the `crm convert` CLI always does otherwise (e.g.
--- a/conda_forge_tick/migrators/v0_to_v1.py
+++ b/conda_forge_tick/migrators/v0_to_v1.py
@@ -1,0 +1,86 @@
+import logging
+import typing
+from pathlib import Path
+from typing import Any
+
+from conda_recipe_manager.parser._message_table import MessageCategory, MessageTable
+from conda_recipe_manager.parser.exceptions import BaseParserException
+from conda_recipe_manager.parser.recipe_parser_convert import RecipeParserConvert
+
+from conda_forge_tick.migrators.core import MiniMigrator
+
+if typing.TYPE_CHECKING:
+    from ..migrators_types import AttrsTypedDict
+
+logger = logging.getLogger(__name__)
+
+
+class GenericV0ToV1Migrator(MiniMigrator):
+    """Convert a schema v0 ``meta.yaml`` recipe to a schema v1 ``recipe.yaml``.
+
+    This only knows about the ``conda-recipe-manager`` (``crm``) conversion
+    mechanics: it converts the recipe and skips (with a warning, leaving the
+    recipe untouched) any conversion that comes back with actionable
+    messages rather than shipping something broken. Subclasses layer
+    package-specific pre/post-processing - and extend ``IGNORED_WARNINGS`` -
+    on top of this class; nothing package-specific lives here.
+    """
+
+    allowed_schema_versions = [0]
+
+    # Warnings that are safe to ignore: crm emits them but the produced
+    # recipe.yaml is already correct. Extend in subclasses, don't rewrite the
+    # safe-to-auto-convert check itself.
+    IGNORED_WARNINGS: tuple[str, ...] = (
+        # crm already removes the deprecated field, the warning is just noise.
+        "Field at `/about/license_family` is no longer supported.",
+    )
+
+    def _actionable_messages(self, msg_tbl: MessageTable) -> list[str]:
+        blocking = list(msg_tbl.get_messages(MessageCategory.EXCEPTION))
+        blocking += msg_tbl.get_messages(MessageCategory.ERROR)
+        blocking += [
+            msg
+            for msg in msg_tbl.get_messages(MessageCategory.WARNING)
+            if not any(ignore in msg for ignore in self.IGNORED_WARNINGS)
+        ]
+        return blocking
+
+    def _convert(self, raw_meta_yaml: str) -> tuple[str | None, list[str]]:
+        """Attempt to convert ``raw_meta_yaml`` (schema v0) to a v1 recipe.
+
+        Returns the converted recipe text, or ``None`` if the conversion is
+        not safe to ship unattended, along with the list of messages that
+        explain why (empty if the conversion is clean).
+        """
+        # crm can *raise* on malformed recipes (ParsingException,
+        # DuplicateKeyException, ...) at construction, before any msg_tbl
+        # exists to record the failure.
+        try:
+            converter = RecipeParserConvert(raw_meta_yaml)
+            v1_content, msg_tbl, _debug = converter.render_to_v1_recipe_format()
+        except BaseParserException as exc:
+            return None, [f"crm raised {type(exc).__name__}: {exc}"]
+
+        blocking = self._actionable_messages(msg_tbl)
+        if blocking:
+            return None, blocking
+        return v1_content, []
+
+    def migrate(self, recipe_dir: str, attrs: "AttrsTypedDict", **kwargs: Any) -> None:
+        meta_yaml_path = Path(recipe_dir) / "meta.yaml"
+        if not meta_yaml_path.exists():
+            return
+
+        v1_content, blocking = self._convert(meta_yaml_path.read_text())
+        if blocking:
+            logger.warning(
+                "Skipping v0 -> v1 conversion for %s: not safe to auto-convert:\n%s",
+                attrs.get("name", recipe_dir),
+                "\n".join(blocking),
+            )
+            return
+
+        assert v1_content is not None
+        (Path(recipe_dir) / "recipe.yaml").write_text(v1_content)
+        meta_yaml_path.unlink()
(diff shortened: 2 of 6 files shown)


conda_forge_tick/migrators/v0_to_v1.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+    def _actionable_messages(self, msg_tbl: MessageTable) -> list[str]:
+        blocking = list(msg_tbl.get_messages(MessageCategory.EXCEPTION))
+        blocking += msg_tbl.get_messages(MessageCategory.ERROR)
+        blocking += [
+            msg
+            for msg in msg_tbl.get_messages(MessageCategory.WARNING)
+            if not any(ignore in msg for ignore in self.IGNORED_WARNINGS)
+        ]
```
jaimergp:
```suggestion
        blocking = [
            *msg_tbl.get_messages(MessageCategory.EXCEPTION),
            *msg_tbl.get_messages(MessageCategory.ERROR),
        ]
        for msg in msg_tbl.get_messages(MessageCategory.WARNING):
            if not any(ignore in msg for ignore in self.IGNORED_WARNINGS):
                blocking.append(msg)
```

I find this cleaner, but feel free to reject or implement partially

conda_forge_tick/migrators/v0_to_v1.py
```diff
@@ -0,0 +1,96 @@
...
+        # crm can raise - or, on some malformed/unusual recipes (e.g.
+        # multi-output), crash with an unrelated exception - at construction
+        # or during conversion, before any msg_tbl exists to record why. Any
+        # of that means this recipe isn't safe to auto-convert.
+        try:
+            converter = RecipeParserConvert(
+                content, flags=RecipeReaderFlags.ALLOW_DUPLICATE_KEYS
+            )
+            v1_content, msg_tbl, _debug = converter.render_to_v1_recipe_format()
+        except Exception as exc:
```
jaimergp, line 424:
```suggestion
        except Exception as exc:
            logger.debug("CRM failed to convert", exc_info=exc)
```

To have more info to debug if needed.

conda_forge_tick/migrators/v0_to_v1.py
```diff
@@ -0,0 +1,96 @@
...
+            return None, blocking
+        return v1_content, []
+
+    def migrate(self, recipe_dir: str, attrs: "AttrsTypedDict", **kwargs: Any) -> None:
+        meta_yaml_path = Path(recipe_dir) / "meta.yaml"
+        if not meta_yaml_path.exists():
+            return
+
+        v1_content, blocking = self._convert(meta_yaml_path.read_text())
+        if blocking:
```
jaimergp, line 448: Don't we need to flag this somewhere as an error in the bot or something? This may look like a successful migration otherwise 🤔

conda_forge_tick/migrators/r_v0_to_v1.py
```diff
@@ -0,0 +1,41 @@
+import typing
+
+from conda_forge_tick.migrators.v0_to_v1 import GenericV0ToV1Migrator
+
+if typing.TYPE_CHECKING:
+    from ..migrators_types import AttrsTypedDict
+
+
+def _is_r_feedstock(attrs: "AttrsTypedDict") -> bool:
```
jaimergp, line 9: I think all R feedstocks are maintained by the same team (`conda-forge/r`) too, so that may be a way to detect them?

https://github.com/conda-forge/r-essentials-feedstock/blob/746ea67ca95e4a6154224e58d4f346c7cebe710b/recipe/meta.yaml#L69-L71

conda_forge_tick/migrators/v0_to_v1.py
pb01ka: I don't think so. It's because the `v1_content` is never written into `recipe.yaml` for this case. The warning being logged is `Skipping v0 -> v1 conversion...` and the control returns. So migration would never actualise. 

May be I am missing something? Do you want to elevate this to an error instead of a warning? Please let me know.

jaimergp: Who's going to see the warning? I'm expecting something we can track in its status page (e.g. like https://conda-forge.org/status/migration/?name=supportwindowsarm64platform), so unless it's an error, it will be a skipped feedstock (or even worse, a "successfully migrated" one in the eyes of the bot metadata?). Need to verify that.

pb01ka: I see. It makes sense. Here are a couple (exactly two 😆 TBH) of options I would like to propose,

**A. Promote it to a real top-level Migrator.** i.e., give it a name, build it as a `GraphMigrator` (AFAICT, it's currently not) with its own `effective_graph`/`pr_limit`, and register it in `make_migrators.py` like `ArchRebuild`/`OSXArm`/`MigrationYaml`. I think this would be the only option matching the exact status-page precedent you shared.

**B. Write a per-node status field via a new MiniMigrator-scoped mechanism.** AFAIK, `pr_info.pre_pr_migrator_status` only covers top-level Migrator exceptions. `RV0ToV1Migrator.migrate()` would write to `attrs["pr_info"]`, and `status_report.py` would need extending to read and report it.

I would recommend option A - I guess it is what we intend to do anyways (?). Please let me know TY.

jaimergp: Ah, true, why is this a minimigrator? 🤔 I had missed that part at the beginning assuming we wanted a full graph migrator.

pb01ka: Well initially I had two options - full graph migrator and mini-migrator. The latter felt better to me for the first draft. However, since now you agree with having a full graph migrator, I can do that. Shouldn't take much time.

isuruf: I think a mini-migrator makes sense. We can piggy back off r 4.6 migrator

isuruf: If the v0->v1 conversion fails, then the PR for r_base 4.6 will still be made right? Then at the end of the migration, we can retry for the handful left.

pb01ka: Thanks @isuruf I think your comment is quite helpful in converging on a better plan than the standalone-migrator approach I'd have gone with earlier. Here's where I've landed:

### 1. Mini-migrator, piggy-backed on the r-base 4.6 migration

Reverting `RV0ToV1Migrator` from a standalone `GraphMigrator` approach back to a `MiniMigrator` approach, and scoping it to the r-base 4.6 rebuild migration specifically. I guess this follows the exact precedent already in this codebase for `RUCRTCleanup`:

```python
if migration_name.startswith("r-base44"):
    piggy_back_migrations.append(RUCRTCleanup())
```

The equivalent hook might look like `if migration_name.startswith("r-base46"): piggy_back_migrations.append(RV0ToV1Migrator())`. Since bumping the r-base pin forces a rebuild of essentially every R feedstock, this gets us close to full coverage of the R feedstock population in one campaign. **Importantly, it also combines the r-base 4.6 rebuild and the recipe format conversion into a single PR per feedstock, instead of two separate ones.**

**One concern worth flagging explicitly**: AFAICT, the r-base 4.6 migration doesn't exist yet. I checked that there's no reference to it anywhere in this code base.
Assuming that the r-base 4.6 migration will be created in the near future, and since the `make_migrators.py` hook is a one-line addition, the `RV0ToV1Migrator` mini-migrator can be built and merged independently of that, and then piggy-backed onto the r-base 4.6 migration once it exists.
If it already exists, then please point me to it and I'll add the piggy-back hook immediately.

**a v0 -> v1 failure shouldn't block the r-base 4.6 PR.** I think, as long as `RV0ToV1Migrator.migrate()` doesn't raise on a skip (it doesn't, it just logs a warning and returns gracefully), the containing r-base 4.6 migrator's PR gets created normally regardless of whether the conversion succeeded, failed, or was skipped for that feedstock.

### 2. Track failures so the *handful left* can actually be retried

The catch with (1) alone: AFAIK, a mini-migrator's `migrate()` return value is discarded by the caller (and also warnings are non-persistent), so no record of which feedstocks were skipped is kept. To retry for the skipped/failed recipe conversions, at the end of the migration wave, we need to have such a record.

So I am thinking of adding a lightweight tracking mechanism (as described in **Option B** above): when `RV0ToV1Migrator.migrate()` skips a feedstock, write a small status note into that node's persisted `pr_info` (e.g. `skipped: True` + the blocking reason(s)) instead of only logging. It should be just enough data to query "which R feedstocks are still on schema v0, and why" after the migration wave finishes, and drive a manual or scripted retry pass over just those.

Please correct me wherever I'm wrong. TY.

jaimergp: Ok Isuru's approach seems sensible, let's do that. I don't think we need to track which ones were skipped. Instead, the 2nd sweep will traverse the full R graph and skip any feedstock that is already `recipe.yaml`.

pb01ka: I see.

> I don't think we need to track which ones were skipped. Instead, the 2nd sweep will traverse the full R graph and skip any feedstock that is already `recipe.yaml`.

It would be slightly wasteful to revisit the recipes (going through all the feedstocks and checking if `recipe.yaml` already exist for each of those) which are already successfully converted. However, I don't mind doing so.

**Final decision** - I will keep the `RV0ToV1Migrator` as a mini-migrator for now. I wouldn't go for step 2 (**Track failures so the *handful left* can actually be retried**) in https://github.com/conda-forge/conda-forge-bot/pull/6381#discussion_r3716045656. It's anyways an independent of the logical implementation of the migrator. So, if the need arises in the future, we can add tracking mechanism later on.

pb01ka:
> I am making this comment just for record purposes. No action necessary (but thoughts/opinions welcome).

With https://github.com/conda-forge/conda-forge-bot/pull/6381/commits/a0e437cb6d72514258d111969386ac54a82ca5b5, `r-kedd` was failing with `Could not patch unrecognized license: GPL (>= 2)`. The conversion was failing. 

However, with https://github.com/conda-forge/conda-forge-bot/pull/6381/commits/08ab378f795f62f2a85b46ddb20a9ff3dcff114a, conversion happened. Here are the relevant files,

**Original `meta.yaml`**

<details>

<summary>Click here</summary>

```yaml
{% set version = "1.0.4" %}

{% set posix = 'm2-' if win else '' %}

package:
  name: r-kedd
  version: {{ version|replace("-", "_") }}

source:
  fn: kedd_{{ version }}.tar.gz
  url:
    - https://cran.r-project.org/src/contrib/kedd_{{ version }}.tar.gz
    - https://cran.r-project.org/src/contrib/Archive/kedd/kedd_{{ version }}.tar.gz
  sha256: bb944389a61cff22bb3bdde938a0ff6c16700a1ca725ef36b391b97812c45b26

build:
  noarch: generic
  number: 2
  rpaths:
    - lib/R/lib/
    - lib/

requirements:
  build:
    - r-base

  run:
    - r-base

test:
  commands:
    - $R -e "library('kedd')"  # [not win]
    - "\"%R%\" -e \"library('kedd')\""  # [win]

about:
  home: https://CRAN.R-project.org/package=kedd
  license: GPL (>= 2)
  summary: Smoothing techniques and computing bandwidth selectors of the nth derivative of a probability density for one-dimensional data.
  license_family: GPL3
  license_file: '{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3'  # [unix]
  license_file: '{{ environ["PREFIX"] }}\R\share\licenses\GPL-3'  # [win]

extra:
  recipe-maintainers:
    - conda-forge/r
```

</details>

---

**Converted `recipe.yaml`**

<details>

<summary>Click here</summary>

```yaml
schema_version: 1

context:
  version: "1.0.4"
  posix: "'m2-' if win else ''"

package:
  name: r-kedd
  version: ${{ version|replace("-", "_") }}

source:
  url:
    - https://cran.r-project.org/src/contrib/kedd_${{ version }}.tar.gz
    - https://cran.r-project.org/src/contrib/Archive/kedd/kedd_${{ version }}.tar.gz
  sha256: bb944389a61cff22bb3bdde938a0ff6c16700a1ca725ef36b391b97812c45b26
  file_name: kedd_${{ version }}.tar.gz

build:
  number: 2
  noarch: generic
  dynamic_linking:
    rpaths:
      - lib/R/lib/
      - lib/

requirements:
  build:
    - r-base
  run:
    - r-base

tests:
  - script:
      - if: not win
        then: $R -e "library('kedd')"
      - if: win
        then: "\"%R%\" -e \"library('kedd')\""

about:
  license: GPL-2.0-or-later
  summary: Smoothing techniques and computing bandwidth selectors of the nth derivative of a probability density for one-dimensional data.
  license_file: ${{ (env.get("PREFIX") ~ '/lib/R/share/licenses/GPL-3') if unix else (env.get("PREFIX") ~ '\\R\\share\\licenses\\GPL-3') if win else '' }}
  homepage: https://CRAN.R-project.org/package=kedd

extra:
  recipe-maintainers:
    - conda-forge/r
```

</details>

pb01ka:
The steps remaining for this PR to be marked as "Ready for review",

- Fix the parsing error encountered `r-stringi`'s v0 to v1 conversion - Out of all the packages in [`r_v1_sample_feedstocks.csv`](https://github.com/pb01ka/r-feedstock-v1-feasibility/blob/6d9bad37ced51487029050cce429a30e31203e3b/r_v1_sample_feedstocks.csv) its the only one left to be addressed by the `RV0ToV1Migrator`. 
- Test the `RV0ToV1Migrator` on a larger sample - ~300 R feedstocks.

danielnachun:
I have a couple thoughts on this having worked quite a bit on the last two migrations.

First is handling of patches - although `rattler-build` has improved a lot in this regard, I still occasionally run into failures because an existing patch is formatted in a way that is incompatible with the patching library `rattler-build` uses.  I think this will probably be something that just has to be handled by maintainers manually - Claude is quite good now at regenerating patches to work with what `rattler-build` wants.  But if there is an automated way to fix patches to work with rattler-build that would save some work.

Second is handling of build scripts - we'll definitely have to regenerate build scripts (or better yet, move the build steps in to the recipe itself).  Existing build scripts reference `${R}` which is now undefined in `rattler-build` (we could change this but I'm unsure we should).  We do rarely have build scripts that have to set environment variables (usually for detecting shared libraries or setting compilation flags).  This could be slightly tricky from an automation perspective because sometimes the package might build correctly even without the flags but will incorrectly use a bundled library, for example, instead of the conda-forge one.

At the same time we still have some packages with old build scripts which build the package like normal for all platforms except `osx-arm64`, where the script instead tries to copy files from a now non-existent prebuilt package.  I assume this is a leftover from some sort of short term work around to support `osx-arm64` before we had cross-compilation working.

So the question is how much of this the migration bot should try to clean up itself versus flagging the issue for manual intervention?

pb01ka:
TYSM @danielnachun for highlighting these important points.

> So the question is how much of this the migration bot should try to clean up itself versus flagging the issue for manual intervention?

I will complete testing my migrator on a larger sample of R-feedstocks. Subsequently I will use it to figure out answers to your questions (the two points you have highlighted and how much the bot will be able to do on its own) by tomorrow EOD. I hope that will work.

TY again. <3.

danielnachun:
That's great.  If your larger test doesn't already hit some of these snags, I can definitely at least find some recipes with more complicated `build.sh` scripts as a smoke test for the bot.  It's harder to anticipate patches failing.

One other thing I realized now too that can come up - in every R migration we always run into at least a handful of CRAN package which have been archived and are therefore unmaintained (see https://cran.r-project.org/web/packages/RandomFields/index.html for example).  Some fraction of those packages will fail to build, and we end up archiving those feedstocks.  

Potentially it would be worth checking the package page to see if it's archived and instead of migrating it, have the bot open an issue indicating the feedstock should be archived.

pb01ka:
**Two questions: recipes with a `license` string `crm` can't patch**

While testing `RV0ToV1Migrator` against a [sample of R feedstocks](https://github.com/pb01ka/r-feedstock-v1-feasibility/blob/main/r_v1_sample_feedstocks.csv), a handful failed to convert because `crm` couldn't map their `license` field to an SPDX identifier or due to duplicate key in `meta.yaml`.

**1. Ambiguous/custom identifiers with no deterministic SPDX equivalent**
- [`r-stringi`: `license: FOSS`](https://github.com/conda-forge/r-stringi-feedstock/blob/cd812ecb08e6da8773029cfe5b501ce629f63929/recipe/meta.yaml#L53)
- [`r-rttf2pt1`: `license: LicenseRef-BSDLike`](https://github.com/conda-forge/r-rttf2pt1-feedstock/blob/53aa6e8a20da03a9a7caee966bb95459a4eb7eef/recipe/meta.yaml#L48)

AFAICT, `FOSS` is most probably a shorthand for "this package is open source," without saying which license. `LicenseRef-BSDLike` at least hints at a family ("BSD-like"), but not a specific variant. Neither maps deterministically to a single SPDX identifier the way `GPL-2` -> `GPL-2.0-only` or `LicenseRef-Unlimited` -> `Unlimited` does, so there isn't a safe automatic mapping to add here.

A few ways we could handle this, roughly in increasing order of acceptability:

1. **Skip the conversion (current behavior).** Safe, but means these recipes need a manual v0 -> v1 conversion later.
2. **Let it through verbatim.** `crm` already leaves the field unchanged when it can't map it, so this is a one-line change (add the warning to `IGNORED_WARNINGS`). A small concern: the v1 recipe schema expects a real SPDX expression in `about.license`, and a vague/custom string like `FOSS` isn't one. This would likely just move the failure from "bot skips the recipe" to "bot opens a PR that fails conda-forge's own recipe-lint/schema validation."
3. **Try to infer the real license from the package's own source.** For `r-stringi`, the actual license lives in the upstream package's `LICENSE` file (e.g. https://github.com/gagolews/stringi/blob/master/LICENSE), which CRAN's `DESCRIPTION`/`meta.yaml` doesn't surface. We could fetch and inspect that file (or CRAN's package page) to guess the SPDX identifier automatically. Worth mentioning as an option but I feel it would be a long shot to implement and maintain, and would still be error-prone (e.g. if the license changes upstream).

Would wait for your thoughts on this.

---

**2. Unselector-ed duplicate top-level sections (e.g. `r-loose.rock`)**

Separately, [`r-loose.rock`](https://github.com/conda-forge/r-loose.rock-feedstock/blob/762d37de124d56b1106521fd0433d124dde6734b/recipe/meta.yaml#L14-L25) failed with a raw `DuplicateKeyException` from `crm` rather than a clean warning. Its `meta.yaml` has two full, unconditional `build:` sections (no `# [selector]` on either) - one declares `noarch: generic`, the other doesn't. That's not the common `key: val  # [selector]` idiom `crm`'s duplicate-key merge handles.
It's most likely a copy-paste mistake in the source recipe, and worth flagging on its own since (I think) plain YAML semantics silently pick one of the two `build:` blocks - meaning `noarch: generic` may already be getting silently dropped in the currently-published package, independent of this migrator.

`RV0ToV1Migrator` (via `GenericV0ToV1Migrator`) now detects a genuinely duplicated, unselector-ed top-level section before `crm` even runs, and reports it clearly:

```
top-level `build:` section appears 2 times with no selector to distinguish them - likely an authoring mistake in the source recipe that needs manual review
```

The recipe is still skipped either way - this only replaces the internal `DuplicateKeyException` with something a maintainer can act on directly. See the exact logs [here](https://github.com/pb01ka/r-feedstock-v1-feasibility/blob/bfef3d403644fedfda39ecf51806f569008d3a6a/v0-to-v1-out/r-loose.rock/v0-to-v1.log).

---

Other than these two questions, from my end I am marking this PR as ready for review.

Now I will work on answering @danielnachun's questions.

pb01ka:
https://github.com/conda-forge/conda-forge-bot/actions/runs/31217780965/job/92995304151?pr=6381 seems to be unrelated failure. A simple re-run should make it pass I guess.

pb01ka:
> If your larger test doesn't already hit some of these snags, I can definitely at least find some recipes with more complicated `build.sh` scripts as a smoke test for the bot. It's harder to anticipate patches failing.

@danielnachun I found out the following sample for the two points you shared. I will experiment using these sample and get back to you with a concrete well-tested response. If you have any other feedstocks worth experimenting with, please share with me, I will include it in my sample.

```
r-msqc
r-pca3d
r-cffr
r-juniperkernel
r-systemfonts
r-rcppdpr
r-geodist
r-leadercluster
r-xtensor
```

pb01ka:
@danielnachun As promised, here are my thoughts on [your comment above](https://github.com/conda-forge/conda-forge-bot/pull/6381#issuecomment-5208894861).

**1. Patches**

I think this can stay out of scope for the migrator (please let me know if you think otherwise).
AFAICT, we only rewrite the recipe YAML's structure and never the files a `patches:` list points at. Also, a patch incompatible with `rattler-build`'s patching library will convert "successfully" and only fail later, at build time.

I also checked how often this actually comes up and AFAICT, only **1** of our sample of ~350 feedstocks - [`r-websocket`](https://github.com/conda-forge/r-websocket-feedstock/blob/main/recipe/meta.yaml#L13-L14) - declares a `patches:` list. That being said, I would still propose a 3-step pipeline (if the problem is way more frequent than I predict using my sample)
- try the patch as-is under `rattler-build`; only on failure,
    - apply it with plain CLI tools against the unpacked source;
    - then re-derive a `rattler-build`-compatible patch from the result

---

**2. Build scripts**

Same as patches, `build.sh` content passes through completely untouched, by both `crm` and our migrator.
As expected, almost every R feedstock  has a `build.sh`. So the useful question here is which *recurring* pattern inside these scripts are worth flagging (and which of those can be automatically fixed by the migrator).
I found the `osx-arm64`/`osx-64` prebuilt-copy related workaround you described in almost third of my sample. Specifically, in the feedstocks I checked (e.g. `r-pca3d`, `r-msqc`), the outer `if` already routes `osx-64` to a normal install, so the `install_name_tool` dylib-path fix - nested inside `target_platform == osx-64` in the `else` branch `osx-arm64` falls into - is unreachable dead code. May be, that block was added in before osx-arm64 support existed, so it's very likely never run. Therefore, I flag it for manual review instead of trying to fix it. In [this commit](https://github.com/conda-forge/conda-forge-bot/pull/6381/commits/978a5ac16368365b80158123781204724f611647), I implemented it as a non-blocking, post-conversion warning (see below),

```
build.sh's `install_name_tool` dylib-path fixes for osx-64 are commonly nested inside the branch that already excludes osx-64 (the one osx-arm64 falls into) - making them dead code that's likely never run. Verify whether that's the case here, and if so, whether osx-arm64 needs the missing fix.
```

I think we can extend this going forward: catalog and add detectors for concrete, recurring patterns as we find them - same as `osx-arm64` - rather than guessing at build-script complexity up front.

Please let me know your opinions on both of these. Please also note [my comment above](https://github.com/conda-forge/conda-forge-bot/pull/6381#issuecomment-5222294371). TY.

danielnachun:
For 1, I think it's fine not to try to fix patches - at the very least Claude Code is very good at regenerating the patches so it's not a huge maintenance burden.

For 2, the pattern of checking the platform in scripts like this should be removed entirely - the `install_name_tool` fix has been irrelevant for a long time as well.  The only issue with outright replacing `build.sh` and `bld.bat` with an in recipe call to `R CMD INSTALL --build . ${R_ARGS}` is just that we do occasionally have to set environment variables or make patches with `sed`.

I think what I should do is just suffer through cloning the conda-forge/feedstocks repo and then having Claude on my machine locally comb through all the build.sh scripts for the R feedstocks to figure out which feedstocks even have this issue and what kind of patterns are there.

The cleanest ideal outcome for this whole thing is that are able to move the vast majority of recipes to have `R CMD INSTALL --build . ${R_ARGS}` in the recipe file itself, with environment variables being set using the recipe mechanisms instead of `export`/`set`.  I'd love to see the R feedstocks get as close to declarative as possible, thought that may take multiple iterations to fully achieve.

Other thing that was missed is that I thought `${R}`/`%R%` is undefined in `rattler-build` (at least it is when I build locally).  I would think that would require editing every `build.sh`/`bld.bat` to fix that?

pb01ka:
@danielnachun Here are the proposed next steps I am planning to take for `build.sh`/`bld.bat` handling, based on the discussion above (open to suggestions):

**1. Auto-replace pure-boilerplate scripts.** When a `build.sh`/`bld.bat` is nothing but the known dead platform-branch + `install_name_tool`/`mv` template (plus the handful of repeated boilerplate lines like `DISABLE_AUTOBREW`, `LD_LIBRARY_PATH`), we can delete it and emit an inline `build.script` instead:

```yaml
build:
  script:
    - if: win
      then: R CMD INSTALL --build . %R_ARGS%
      else: R CMD INSTALL --build . ${R_ARGS}
```

This should match rattler-build's own `generate-recipe cran` tool output exactly - see [`cran.rs`](https://github.com/prefix-dev/rattler-build/blob/3b0edc315efac949088f21c373c2182a7b66fb5e/crates/rattler_build_recipe_generator/src/cran.rs#L324-L329) and the [`r-knitr` docs example](https://github.com/prefix-dev/rattler-build/blob/3b0edc315efac949088f21c373c2182a7b66fb5e/docs/snippets/recipes/r-knitr.yaml#L14). Both use plain `R`, not `${R}`/`%R%`. Therefore I think this should discharge the `${R}` question entirely (please let me know if you think otherwise).

---

**2. Anything else: leave the script's content alone, but still rename `bld.bat` -> `build.bat`.** rattler-build only auto-discovers `build.bat` on Windows now:

> We still support the `build.sh` script, but the `bld.bat` script was renamed to `build.bat` in order to be more consistent with the `build.sh` script.

Reference - [`docs/converting_from_conda_build.md`](https://github.com/prefix-dev/rattler-build/blob/3b0edc315efac949088f21c373c2182a7b66fb5e/docs/converting_from_conda_build.md?plain=1#L105-L106)

AFAICT, almost all feedstocks in my sample ships a `bld.bat`. Without this rename, essentially every converted R feedstock silently loses its Windows build script. The rename seems to be safe (a pure filename change, no semantic risk), so it happens for every recipe that keeps a real script.

---

**3. Boilerplate detection stays conservative.** "Pure boilerplate" means an exact match against the known dead template. Any unexpected line like a `sed` patch, a custom `export`, or anything beyond the boilerplate lines already catalogued falls through to "leave the script alone, flag for manual review," same as today. In fact, if I'm not wrong - as per your suggestion, we can use Claude to figure out such boilerplate detection patterns.

beckermr (changes requested):
Much of the extra code here is workarounds for bugs in CRM itself. Fixes to CRM need to be made upstream along with better testing. 

Also, we need to understand how much jinja2 and selector text CRM is calling via `eval` on or similar. My inclination is to do this conversion in a container to isolate potentially unsafe things.

pb01ka:
@beckermr TYSM for your review. Indeed its true that much of the post and pre-processing code here is to workaround bugs in CRM. 

I will open a tracking issue in the CRM repository in reference to your review. I will also attempt to fix these bugs. 

TY again. ❤️

pb01ka:
@beckermr I have created the issue at - https://github.com/conda/conda-recipe-manager/issues/561. I will wait for comments on it before I send PR(s) to CRM repository.

beckermr:
I've upstreamed my own fix into crm but none of them address the issues you've found.

PR #6440, opened 2026-08-07
Fix node filtering for arch migrators

isuruf:
#### Description:

An alternative to https://github.com/conda-forge/conda-forge-bot/pull/6439

#### Checklist:

### DIFF
--- a/conda_forge_tick/migrators/core.py
+++ b/conda_forge_tick/migrators/core.py
@@ -925,7 +925,7 @@ class GraphMigrator(Migrator):
             ):
                 continue
 
-            if self.filter_node_migrated(payload):
+            if super().filter_node_migrated(payload):
                 continue
 
             muid = frozen_to_json_friendly(self.migrator_uid(payload))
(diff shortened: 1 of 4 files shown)


conda_forge_tick/migrators/core.py
```diff
@@ -896,6 +896,8 @@ def all_predecessors_issued(self, attrs: "AttrsTypedDict") -> bool:
             ):
                 continue
 
+            if self.filter_node_migrated(payload):
```
beckermr:
```suggestion
            if super().filter_node_migrated(payload):
```

conda_forge_tick/migrators/core.py
```diff
@@ -923,6 +925,9 @@ def predecessors_not_yet_built(self, attrs: "AttrsTypedDict") -> bool:
...
```
beckermr:
```suggestion
            if super().filter_node_migrated(payload):
```

h-vetinari (commented):
Admittedly not very familiar with all the components here, but I do prefer this PR over #6439.

Key is that this function already exists in the core infrastructure, so using this also for arch migrators seems consistent
https://github.com/conda-forge/conda-forge-bot/blob/7c8d6cc052eee97084b0eabe09d809dcc08cc549/conda_forge_tick/migrators/core.py#L506-L510

PR #6569, opened 2026-08-26
exclude r-languageserver in riscv64 too

isuruf:
#### Checklist:

### DIFF
--- a/conda_forge_tick/migrators/arch.py
+++ b/conda_forge_tick/migrators/arch.py
@@ -505,6 +505,7 @@ class LinuxRISCV64(_CrossCompileRebuild):
         "intel-compiler-repack",
         "intel_repack",
     }
+    excluded_dependencies = {"r-languageserver"}
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


conda_forge_tick/migrators/arch.py
```diff
@@ -505,6 +505,7 @@ class LinuxRISCV64(_CrossCompileRebuild):
         "intel-compiler-repack",
         "intel_repack",
     }
+    excluded_dependencies = {"r-languageserver"}
```
h-vetinari, line 508: I think this ended up removing `conda-forge-ci-setup` from the migrator? 🤔 

https://github.com/conda-forge/conda-forge-bot/blob/2bef536c08b7846e5362ddda40f5d304eda65180/conda_forge_tick/migrators/arch.py#L33-L37

Was that intentional? Or just the price of getting all the R-stuff out of the migrator?

isuruf: No, see https://github.com/conda-forge/conda-forge-bot/pull/6570 for a better solution

PR #6674, opened 2026-09-08
Prune blas implementations in arch migrations

h-vetinari:
When looking at the arch migrator pages, e.g. the one for [riscv64](https://conda-forge.org/status/migration/?name=supportlinuxriscv64platform), there's a huge amount of packages that get pulled in through the blas metapackage, and all its many implementations. We don't need all of those to build `blas`, just `lapack` and `openblas` suffice. But the bot can't know that, so provide a function for the graph that prunes such unwanted dependencies (while still keeping those that are needed elsewhere). 

As a concrete example, all the following dependencies lead to `blas`, and none of them are truly necessary for the riscv64 migration: 

image

.
PS. I used some LLM to simplify the graph traversal and for writing tests, but idea/implementation/code/comments are all written by myself.
PPS. The existing `pluck` cannot be used for this purpose, because it explicitly keeps the ancestors and fuses the edges around the removed node.
PPPS. This function is obviously written with broader applicability in mind. We can later also apply it to other tricky meta-packages where necessary or beneficial.

### DIFF
--- a/conda_forge_tick/migrators/arch.py
+++ b/conda_forge_tick/migrators/arch.py
@@ -137,10 +137,8 @@ def _filter_stubby_and_ignored_nodes(graph, outputs_lut, ignored_packages):
     # special handling for meta-packages
     if "blas" in graph.nodes("payload"):
         # blas has many implementations, and its feedstock depends on all of them;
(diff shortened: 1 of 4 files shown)


beckermr (approved):
LGTM and thank you!

PR #6680, opened 2026-09-09
Treat already-migrated predecessors as built in arch migrations

velonica0:
## Problem

`GraphMigrator.predecessors_not_yet_built` decides whether a parent is built by looking only at `pr_info.PRed`. A feedstock migrated by hand or by a rerender never gets a `PRed` entry, and nothing can add one after the fact — so such a parent blocks all of its children permanently.

This is live on the riscv64 migration: `blas` and `lapack` both have `build_platform: {linux_riscv64: linux_64}` and ship `linux-riscv64` artifacts, but have no `LinuxRISCV64` record in `PRed`. That stalls `gsl` → `r-base` → the 29 `r-*` packages, i.e. 30 of the 43 feedstocks currently shown as "awaiting parents" on the [riscv64 status page](https://conda-forge.org/status/migration/?name=supportlinuxriscv64platform).

The bot already knows those parents are migrated — `filter_node_migrated` reads it off `conda-forge.yml`, which is why the status page lists them as done. The two checks just disagreed about what "built" means.

## Change

- New `GraphMigrator.predecessor_already_migrated()` hook, defaulting to `False`, consulted before the `PRed` lookup. Non-arch migrators are unaffected.
- `ArchRebuild` and `_CrossCompileRebuild` override it with the existing `conda-forge.yml` provider/`build_platform` check, extracted to a shared `_arches_are_configured()` helper (it was duplicated verbatim in both classes).

The hook deliberately ignores `PRed` records rather than reusing `filter_node_migrated`, which also returns true merely because a record exists — reusing it would let children race ahead of a parent whose PR is still open, for every graph migrator. `test_predecessor_with_open_pr_still_blocks` covers that.

Fixes aarch64, osx-arm64 and win-arm64 the same way.

## Tests

`tests/test_graph_migrator_predecessors.py` — five cases on a `lapack → gsl` graph: migrated via `build_platform`, migrated via `provider`, unmigrated parent still blocks, open PR still blocks, merged PR counts as built.

cc @conda-forge/help-riscv64

### DIFF
--- a/conda_forge_tick/migrators/arch.py
+++ b/conda_forge_tick/migrators/arch.py
@@ -22,6 +22,7 @@ from conda_forge_tick.migrators_types import AttrsTypedDict, MigrationUidTypedDi
 from conda_forge_tick.os_utils import pushd
 from conda_forge_tick.utils import (
     as_iterable,
+    get_keys_default,
     pluck,
     prune,
     yaml_safe_dump,
@@ -150,23 +151,23 @@ def _arches_are_configured(attrs: "AttrsTypedDict", arches: dict) -> bool:
     This reads the state off ``conda-forge.yml``, so it is true for feedstocks that
     were migrated by hand or by a rerender as well as for ones the bot PRed.
 
-    The default is false: an empty ``arches`` means there is nothing to detect, not
-    that the feedstock is configured for everything.
+    Any missing arch fails the check.
     """
-    configured = False
-    for arch in arches:
-        configured_arch = (
-            attrs.get("conda-forge.yml", {}).get("provider", {}).get(arch)
+
+    def _already_has_arch(arch):
+        # This arch has to be in provider or build_platform
+        return get_keys_default(
+            attrs, ["conda-forge.yml", "provider", arch], {}, None
         ) or (
-            attrs.get("conda-forge.yml", {}).get("build_platform", {}).get(arch)
+            get_keys_default(
+                attrs, ["conda-forge.yml", "build_platform", arch], {}, None
+            )
             not in [None, arch]
         )
-        if not configured_arch:
-            # This arch is not in provider or build_platform
-            return False
-        configured = True
 
-    return configured
+    # ternary condition because `all` over an empty list is always true; we want
+    # to default to False rather than marking a feedstock as done
+    return all(_already_has_arch(arch) for arch in arches) if arches else False
 
 
 class ArchRebuild(GraphMigrator):
--- a/conda_forge_tick/migrators/arch.py
+++ b/conda_forge_tick/migrators/arch.py
@@ -170,7 +170,27 @@ def _arches_are_configured(attrs: "AttrsTypedDict", arches: dict) -> bool:
     return all(_already_has_arch(arch) for arch in arches) if arches else False
 
 
-class ArchRebuild(GraphMigrator):
+class _ArchesConfiguredMixin:
+    """Answers "is this feedstock already building our arches?" off ``conda-forge.yml``.
+
+    Shared by the arch migrators, whose migrated state is visible in the feedstock
+    itself rather than only in the bot's PR records.
+    """
+
+    arches: dict = {}
+
+    def filter_node_migrated(
+        self, attrs: "AttrsTypedDict", not_bad_str_start: str = ""
+    ):
+        return _arches_are_configured(
+            attrs, self.arches
+        ) or super().filter_node_migrated(attrs, not_bad_str_start)  # type: ignore[misc]
+
+    def predecessor_already_migrated(self, attrs: "AttrsTypedDict") -> bool:
+        return _arches_are_configured(attrs, self.arches)
+
+
+class ArchRebuild(_ArchesConfiguredMixin, GraphMigrator):
     """A Migrator that adds aarch64 builds to feedstocks."""
 
     allowed_schema_versions = {0, 1}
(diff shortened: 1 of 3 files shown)


conda_forge_tick/migrators/arch.py
```diff
@@ -144,6 +144,26 @@ def _filter_stubby_and_ignored_nodes(graph, outputs_lut, ignored_packages):
...
+            attrs.get("conda-forge.yml", {}).get("provider", {}).get(arch)
+        ) or (
+            attrs.get("conda-forge.yml", {}).get("build_platform", {}).get(arch)
+            not in [None, arch]
+        )
+        if not configured_arch:
+            # This arch is not in provider or build_platform
+            return False
+
+    return True
```
h-vetinari: This should have the opposite default. It should be false by default, and only true if we can detect it.

conda_forge_tick/migrators/core.py
```diff
@@ -958,6 +958,20 @@ def all_predecessors_issued(self, attrs: "AttrsTypedDict") -> bool:
 
         return True
 
+    def predecessor_already_migrated(self, attrs: "AttrsTypedDict") -> bool:
+        """If true, treat a predecessor as built even without a bot PR record.
+
+        Migrators whose migrated state is visible in the feedstock itself -- the arch
+        migrators, which read it off ``conda-forge.yml`` -- override this. Without it
```
h-vetinari, line 965: Actually, we could also do this for normal (non-arch) migrators, because we have a specific file
```
.ci_support/migrations/<the_migrator>.yaml
```
to look for in the repo state (with some care to read & match the `migration_number`). Not as a requirement for this PR, but this would help in other situations as well.

h-vetinari (commented):
This looks like a very solid improvement to me. Thanks a lot for working on this!

conda_forge_tick/migrators/arch.py
velonica0: fixed. _CrossCompileRebuild.arches defaults to {}, so the old return True fired vacuously and would have marked every predecessor as built.

conda_forge_tick/migrators/core.py
velonica0: One wrinkle: .ci_support/migrations/ isn't in the graph data today. feedstock_parser globs .ci_support/*.yaml non-recursively and keeps only the first file.
So this would also need the parser to capture those files (plus a models/node_attributes.py update), not just an override.

tests/test_graph_migrator_predecessors.py
```diff
@@ -0,0 +1,125 @@
...
+
+def test_no_arches_is_not_configured():
+    # an empty `arches` means there is nothing to detect, not that the feedstock is
+    # configured for everything -- the latter would mark every predecessor as built
+    parent = _payload(
+        "lapack",
+        conda_forge_yml={"build_platform": {"linux_riscv64": "linux_64"}},
+    )
+
+    assert not _arches_are_configured(parent, {})
```
h-vetinari, line 125: It's not so much empty `arches` that I was worried about, just that it's a dangerous default to mark things as done. This test is mostly confusing, but as long as Matt doesn't mind, it's fine to keep it for me.

conda_forge_tick/migrators/arch.py
```diff
@@ -144,6 +144,31 @@ def _filter_stubby_and_ignored_nodes(graph, outputs_lut, ignored_packages):
...
+        configured_arch = (
+            attrs.get("conda-forge.yml", {}).get("provider", {}).get(arch)
+        ) or (
+            attrs.get("conda-forge.yml", {}).get("build_platform", {}).get(arch)
+            not in [None, arch]
+        )
+        if not configured_arch:
+            # This arch is not in provider or build_platform
+            return False
+        configured = True
```
h-vetinari: there's a lot of redundancy here. These lines can go completely. But more than that, the whole thing needs to respect the fact that any missing architecture should fail the test. I think the best fix would be to replace the whole thing with an `all` call over a list comprehension
```python
def already_has_arch(arch):
        return (
            attrs.get("conda-forge.yml", {}).get("provider", {}).get(arch)
        ) or (
            attrs.get("conda-forge.yml", {}).get("build_platform", {}).get(arch)
            not in [None, arch]
        )
# ternary condition because `all` over empty list is always true; we want to default to False
return all(already_has_arch(arch) for arch in arches) if arches else False
```

conda_forge_tick/migrators/core.py
h-vetinari: Could you write this up in an issue? Then it doesn't get forgotten :)

beckermr: We can keep a list of files found. We should not keep the contents probably.

conda_forge_tick/migrators/arch.py
beckermr: You really need to use `get_keys_default` here. Sometimes the bot will set a property to `null` in its json and so then python like this

```python
attrs.get("key1", {}).get("key2, "blah")
```

will error since `key1` will have a value of `None`. 

The other option is to include an `or`:

```python
(attrs.get("key1", {}) or {}).get("key2, "blah")
```

However, that gets very annoying to type and look at.

`get_keys_default` does all of this for you.

h-vetinari: > You really need to use `get_keys_default` here.

In fairness to @velonica0, this is a verbatim copy of the code we've had thus far (which didn't ever blow up AFAICT). I had also been wondering about the missing fallbacks, but thought (without checking) "perhaps it's something special to `attrs: "AttrsTypedDict"`?" 

In any case, I think this that cleaning this up is a fair ask if we're breaking this out into a function, so +1 for `get_keys_default`

velonica0: The helper is now all(_already_has_arch(arch) for arch in arches) if arches else False, with get_keys_default for both lookups.

conda_forge_tick/migrators/core.py
velonica0: I have already created #6691.

h-vetinari (approved):
Thanks a lot!

PR #6763, opened 2026-09-14
Let MigrationYaml detect already-migrated predecessors from ci_support_migrations

### LINKED ISSUE #6691: Let non-arch migrators detect already-migrated predecessors from `.ci_support/migrations/`
Follow-up from #6680, requested by @h-vetinari [here](https://github.com/conda-forge/conda-forge-bot/pull/6680#discussion_r3970031689).

## Background

`GraphMigrator.predecessors_not_yet_built` answers "has this parent been built?" purely from `pr_info.PRed`. A feedstock that got a migration some other way — edited by hand, or a rerender that landed the change — never gets a `PRed` entry, and nothing can add one after the fact. Such a parent blocks every descendant permanently.

Not hypothetical: on the riscv64 migration `blas` and `lapack` are both built for `linux-riscv64` with no `LinuxRISCV64` record, which stalls `gsl` → `r-base` → ~29 `r-*` packages.

#6680 fixes this for the arch migrators via a `predecessor_already_migrated` hook they answer from `conda-forge.yml`. Non-arch migrators keep the old behaviour because they have no equivalent signal wired up.

## Proposal

CFEP-09 migrations have an equivalent on-disk signal — `MigrationYaml.migrate` writes the pinning yaml into the feedstock (`migration_yaml.py:442-445`):

```python
os.makedirs("migrations", exist_ok=True)
with pushd("migrations"):
    with open(f"{self.name}.yaml", "w") as f:
        f.write(self.yaml_contents)
```

So `.ci_support/migrations/<name>.yaml` marks "this migration was applied here", and `MigrationYaml` could override the same hook. It already carries `migration_number` as its `obj_version` (`migration_yaml.py:286`).

## What it needs

1. **`feedstock_parser.py`** to capture those files. Today it globs `.ci_support/*.yaml` non-recursively (`feedstock_parser.py:358-372`), so `migrations/` is excluded entirely, and it keeps only the *first* file's contents under `ci_support_<name>`. Per @beckermr ([here](https://github.com/conda-forge/conda-forge-bot/pull/6680#discussion_r3971033336)), this should record the **list of files found, not their contents** — these yamls carry the full pinning payload and a multi-paragraph `commit_message`, which has no business being replicated across ~25k feedstocks in `node_attrs`.
2. **`models/node_attributes.py`** to document the new field.
3. **A `MigrationYaml.predecessor_already_migrated` override.** No change needed in `core.py` — the hook already dispatches per-migrator.

velonica0:
Closes #6691. Extends #6680's `predecessor_already_migrated` hook to CFEP-09 migrations, using the `ci_support_migrations` data added in #6695 and #6704 — thanks @beckermr.

## Change

`MigrationYaml.predecessor_already_migrated` looks up `ci_support_migrations[self.name]` — `self.name` is the yaml filename stem, the same string `migrate()` uses when writing `.ci_support/migrations/<name>.yaml` — and treats the predecessor as built only if **both** recorded fields match:

```python
return bool(
    info.get("migration_number") == self.obj_version
    and info.get("migrator_ts") == self.loaded_yaml.get("migrator_ts")
)
```


conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -312,6 +312,25 @@ def __init__(
...
+        feedstock holding an older copy has not had *this* migration applied.
+        """
+        info = get_keys_default(attrs, ["ci_support_migrations", self.name], {}, None)
+        if not info:
+            return False
+
+        return bool(
+            info.get("migration_number") == self.obj_version
+            and info.get("migrator_ts") == self.loaded_yaml.get("migrator_ts")
+        )
```
h-vetinari: The call to bool is a bit redundant here, but in any case: it's nice to see how easily this can be implemented 🤩

h-vetinari (approved):
LGTM; I'll let @beckermr take a look too of course

conda_forge_tick/migrators/migration_yaml.py
velonica0: Thank you for your suggestion!

conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -312,6 +312,24 @@ def __init__(
...
+    def predecessor_already_migrated(self, attrs: "AttrsTypedDict") -> bool:
+        """Treat a predecessor as built if the feedstock carries this exact migration.
+
+        ``migrate`` writes the migration yaml to ``.ci_support/migrations/<name>.yaml``
+        and the feedstock parser records each such file's ``migration_number`` and
+        ``migrator_ts`` under ``ci_support_migrations``. Both have to match: a
+        bumped number is a distinct migration to the bot (it is part of the uid),
+        and a changed timestamp is a distinct migration to conda-smithy, so a
+        feedstock holding an older copy has not had *this* migration applied.
+        """
```
beckermr, line 324:
```suggestion
        """
        # Some of the graph nodes have this field as a list due to
        # a change in the format of the data. We skip this test for
        # these nodes for now. Once all of the graph data is converted
        # this bit will be a no-op.
        if isinstance(attrs.get("ci_support_migrations", None), list):
            return False
```

beckermr: We need to account for the initial format here, at least for a while. Eventually all data should be converted automatically, but this can take a bit.

velonica0: Thank you for your suggestion!

conda_forge_tick/migrators/migration_yaml.py
```diff
@@ -312,6 +312,30 @@ def __init__(
...
+        # Some of the graph nodes have this field as a list due to
+        # a change in the format of the data. We skip this test for
+        # these nodes for now. Once all of the graph data is converted
+        # this bit will be a no-op.
+        if isinstance(attrs.get("ci_support_migrations", None), list):
+            return False
```
beckermr, line 330:
```suggestion
            return False

```

beckermr (approved):

PR #6805, opened 2026-09-18
fix: migrate schema to avoid validation errors

beckermr:
#### Description:

As it says above...

#### Checklist:


conda_forge_tick/make_graph.py
```diff
@@ -170,6 +170,12 @@ def _migrate_schema(name, sub_graph):
         with lazy_json_transaction():
             sub_graph.pop("last_updated")
 
+    if "ci_support_migrations" in sub_graph and isinstance(
+        sub_graph["ci_support_migratoons"], list
```
beckermr:
```suggestion
        sub_graph["ci_support_migrations"], list
```

PR #6809, opened 2026-09-18
fix: keep track of cwd for lazy json

beckermr:
#### Description:

The LazyJSON classes refer to a specific location on disk sometimes. As the current working directory changes, they can get out of sync and write back to the wrong spot. This PR keeps track of the CWD when the object is created and makes sure to use that CWD when doing operations. By extension, the backends which deal with files also have a concept of a CWD and those now track it too.

#### Checklist:

- [x] tests

This bug may be the cause of the weird file location errors I saw when trying to deploy the bot data as it runs.


conda_forge_tick/lazy_json_backends.py
```diff
@@ -776,7 +792,7 @@ def hset(self, name, key, value):
             {
                 "$set": {
                     "node": key,
-                    "value": orjson.loads(value),
+                    "value": orjson.loads(value, cwd=self._cwd),
```
beckermr:
```suggestion
                    "value": orjson.loads(value),
```

conda_forge_tick/lazy_json_backends.py
```diff
@@ -796,7 +812,7 @@ def hmset(self, name, mapping):
                     {
                         "$set": {
                             "node": key,
-                            "value": orjson.loads(value),
+                            "value": orjson.loads(value, cwd=self._cwd),
```
beckermr:
```suggestion
                            "value": orjson.loads(value),
```

conda_forge_tick/settings.py
```diff
@@ -137,7 +137,7 @@ def node_attrs_github_backend_raw_base_url(self) -> str:
     Therefore, you don't need to set this when debugging locally.
     """
 
-    frac_update_node_attrs: Fraction = 0.1
+    frac_update_node_attrs: Fraction = 0.01
```
beckermr, line 141:
```suggestion
    frac_update_node_attrs: Fraction = 0.1
```

beckermr:
This PR does not appear to fix the deploy issue for updating nodes.

PR #6821, opened 2026-09-20
Add a `run-minimigrators` command

mgorny:
#### Description:

Add a `run-minimigrators` command that runs the standard set of minimigrators against a local feedstock.  This enables users to easily resolve a bunch of lints and other cleanups that normally happen only when the bot is filing a PR.

#### Checklist:


mgorny:
That said, I don't seem to be able to figure out the typing here. I kinda feel like the existing code relies on some of the interim function being non-typed to avoid the type mismatches.

beckermr:
Yes, the typing is incomplete and not maintained. Many contributors want typing, but I personally don't care for typing. I find it pedantic, and the errors it fails with are opaque/confusing. So I simply ignore it when it is in my way, and don't always bother to add it in depending on what I am doing.

beckermr (commented):
I know everyone loves `Path` objects these days, but this code doesn't use them consistently. You'll likely encounter fewer errors by sticking with normal strings.

beckermr:
My final comment here is that while running the mini-migrators is great, we may want to invest effort directly into smithy in the form of a formatter to go with the linter.

mgorny:
Yes, there is an ongoing effort into that. Unfortunately, it requires time to finish, and I personally can't afford to help it right now. Hence, this quick hack.

mgorny:
> I know everyone loves `Path` objects these days, but this code doesn't use them consistently. You'll likely encounter fewer errors by sticking with normal strings.

They're only used directly in the added code, and passed to `populate_feedstock_attributes()` which internally uses `Path` and converts `str` to `Path` if passed.

mgorny:
I went for resolving the typing issues in the simplest way possible then.

mgorny:
Thanks!

PR #6827, opened 2026-09-21
Refactor migrator code in common

luhenry:
#### Description:
Refactor common code between cross-compile and native-compile migrators to make it easier to add/move/delete new migrators

#### Checklist:


conda_forge_tick/migrators/arch.py
```diff
@@ -313,180 +310,140 @@ def migrate(
...
-            with open("conda-forge.yml") as f:
-                y = yaml_safe_load(f)
+    migration_pr_title = "Linux aarch64 Migrator"
+    migration_branch_suffix = "arch"
+    migration_name = "aarch64"
+    migration_team = "<code>@</code>conda-forge/arm-arch"
 
-            y_orig = copy.deepcopy(y)
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("name", "aarch64 and ppc64le addition")
```
h-vetinari, line 387: This should not be a setdefault IMO, but
```suggestion
        kwargs["name"] = "aarch64 and ppc64le addition"
```
and keep the original `# do not change the name, but this doesn't actually add ppc anymore` comment.

conda_forge_tick/migrators/arch.py
```diff
@@ -495,39 +452,17 @@ class OSXArm(_CrossCompileRebuild):
     pkg_list_filename = "osx_arm64.txt"
     arches = {"osx_arm64": "osx_64"}
 
+    migration_pr_title = "ARM OSX Migrator"
+    migration_name = "ARM OSX"
+    migration_team = "<code>@</code>conda-forge/help-osx-arm64"
+    migration_branch_suffix = "arm_osx"
+
     def __init__(self, *args, **kwargs):
         kwargs.setdefault("name", "arm osx addition")
```
h-vetinari, line 461: same here IMO. If we really ever need to inherit from `OSXArm`, we can cross that bridge when we get to it.

h-vetinari (commented):
this PR is quite a nice cleanup (net -100 lines), but it does a lot of things. I'm pretty certain we don't really need it for switching riscv/win-arm to native builds.

PR #6772, opened 2026-09-14
Make LinuxRISCV64 migrator a native rebuild

luhenry:
- **Refactor migrator code in common**
- **Make LinuxRISCV64 migrator a native rebuild**

#### Description:

After discussion at the Conda Community meeting, we do want to make the RISC-V migration to target native builds rather than cross-compiles

#### Checklist:

- [x] Pydantic model updated or no update needed

PR #6831, opened 2026-09-21
Use native providers for linux-riscv64, osx-arm64 & win-arm64 migrations

h-vetinari:
Minimal changes condensed from #6319 & #6772


h-vetinari:
I've been testing native builds in 
* https://github.com/conda-forge/numpy-feedstock/pull/401 (v1 recipe)
* https://github.com/conda-forge/scipy-feedstock/pull/331 (v0 recipe)

Basically the only thing missing from my POV is https://github.com/conda-forge/conda-smithy/pull/2701

h-vetinari:
The smithy PR was merged and "only" needs a new release, after which this PR is ready.

In the meantime, I'm reverting the native riscv builds here for now, pending further discussion, because the [current state](https://github.com/conda-forge/conda-forge.github.io/issues/1744#issuecomment-5810481277) of things would IMO lead to unacceptable bottlenecks if native builds were rolled out widely.

CC @isuruf, since you very much wanted native builds; I think a ~10x slowdown compared to other arches is too much though. The osx-arm64 effort proved that we can migrate large chunks of the ecosystem even in cross-compilation (and riscv has the advantage of at least being able to run emulated tests) - I don't see why we couldn't start out like this for riscv too, and keep native builds as opt-in for where they're really necessary/important/useful (at least until we have a larger/faster fleet of riscv runners).

isuruf:
> CC @isuruf, since you very much wanted native builds; I think a ~10x slowdown compared to other arches is too much though. The osx-arm64 effort proved that we can migrate large chunks of the ecosystem even in cross-compilation (and riscv has the advantage of at least being able to run emulated tests) - I don't see why we couldn't start out like this for riscv too, and keep native builds as opt-in for where they're really necessary/important/useful (at least until we have a larger/faster fleet of riscv runners).

osx-arm64 was (or was going to be) a big chunk of our users. That itself is enough motivation to keep cross compiling. RISC-V is a niche system and if we want to enable a large chunk of our ecosystem and put the maintenance burden on our maintainers, we should know that it is being tested. Also providing riscv runners to us is a way for the risc-v community to tell us that they are serious about it. If the native riscv runners are getting saturated, then we should pause the migration instead of carrying it out with cross compilation.

